PR #102, opened 2023-05-17
MacOS window-bar corrections (traffic-lights)

flightmansam:
1. Corrected icon resolutions to 32px (12px scaled)
2. Added correct iconography (see attached traffic-lights.svg in PR)
3. Corrected behaviour onOver (all icons should show)
4. Corrected onBlur/onFocus of the app (icons should go grey).

- window-focused and -blurred ipc from the blur and focus listener in electron.
- window.electron.browser.hasFocus() method kept for now (in case it proves useful).

https://github.com/jeffvli/feishin/assets/34013162/4c46e891-3664-4d96-95ff-63b88bba9b79


flightmansam:
TODO: Allow the traffic-lights to have the hover behaviour when the app is not focused (this may be a limitation of doing a home brew react traffic-lights rather than the version supplied by electon). You can see the current behaviour at the end of the example video (see above).

flightmansam:
Below SVG made in Inkscape 1.2

trafic lights

jeffvli:
Overall looks good!

> TODO: Allow the traffic-lights to have the hover behaviour when the app is not focused (this may be a limitation of doing a home brew react traffic-lights rather than the version supplied by electon). You can see the current behaviour at the end of the example video (see above).

Regarding the above: you can probably just achieve this by updating the conditional ternary to prioritize `hoverMin`.

For example on the minimize button:
```jsx
<MacOsButton
  minButton
  className="button"
  id="min-button"
  onClick={handleMinimize}
>
  <img
    alt=""
    className="icon"
    draggable="false"
    src={hoverMin ? macMinHover : hasFocus ? (hoverMin ? macMinHover : macMin) : macBlurred}
  />
</MacOsButton>
```

Possibly more readable if split out into a separate function:

```jsx
const getMinIcon = () => {
  if (hoverMin) return macMinHover;
  if (!hasFocus) return macBlurred;
  return macMin;
};

<img
  ...
  src={getMinIcon()}
/>
```

flightmansam:
> Regarding the above: you can probably just achieve this by updating the conditional ternary to prioritize hoverMin.

In my testing, I couldn't see the window-tab onOver firing at all when the app isn't focused. Im not a react guy so maybe there is something more to this that I don't know?

jeffvli:
If `onMouseOver` isn't firing for you, it may be different behaviors based on OS. I'm on Windows and don't have a macOS device to test unfortunately.

On Windows, the `onMouseOver` events will be triggered regardless of whether the window is focused or not.

flightmansam:
Ayyy I think that's it. I do recall coming across a thread a few days ago that has something along those lines...

I notice the default-layout object's onOver/onEnter calls do kinda work if the app isn't focused. However if you bring your cursor into the app (and it isn't focused) you will get both a onMouseOver and a onMouseLeave call (even though the mouse didn't leave...).

in default-layout.tsx:
```javascript
export const DefaultLayout = ({ shell }: DefaultLayoutProps) => {
  const { windowBarStyle } = useWindowSettings();

  return (
    <>
      <Layout
        id="default-layout"
        windowBarStyle={windowBarStyle}
        onMouseOver={() => console.log("over")}
        onMouseLeave={() => console.log("leave")}
      >
        {windowBarStyle !== Platform.WEB && <WindowBar />}
        <MainContent shell={shell} />
        <PlayerBar />
      </Layout>
    </>
  );
};
```

https://github.com/jeffvli/feishin/assets/34013162/c86f9d2e-11c5-4bfb-95cb-4590ec693d62

jeffvli:
And that only works on the defaultlayout component?

Also, have you tried testing `onMouseEnter` and `onMouseOut` as well?

flightmansam:
Same behaviour. 

I even added an onMouseEnter listener to the window controls container (`bar: enter`), and it isn't heard at all when app not focused. 

https://github.com/jeffvli/feishin/assets/34013162/9a6b5fdc-8b2b-4e19-8839-f3ab711a1f80

flightmansam:
As much as I hate to say it (given the beautiful work that has gone into this already), using the default electron one solves everything.  And it is one line of code...

```javascript
 mainWindow = new BrowserWindow({
    frame: false,
    titleBarStyle: 'hidden',
    trafficLightPosition: { x: 10, y: 10 },
    ...
```

It looks to be okay from a Windows 11 point of view too?? You might need to confirm....


https://github.com/jeffvli/feishin/assets/34013162/647acf52-af44-4084-bfdd-685e4a6fb9a0

jeffvli:
Oh I had no idea about that macOS functionality for electron.
Ref: https://www.electronjs.org/docs/latest/tutorial/window-customization#apply-custom-title-bar-styles-macos-windows

Can you give me a screenshot of what it looks like on macOS? If using that new method, it might be better to add a separate option to the "Window bar style" setting to use native macOS traffic lights as opposed to replacing the macOS one completely.

flightmansam:
https://github.com/jeffvli/feishin/assets/34013162/211db431-095f-4f65-8e23-007793c0526c

jeffvli:
Are you also using that with the `Web (hidden)` window bar style? 

If so, then a new option would definitely need to be added since the html window controls would still be there.

Procedure would be:

- Add new option to the window bar style select dropdown for `Native (macOS)`.
- Save window bar setting selection to local settings file as well as settings store
- Set the BrowserWindow options based on the stored settings in `main.ts`
- Update all components that handle the WindowBar render as well as the web controls rendering

I think I may just take over this implementation since I may want to do a couple of refactors regarding this, but do let me know if you are already working on it or you want to try finishing this PR.

flightmansam:
I feel like you have bigger fish to fry with this project.  Let me have a crack at it. And if it is not to your liking, no harm done as I would have learnt something. 

The electron window decorations can be hidden dynamically (from main) so if we still want a drop down to react in real time I'll have to pass an ipc message, will that work for you?

jeffvli:
> I feel like you have bigger fish to fry with this project. Let me have a crack at it. And if it is not to your liking, no harm done as I would have learnt something.

Sure, that works for me!

> The electron window decorations can be hidden dynamically (from main) so if we still want a drop down to react in real time I'll have to pass an ipc message, will that work for you?

That would be perfect. Since it's closely related, could you take a look at #109 as well? A `native` option would also hide the other windowbars and set the `frame` property on [BrowserWindow](https://www.electronjs.org/docs/latest/api/browser-window#new-browserwindowoptions) to true.

flightmansam:
Okay sounds groovey!

jeffvli:
Changes were implemented in #122.
Thanks for your work regardless!

PR #111, opened 2023-05-21
Add ability to save/restore queue

kgarner7:
This PR adds an event that is sent from the renderer to main on close, triggering a save of the queue state to a file. On start, this state is then restored.
Performance-wise, for ~6k tracks the saved size was 1.7 MB, and it took maybe a second or two.
Also makes .husky/pre-commit executable (otherwise it would not run)


src/renderer/features/settings/components/general/control-settings.tsx
```diff
@@ -170,6 +173,26 @@ export const ControlSettings = () => {
...
+                ...settings,
+                resume: e.currentTarget.checked,
+              },
+            });
+          }}
+        />
+      ),
+      description: 'When exiting, save the current play queue and restore it when reopening',
+      isHidden: false,
+      title: 'Save queue state on close',
```
jeffvli: Should probably move the `isElectron()` over to `isHidden` rather than disabled.

jeffvli:
Made a couple changes

- Renamed the preload dispatch actions to match the function names
- Moved the save/restore actions to the root `App.tsx` instead of `useCenterControls`. That file is getting a bit too big for its own good.
- Added a `restoreQueue` reducer action to the player store -- this will directly return the updated PlayerData
- Removed all properties except `current` and `queue` when saving to the `queue` file -- the rest of the properties are already persisted via the zustand store persist module in local storage
- Added a new "pause" parameter to the `mpvPlayer.setQueue()` function -- mpv automatically starts playback when a file is loaded so this would immediately begin playback when opening the app which isn't ideal in most cases

PR #113, opened 2023-05-21
Fix full screen overflow

kgarner7:
With large enough song titles, in full screen mode the track description + image takes up the entire screen (as opposed to showing part of the queue). This PR does a few style changes to make the overflowing text actually limited in width, and restrict the image to 60% of the window. (not sure if other numbers are better)

image

image


jeffvli:
Thanks! I did notice there was an issue with it but wasn't quite sure how to fix it.

Issue #52, opened 2023-03-14
Lyrics Support for Jellyfin


Maxplosion:
Hi!

just wanted to let you know that in the next Jellyfin release lyrics support should be included see:
https://github.com/jellyfin/jellyfin/pull/8381

It would be great if this would also be implemented in feishin sometime in the future 

Thanks and keep up the great work!

Twcd:
Who is the provider for the lyrics ?

Maxplosion:
you can either embed them into the file as synced or unsynced lyrics or have them externally as .lyc or .txt file

there is no automated lyric load in jellyfin so you provide them on your own or with an external tool

jeffvli:
Awesome, glad that Jellyfin has added native lyrics support!

I added a section for lyrics over in #27, so I'll add this when I have a chance.

jeffvli:
Just a note once this feature is released: local lyrics via txt/lrc files in Jellyfin is currently only available on Jellyfin's `unstable` release.

NicholasFlamy:
> Just a note once this feature is released: local lyrics via txt/lrc files in Jellyfin is currently only available on Jellyfin's `unstable` release.

Does this support embedded lyrics from Jellyfin? I would think not because I helped move that along and it got merged about a week ago.

Daivy03:
Jellyfin has implemented lyrics from version 10.9.0 and up
Refer to this[ jellyfin forum thread](https://forum.jellyfin.org/t-10-9-lyrics-for-songs?highlight=lyric)

PR #118, opened 2023-05-23
Initial implementation for lyrics

kgarner7:
This is a draft of a potential implementation for displaying lyrics (both synced and not). Currently only Navidrome, because it looks like the Jellyfin fetcher doesn't currently get lyrics?
Inspired by https://github.com/navidrome/react-music-player. It works best if you just start a song from zero and continue (and performs better in the electron app for some reason...); seeking has mixed results

### DIFF
--- a/src/main/preload/lyrics.ts
+++ b/src/main/preload/lyrics.ts
@@ -1,17 +1,17 @@
 import { IpcRendererEvent, ipcRenderer } from 'electron';
 import { QueueSong } from '/@/renderer/api/types';
 
-const fetchLyrics = (song: QueueSong) => {
+const fetchRemoteLyrics = (song: QueueSong) => {
   ipcRenderer.send('lyric-fetch', song);
 };
 
-const getLyrics = (
+const retrieveRemoteLyrics = (
   cb: (event: IpcRendererEvent, songName: string, source: string, lyric: string) => void,
 ) => {
   ipcRenderer.on('lyric-get', cb);
 };
 
 export const lyrics = {
-  fetchLyrics,
-  getLyrics,
+  fetchRemoteLyrics,
+  retrieveRemoteLyrics,
 };
--- a/src/main/preload/lyrics.ts
+++ b/src/main/preload/lyrics.ts
@@ -5,7 +5,7 @@ const fetchRemoteLyrics = (song: QueueSong) => {
   ipcRenderer.send('lyric-fetch', song);
 };
 
-const retrieveRemoteLyrics = (
+const remoteLyricsListener = (
   cb: (event: IpcRendererEvent, songName: string, source: string, lyric: string) => void,
 ) => {
   ipcRenderer.on('lyric-get', cb);
@@ -13,5 +13,5 @@ const retrieveRemoteLyrics = (
 
 export const lyrics = {
   fetchRemoteLyrics,
-  retrieveRemoteLyrics,
+  remoteLyricsListener,
 };
--- a/src/main/preload/lyrics.ts
+++ b/src/main/preload/lyrics.ts
@@ -1,0 +1,17 @@
+import { IpcRendererEvent, ipcRenderer } from 'electron';
+import { QueueSong } from '/@/renderer/api/types';
+
+const fetchLyrics = (song: QueueSong) => {
+  ipcRenderer.send('lyric-fetch', song);
+};
+
+const getLyrics = (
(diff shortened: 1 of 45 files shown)


jeffvli:
To do:
 - [x] Add Jellyfin lyrics (referenced in #52)
   - [x] Support synced format
   - [x] Support unsynced format
 - [x] Add configurable fetchers for external lyrics (Reference fetchers from [Sunamu](https://github.com/NyaomiDEV/Sunamu/tree/master/src/main/lyricproviders))
   - [x] Netease
   - [x] Genius
 - [x] Add fancier animations for synced lyrics display (use framer motion)
 - [x] Add configurable offset for synced lyrics timing
 - [x] ...more tbd

kgarner7:
All of the above have been added save for framer-motion (and tbd)

jeffvli:
Thanks, I'll review this over the weekend!

src/renderer/features/settings/components/playback/lyric-settings.tsx
```diff
@@ -0,0 +1,110 @@
+import { NumberInput, Switch } from '@mantine/core';
```
jeffvli: These need to be imported from `/@/renderer/components` rather than mantine.

src/renderer/features/lyrics/lyrics.tsx
```diff
@@ -0,0 +1,131 @@
...
+
+const lyrics = isElectron() ? window.electron.lyrics : null;
+
+const ipc = isElectron() ? window.electron.ipc : null;
+
+// use by https://github.com/ustbhuangyi/lyric-parser
+const timeExp = /\[(\d{2,}):(\d{2})(?:\.(\d{2,3}))?]([^\n]+)\n/g;
+
+export const Lyrics = () => {
+  const currentServer = useCurrentServer();
```
jeffvli: The queue is independent of the current server, meaning you'll need to use `serverId` from the `currentSong` to determine what the actual server is.

src/renderer/features/lyrics/lyrics.tsx
```diff
@@ -0,0 +1,131 @@
...
+
+    return () => {
+      ipc?.removeAllListeners('lyric-get');
+    };
+  }, []);
+
+  useEffect(() => {
+    if (currentSong && !currentSong.lyrics && !remoteLyrics.isLoading && !remoteLyrics.isSuccess) {
+      lyrics?.fetchLyrics(currentSong);
+    }
```
jeffvli: This check is not working as expected.
In `lyric-query.ts` -- `useSongLyrics`, the `queryFn` returns null if the server is not Jellyfin.

This causes `remoteLyrics.isSuccess` to always equal true because no error was thrown in the queryFn. The condition will always be falsy and the fetchLyrics is skipped if the song is Navidrome.

Let's split the conditional to make it clearer
```typescript
    const hasTaggedLyrics = currentSong && currentSong.lyrics;
    const hasLyricsResponse =
      !remoteLyrics.isLoading && remoteLyrics?.isSuccess && remoteLyrics?.data !== null;

    if (!hasTaggedLyrics && !hasLyricsResponse) {
      lyrics?.fetchLyrics(currentSong);
    }
```

src/main/preload/lyrics.ts
```diff
@@ -0,0 +1,17 @@
...
+
+const fetchLyrics = (song: QueueSong) => {
+  ipcRenderer.send('lyric-fetch', song);
+};
+
+const getLyrics = (
+  cb: (event: IpcRendererEvent, songName: string, source: string, lyric: string) => void,
+) => {
+  ipcRenderer.on('lyric-get', cb);
+};
```
jeffvli: These two function names are a bit confusing as to what they actually do.

`fetchLyrics` -- fetches lyrics from remote sources as opposed to from the server. Rename this to `fetchRemoteLyrics` or similar

`getLyrics` -- listens for and retrieves the lyrics response from the remote sources. getLyrics is already used in the api controller, so rename to `retrieveRemoteLyrics` or similar

jeffvli: Thinking about it again, the listener function should probably be `remoteLyricsListener` since `retrieveRemoteLryics` makes it also sound like a fetch function. In the `mpv-player` preload, I had all of the listener functions under a separate named export called `mpvPlayerListener` so it was easy to identify. Since it's combined here, adding `listener` somewhere in the name should be used so it can be easily identified as such.

PR #119, opened 2023-05-23
Fix MPRIS on Linux, muted, and add media session

kgarner7:
Changes:
- Fix MPRIS: shuffle is a boolean, not a string.
- Fix MPRIS: handle volume properly (get/set)
- Make mute button actually work in the web ui
- Add navigator.mediasession to show current playing track/album/info and support and support media keys

### DIFF
--- a/src/renderer/features/player/hooks/use-center-controls.ts
+++ b/src/renderer/features/player/hooks/use-center-controls.ts
@@ -24,6 +24,7 @@ const mpvPlayerListener = isElectron() ? window.electron.mpvPlayerListener : nul
 const ipc = isElectron() ? window.electron.ipc : null;
 const utils = isElectron() ? window.electron.utils : null;
 const mpris = isElectron() && utils?.isLinux() ? window.electron.mpris : null;
+const mediaSession = !isElectron() ? navigator.mediaSession : null;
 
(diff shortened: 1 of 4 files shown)


src/renderer/features/player/hooks/use-center-controls.ts
```diff
@@ -668,10 +754,8 @@ export const useCenterControls = (args: { playersRef: any }) => {
       });
 
       mpris.requestVolume((_e: any, data: { volume: number }) => {
-        const currentVolume = usePlayerStore.getState().volume;
-        const resultingVolume = data.volume + currentVolume;
+        let newVolume = data.volume * 100;
```
gelaechter: newVolume should be rounded using `Math.round()` so we don't get values like 50.000000000.

gelaechter: Also make sure to notify other mpris clients of the volume change by again calling `mpris?.updateVolume(data.volume);` when receiving the volume request, so they can keep track of Feishins volume.
This is important, so clients like playerctl work properly.

gelaechter (commented):
Thanks for your work and the quick implementation, apart from my other two comments this seems to work great :)

jeffvli (approved):

PR #131, opened 2023-06-05
Lyrics Improvements

### LINKED ISSUE #52: Lyrics Support for Jellyfin

### LINKED ISSUE #98: [Feature Request] Lyrics view
I would like to be able to view the lyrics of songs while playing them. (e.g. taking the lyrics from Musixmatch)

kgarner7:
- Make the settings text actually consistent with behavior
- Add metadata (artist/track name) for fetched tracks
- Add ability to remove incorrectly fetched lyric
- Add lyric fetch cache; save the last 10 fetches
- Add ability to change offset in full screen, add more comments

Sorry for another one of these so close.
Todo (?):
- [x] Add spinner when fetching? Not sure what counts as fetching; Jellyfin fetch? RPC Fetch? If only the first, this is easy, otherwise you'll need either a timeout or an ACK for no response...
- [x] Add the ability to manually search for lyrics (this would require a decent amount of work)...


jeffvli:
After reviewing the implementation again, I believe it can be greatly simplified especially if we want to support loading indicators as well as manual search.

The current implementation is a three way back and forth process that follows some of the existing patterns in the app:
  1. When song changes, send a fetch request to the main process from the renderer
  2. Main process begins the fetch
  3. Listener function on the renderer receives the output of the fetch from the main process and then we can perform actions on it

This works but we don't have a way to easily check when the function is actually running unless we were to add an additional listener for that.

After a bit of research, it turns out you can return a promise from the main process using [ipcRenderer.invoke and ipcMain.handle](https://stackoverflow.com/questions/45148110/how-to-add-a-callback-to-ipc-renderer-send). Using this method we can now fetch the lyrics on demand.

If you don't mind, I can rewrite the lyric fetching process and then we can work on the other features after.

jeffvli:
I pushed some changes to a [separate branch](https://github.com/jeffvli/feishin/commits/lyrics-enhancements).

- Replaces the original lyric fetch and listener pattern
- Moves all lyric fetching logic into the query hook to support loaders and use built-in caching via react-query
  - We can now potentially remove the in-memory caching in the main process 

Can you review and let me know your thoughts? If you're ok with it, I'll push it to your branch or you can just merge it into yours.

TODO:
- [x] Add manual search
  - [x] `MAIN` For each source, if possible by the source, add a function that returns an array of search results. Parameters for the function should include "name" and "artist"
  - [x] `MAIN` Add new `invoke` and `handle` functions to fetch the search results
  - [x] `RENDERER` Add new query that fetches the lyric search results
  - [x] `RENDERER` Add modal form UI with inputs for "name" and "artist" pre-filled with existing song data
- [ ] ~~Persist overridden song lyrics (tbd, need to determine best way to do this)~~ This will come in a later PR

kgarner7:
Looks good! I merged it in, and brought back the metadata/clear options.

PR #132, opened 2023-06-07
initial implementation for password saving

kgarner7:
For now, it just saves the password in the edit page, uses electron's safeStorage

### DIFF
--- a/src/renderer/api/jellyfin/jellyfin-api.ts
+++ b/src/renderer/api/jellyfin/jellyfin-api.ts
@@ -267,8 +267,27 @@ axiosClient.defaults.paramsSerializer = (params) => {
   return qs.stringify(params, { arrayFormat: 'repeat' });
 };
 
+let authSuccess = true;
+let shouldDelay = false;
+
+const RETRY_DELAY_MS = 1000;
+const MAX_RETRIES = 5;
+
+const waitForResult = async (count = 0): Promise<void> => {
+  return new Promise((resolve) => {
+    if (count === MAX_RETRIES || !shouldDelay) resolve();
+
+    setTimeout(() => {
+      waitForResult(count + 1)
+        .then(resolve)
+        .catch(resolve);
+    }, RETRY_DELAY_MS);
+  });
+};
+
 axiosClient.interceptors.response.use(
   (response) => {
+    authSuccess = true;
     return response;
   },
   (error) => {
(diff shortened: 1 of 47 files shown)


src/renderer/features/settings/components/general/control-settings.tsx
```diff
@@ -192,6 +192,28 @@ export const ControlSettings = () => {
...
+              },
+            });
+          }}
+        />
+      ),
+      description:
+        'Save server passwords (encrypted by OS). Deselecting this will immediately remove any saved passwords',
+      isHidden: !isElectron(),
+      title: 'Save passwords',
+    },
```
jeffvli: This feels a bit awkward in the actual settings page. The toggle should be moved to the "Manage servers" modal alongside the other 2 toggles.

src/renderer/api/jellyfin/jellyfin-api.ts
```diff
@@ -269,19 +273,52 @@ axiosClient.interceptors.response.use(
...
+            error.config.headers['X-MediaBrowser-Token'] = token;
+
+            return axiosClient.request(error.config);
+          })
+          .catch((newError: any) => {
+            console.error('Error when trying to reauthenticate: ', newError);
+            authenticationFailure(currentServer);
+          });
       }
+      authenticationFailure(currentServer);
```
jeffvli: This feature can actually be omitted completely for Jellyfin.
Jellyfin's access tokens do not expire thus if the user receives a 401 it's likely their password has been changed.

jeffvli (changes requested):

src/renderer/api/jellyfin/jellyfin-api.ts
kgarner7: Neat! I just removed that

src/renderer/api/navidrome/navidrome-api.ts
```diff
@@ -182,54 +228,78 @@ axiosClient.interceptors.response.use(
...
+            });
 
-  return {
-    params: notNilParams,
-    path,
-  };
-};
+            if (res.status !== 200) {
+              throw new Error('Failed to authenticate');
+            }
```
jeffvli, line 292: An additional conditional could be added to catch the `http 429` error so that it doesn't log you out in the case that multiple login requests are sent. (Retry the request with the token in the store which should be updated from the previous requests).

Edit: Or we can just inform users to configure the [ND_AUTHREQUESTLIMIT](https://www.navidrome.org/docs/usage/configuration-options/#advanced-configuration) to a higher value or disable it.

jeffvli (approved):

src/renderer/api/navidrome/navidrome-api.ts
kgarner7: Given that the default timeout is 20 seconds, I think it probably makes more sense to just give them a warning

PR #136, opened 2023-06-10
Add LrcLib Fetcher

kgarner7:
This PR primarily adds LrcLib (backing infrastructure of https://github.com/tranxuanthang/lrcget).
Pros:
- Single search is *extremely* precise
- returns synchronized and unsynchronized (if it exists)

Cons:
- getSearchResults is less accurate, and some parameters unknown
- some queries might just time out

Minor changes:
- remove any in genius.ts (the type is well defined already)
- do not show search/clear buttons if there are no lyric fetchers

### DIFF
--- a/src/main/features/core/lyrics/index.ts
+++ b/src/main/features/core/lyrics/index.ts
@@ -14,6 +14,11 @@ import {
   getLyricsBySongId as getGenius,
 } from './genius';
 import {
+  query as queryLrclib,
+  getSearchResults as searchLrcLib,
+  getLyricsBySongId as getLrcLib,
+} from './lrclib';
+import {
   query as queryNetease,
   getSearchResults as searchNetease,
   getLyricsBySongId as getNetease,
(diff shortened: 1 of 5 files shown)


src/main/features/core/lyrics/index.ts
```diff
@@ -61,7 +69,12 @@ const getRemoteLyrics = async (song: QueueSong) => {
   let lyricsFromSource = null;
 
   for (const source of sources) {
-    const params = { artist: song.artistName, name: song.name };
+    const params = {
+      album: song.album || song.name,
```
jeffvli, line 73: Is there a reason the album parameter is falling back to the song name instead of being left undefined if there's no album?

kgarner7: Yes; lrclib requires you to provide an album parameter, and I suspect if there is no album name then it might be a single. I guess an alternative is to just make that assumption in the lrclib fetcher?

jeffvli: Ok yeah it seems all the parameters are required in the search query. 
It's fine to leave as-is then.

Issue #130, opened 2023-06-03
Fullscreen player interface does not scale well at small window sizes


Gecked-Deck:
## Expected Behavior

The fullscreen player's UI should scale properly to smaller window sizes

## Current Behavior

The fullscreen player UI does not scale well. At the smallest window size the album art is unrecognisable and the volume slider is cut off (This is also an issue in the normal player).

![image](https://github.com/jeffvli/feishin/assets/104325921/08e18b98-9ad8-4c46-a480-cc455bee8dab)

The stop, shuffle, repeat and shuffle all buttons also get hidden at small window sizes. (Large vs Small window comparison)

![image](https://github.com/jeffvli/feishin/assets/104325921/a029fae2-94fe-4664-a965-9c356a960854)
![image](https://github.com/jeffvli/feishin/assets/104325921/573e4545-5817-4c8b-908f-b5365314e6e9)

## Steps to Reproduce

1. Open Feishin

2. Enter the fullscreen player

3. Resize the window

## Context

It would be nice for the fullscreen player to scale nicely, as I like to have my music player pinned on top of other apps however in it's current state it looks slightly broken.

## Your Environment

- Application version : 0.1.1
- Operating System and version  : Fedora Linux 38.20230531.0 (Kinoite) x86_64
- Server and version  : Jellyfin v10.8.10

jeffvli:
This is mostly fixed in the latest dev versions.

Unfortunately not all the controls can fit when sized to the minimum width on 100% scaling. At least 2 items will be hidden (if skip buttons are enabled).

Issue #124, opened 2023-05-25
Cover art proportion problem at certain screen resolution in fullscreen player


AuroraViola:
## Expected Behavior

In fullscreen player song cover art should be a perfect square

## Current Behavior

It is not a perfect square
![image](https://github.com/jeffvli/feishin/assets/110638310/0980d380-cdce-44f5-96ad-3b4980d383d0)

## Steps to Reproduce

1. Activate fullscreen player
2. Scale the window to 1366x768 (my monitor resolution)
3. look at the cover art proportion

## Context

I don't like a cover art that should be a square being a rectangle

## My Environment

- Application version: v0.1.1
- Operating System and version: Fedora 38
- Server and version: Navidrome v0.49.3

jeffvli:
Related to #130 and #137

PR #140, opened 2023-06-11
Add zooming functionality

gelaechter:
Hey, this should allow scaling the app and maybe even be a quick fix for #124 and #130.
I noticed that we aren't able to use the plus key in hotkeys. There doesn't seem to be a way around it, since [Mantines hotkey parser splits hotkey strings at pluses](https://github.com/mantinedev/mantine/blob/8c12a76c56da51af34213f18dd67c8b72a0ddb44/src/mantine-hooks/src/use-hotkeys/parse-hotkey.ts#L18). Do you maybe know a way around this to allow mod+plus and mod+minus to be the hotkeys for zooming?

### DIFF
--- a/src/renderer/features/settings/components/general/application-settings.tsx
+++ b/src/renderer/features/settings/components/general/application-settings.tsx
@@ -1,9 +1,12 @@
-import { Select } from '/@/renderer/components';
+import isElectron from 'is-electron';
+import { NumberInput, Select } from '/@/renderer/components';
 import {
   SettingsSection,
   SettingOption,
 } from '/@/renderer/features/settings/components/settings-section';
 import { useGeneralSettings, useSettingsStoreActions } from '/@/renderer/store/settings.store';
+
+const localSettings = isElectron() ? window.electron.localSettings : null;
 
 const FONT_OPTIONS = [
   { label: 'Archivo', value: 'Archivo' },
@@ -52,6 +55,30 @@ export const ApplicationSettings = () => {
       description: 'Sets the application content font',
       isHidden: false,
       title: 'Font (Content)',
+    },
+    {
+      control: (
+        <NumberInput
+          disabled={!isElectron()}
+          max={300}
+          min={50}
+          value={settings.zoomFactor}
+          onChange={(e) => {
+            console.log(e);
+            if (!e) return;
+            setSettings({
+              general: {
+                ...settings,
+                zoomFactor: e,
+              },
+            });
+            localSettings.setZoomFactor(e);
+          }}
+        />
+      ),
+      description: 'Sets the application zoom factor in percent',
+      isHidden: false,
+      title: 'Zoom factor',
     },
   ];
 
--- a/src/renderer/features/settings/components/general/application-settings.tsx
+++ b/src/renderer/features/settings/components/general/application-settings.tsx
@@ -63,16 +63,18 @@ export const ApplicationSettings = () => {
           max={300}
           min={50}
           value={settings.zoomFactor}
-          onChange={(e) => {
-            console.log(e);
+          onBlur={(e) => {
             if (!e) return;
+            const newVal = e.currentTarget.value
+              ? Math.min(Math.max(Number(e.currentTarget.value), 50), 300)
+              : settings.zoomFactor;
             setSettings({
               general: {
                 ...settings,
-                zoomFactor: e,
+                zoomFactor: newVal,
               },
             });
-            localSettings.setZoomFactor(e);
+            localSettings.setZoomFactor(newVal);
           }}
         />
       ),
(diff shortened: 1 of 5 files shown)


jeffvli:
Thanks for the PR!

> I noticed that we aren't able to use the plus key in hotkeys. There doesn't seem to be a way around it, since [Mantines hotkey parser splits hotkey strings at pluses](https://github.com/mantinedev/mantine/blob/8c12a76c56da51af34213f18dd67c8b72a0ddb44/src/mantine-hooks/src/use-hotkeys/parse-hotkey.ts#L18). Do you maybe know a way around this to allow mod+plus and mod+minus to be the hotkeys for zooming?

Apparently you can get around this by modifying the `debouncedSetHotkey` function in `hotkey-manager-settings.tsx` to replace any `+` keys with just `equal`.

```typescript
if (!IGNORED_KEYS.includes(e.key)) {
  if (e.code.includes('Numpad')) {
    if (e.key === '+') keys.push('numpadadd');
    else if (e.key === '-') keys.push('numpadsubtract');
    else if (e.key === '*') keys.push('numpadmultiply');
    else if (e.key === '/') keys.push('numpaddivide');
    else if (e.key === '.') keys.push('numpaddecimal');
    else keys.push(`numpad${e.key}`.toLowerCase());
  } else if (e.key === '+') {
    keys.push('equal');
  } else {
    keys.push(e.key?.toLowerCase());
  }
}
```

src/renderer/features/settings/components/general/application-settings.tsx
```diff
@@ -53,6 +56,30 @@ export const ApplicationSettings = () => {
...
+            if (!e) return;
+            setSettings({
+              general: {
+                ...settings,
+                zoomFactor: e,
+              },
+            });
+            localSettings.setZoomFactor(e);
+          }}
+        />
```
jeffvli: This would be better as on `onBlur` since it's a bit weird when you modify it and it instantly updates with a 1 digit value. Also stray console.log is being output here.

src/renderer/layouts/default-layout.tsx
```diff
@@ -52,7 +54,26 @@ export const DefaultLayout = ({ shell }: DefaultLayoutProps) => {
...
+  const updateZoom = (increase: number) => {
+    const newVal = settings.zoomFactor + increase;
+    if (newVal > 300 || newVal < 50 || !isElectron()) return;
+    setSettings({
+      general: {
+        ...settings,
+        zoomFactor: newVal,
+      },
+    });
+    console.log(settings.zoomFactor);
```
jeffvli: Stray console.log

src/renderer/layouts/default-layout.tsx
```diff
@@ -52,7 +54,26 @@ export const DefaultLayout = ({ shell }: DefaultLayoutProps) => {
...
+    localSettings?.setZoomFactor(settings.zoomFactor);
+  };
+  localSettings?.setZoomFactor(settings.zoomFactor);
+
+  useHotkeys([[bindings.zoomIn.hotkey, () => updateZoom(5)]]);
+  useHotkeys([[bindings.zoomOut.hotkey, () => updateZoom(-5)]]);
   useHotkeys([[bindings.globalSearch.hotkey, () => handlers.open()]]);
```
jeffvli: You can combine this into a single useHotkeys hook instead of instantiating it three times. 
- [Example from docs](https://mantine.dev/hooks/use-hotkeys/)

Also, since this is a desktop-only feature you should probably conditionally add these hotkeys.

```typescript
const zoomHotkeys: HotkeyItem[] = [
  [bindings.zoomIn.hotkey, () => updateZoom(5)],
  [bindings.zoomOut.hotkey, () => updateZoom(-5)],
];

useHotkeys([
  [bindings.globalSearch.hotkey, () => handlers.open()],
  ...(isElectron() ? zoomHotkeys : []),
]);
```

jeffvli (changes requested):

gelaechter:
Thanks for your review! I implemented your comments.
I also implemented your suggestion to replace + with 'equals'. It makes sense that this works on the US layouts since plus is basically shift+equals but keep in mind that this doesn't necessarily work on other layouts where plus is a dedicated key. At least it doesn't work on the German layout. But again, I don't believe there is something we can do about that.

jeffvli:
Have you tried using `add`?

According to [this](https://developer.mozilla.org/en-US/docs/Web/API/UI_Events/Keyboard_event_key_values) it's a numpad value but it might work still.

gelaechter:
Yep, I tried a few: add, Add, plus, Plus, but none work.

jeffvli:
Seems like it might not be possible then unless we switch to a different hotkeys handler or we fork the useHotkeys hook into this repo and modify it to work. Unless you want to try that first this should be good to merge.

gelaechter:
You can merge it now, I might open an issue at mantines repo later.

jeffvli (approved):

PR #157, opened 2023-06-21
Fix label-font-family typo

zackslash:
Corrects a typo found while attempting to diagnose a different font issue.


jeffvli:
Good catch!

PR #164, opened 2023-06-25
Add remote control

kgarner7:
(I know you're out when I post this, so please do not reply).
This PR seeks to add the ability to add a remote control server/socket. Notably, it:
- adds a new build target (remote), which builds a single page app that can be served over HTTP(S)
- adds a new 'core' feature, remote, that can spawn an HTTP server to serve the above file, websocket connection
- creates remote rpc object, moves some MPRIS events over to remote (and also adds strict typing to the exported preloads)


src/main/features/core/remote/index.ts
```diff
@@ -0,0 +1,569 @@
...
+
+type SendData = ServerEvent & {
+    client: StatefulWebSocket;
+};
+
+function send({ client, event, data }: SendData): void {
+    if (client.readyState === WebSocket.OPEN) {
+        client.send(JSON.stringify({ data, event }));
+    }
+}
```
graydosh: For `broadcast()` why not incorporate `send()` inside since it's got the same logic as broadcast?

```suggestion
type SendData = ServerEvent & {
    client: StatefulWebSocket;
};

function send({ client, event, data }: SendData): void {
    if (client.readyState === WebSocket.OPEN) {
        client.send(JSON.stringify({ data, event }));
    }
}

function broadcast({ event, data }: ServerEvent): void {
    if (wsServer) {
        for (const client of wsServer.clients) {
          send({ client, event, data });
        }
    }
}
```

src/main/features/core/remote/index.ts
```diff
@@ -0,0 +1,569 @@
...
+                                });
+
+                                if (mprisPlayer) {
+                                    mprisPlayer.volume = volume / 100;
+                                }
+                                break;
+                            }
+                            case 'favorite': {
+                                const { favorite, id } = json;
+                                if (id && json.id === currentSong.song?.id) {
```
graydosh: nitpick: Why use `id && json.id`? Isn't `id && id === currentSong.song?.id` sufficient?

src/main/features/linux/mpris.ts
```diff
@@ -95,76 +103,69 @@ ipcMain.on('mpris-update-seek', (_event, arg) => {
...
+
+ipcMain.on('update-shuffle', (_event, shuffle: boolean) => {
+    mprisPlayer.shuffle = shuffle;
+});
+
+ipcMain.on('update-song', (_event, args: SongUpdate) => {
+    const { song, status, repeat, shuffle } = args || {};
+
+    try {
+        mprisPlayer.playbackStatus = status === PlayerStatus.PLAYING ? 'Playing' : 'Paused';
```
graydosh, line 127: Should the status also support ["Stopped"](
https://specifications.freedesktop.org/mpris-spec/2.2/Player_Interface.html#Property:PlaybackStatus)?

src/main/features/linux/mpris.ts
```diff
@@ -95,76 +103,69 @@ ipcMain.on('mpris-update-seek', (_event, arg) => {
...
+            : null;
+
+        mprisPlayer.metadata = {
+            'mpris:artUrl': upsizedImageUrl,
+            'mpris:length': song.duration ? Math.round((song.duration || 0) * 1e6) : null,
+            'mpris:trackid': song?.id
+                ? mprisPlayer.objectPath(`track/${song.id?.replace('-', '')}`)
+                : '',
+            'xesam:album': song.album || null,
+            'xesam:albumArtist': song.albumArtists?.length ? song.albumArtists[0].name : null,
```
graydosh: Why is `albumArtist` being condensed to just one artist? [The spec](https://www.freedesktop.org/wiki/Specifications/mpris-spec/metadata/#xesam:albumartist) specifies that it can be an array of strings.

```suggestion
            'xesam:albumArtist': song.albumArtists?.length
              ? song.albumArtists.map((artist: RelatedArtist => artist.name) : null,
```

src/remote/types.ts
```diff
@@ -0,0 +1,56 @@
...
+export interface ServerSong {
+    data: SongUpdateSocket;
+    event: 'song';
+}
+
+export type ServerEvent = ServerError | ServerFavorite | ServerRating | ServerSong | ServerProxy;
+
+export interface ClientSimpleEvent {
+    event: 'next' | 'pause' | 'play' | 'previous' | 'proxy' | 'repeat' | 'shuffle';
+}
```
graydosh, line 37: Should the simple event also include "stop" or is not possible to stop the player and instead only pause?

src/renderer/app.tsx
```diff
@@ -35,6 +41,7 @@ export const App = () => {
     const { bindings } = useHotkeySettings();
     const handlePlayQueueAdd = useHandlePlayQueueAdd();
     const { clearQueue, restoreQueue } = useQueueControls();
+    const { enabled } = useRemoteSettings();
```
graydosh: Maybe a more descriptive name like `remoteEnabled`?

```suggestion
    const { enabled: remoteEnabled } = useRemoteSettings();
```

src/renderer/features/settings/components/general/remote-settings.tsx
graydosh, line 1: General note for settings/remote control: maybe these settings should include an IP whitelist/blacklist setting as well. List can contain either IPs, IP ranges, or CIDR notation (i.e. 192.168.0.0/24) or be empty to signify all/none depending on whitelist/blacklist setting.

RetroJect (commented):
Great job on this! I came to this project looking to see if they had any plans for websockets and was planning on doing it myself if not, but you beat me to it!

src/main/features/linux/mpris.ts
kgarner7: Not here; this only fires when the track changes (there is a separate stop event send via IPC), so it is either playing or paused

src/remote/types.ts
kgarner7: I believe the player is designed such that a 'stop' stops the MPV player (but in the logs it's reported as a pause, and the MPRIS connector maps stop to pause as well).

src/renderer/features/settings/components/general/remote-settings.tsx
kgarner7: I was considering something like HTTP basic auth (user could configure username/password), but didn't go the whole way there. I do agree some form of security would be nice, but I don't think IP allow/blocklist should be on the application (this feels more like the point of a firewall)

src/renderer/features/settings/components/general/remote-settings.tsx
```diff
@@ -0,0 +1,97 @@
+import isElectron from 'is-electron';
+import { SettingsSection } from '/@/renderer/features/settings/components/settings-section';
+import { useRemoteSettings, useSettingsStoreActions } from '/@/renderer/store';
+import { NumberInput, Switch, toast } from '/@/renderer/components';
+import { debounce } from 'lodash';
```
jeffvli: Change lodash import to `import debounce from 'lodash/debounce` to take advantage of tree shaking.

src/remote/app.tsx
```diff
@@ -0,0 +1,85 @@
+import { useEffect } from 'react';
+import { MantineProvider } from '@mantine/core';
+import './styles/global.scss';
+import '@ag-grid-community/styles/ag-grid.css';
```
jeffvli, line 4: Do you need ag-grid styles here?

jeffvli:
Overall looks fine to me.

I'll likely move the settings options to a different tab in the future but that will come later as I add some other integrations such as discord rpc.

If you can fix the current linting issues this should be good to merge.

src/remote/types.ts
graydosh: Got it, that makes sense

src/renderer/features/settings/components/general/remote-settings.tsx
graydosh: I was thinking in terms of being on a network that you don't have control over the network's firewall. Sure you could setup something on the machine itself but having that kind of feature built-in to the settings of the app could be useful as well

PR #167, opened 2023-06-26
Add dockerfile

cedvan:
Add dockerfile to permit start Feishin web interface with docker


jeffvli:
Added a separate implementation in #245

PR #173, opened 2023-07-02
Add fix to hide the "More from this artist" carousel when it contains no albums

BlackDragon-B:
This fix checks if there is actually an album to display since the "More from this artist" title without any albums looks quite silly.

### DIFF
--- a/src/renderer/features/albums/components/album-detail-content.tsx
+++ b/src/renderer/features/albums/components/album-detail-content.tsx
@@ -192,24 +192,20 @@ export const AlbumDetailContent = ({ tableRef }: AlbumDetailContentProps) => {
     });
 
     const carousels = [
-        ...[
-            artistQuery?.data?.items.length
-                ? {
-                      data: artistQuery?.data?.items,
-                      loading: artistQuery?.isLoading || artistQuery.isFetching,
-                      pagination: {
-                          handleNextPage: () => handleNextPage('artist'),
-                          handlePreviousPage: () => handlePreviousPage('artist'),
-                          hasPreviousPage: pagination.artist > 0,
-                          itemsPerPage,
-                      },
-                      title: 'More from this artist',
-                      uniqueId: 'mostPlayed',
-                  }
-                : {},
-        ],
+        {
+            data: artistQuery?.data?.items,
+            isHidden: !artistQuery?.data?.items.length,
+            loading: artistQuery?.isLoading || artistQuery.isFetching,
+            pagination: {
+                handleNextPage: () => handleNextPage('artist'),
+                handlePreviousPage: () => handlePreviousPage('artist'),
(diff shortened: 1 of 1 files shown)


src/renderer/features/albums/components/album-detail-content.tsx
```diff
@@ -191,20 +191,22 @@ export const AlbumDetailContent = ({ tableRef }: AlbumDetailContentProps) => {
...
-                handleNextPage: () => handleNextPage('artist'),
-                handlePreviousPage: () => handlePreviousPage('artist'),
-                hasPreviousPage: pagination.artist > 0,
-                itemsPerPage,
-            },
-            title: 'More from this artist',
-            uniqueId: 'mostPlayed',
-        },
-    ];
+    const carousels = artistQuery?.data?.items.length
```
jeffvli: Thanks for the PR!

This should be changed to use the spread operator to conditionally add that single object to the array rather than conditionally rendering the entire array so that any new carousels can be added easily.

jeffvli (changes requested):

src/renderer/features/albums/components/album-detail-content.tsx
BlackDragon-B: Thanks for the feedback,

Would something like this work better?
f75daffed5b040109de88b58240af3e2d247cf3e

jeffvli: Actually, add a `isHidden` property to the carousel object and then do an array filter before the `map` statement in the jsx render.

If you look at `album-artist-detail-content.tsx` you can see an example of how it was done previously.

src/renderer/features/albums/components/album-detail-content.tsx
```diff
@@ -192,20 +192,24 @@ export const AlbumDetailContent = ({ tableRef }: AlbumDetailContentProps) => {
...
+                          itemsPerPage,
+                      },
+                      title: 'More from this artist',
+                      uniqueId: 'mostPlayed',
+                  }
+                : {},
+        ],
     ];
-
+    console.log(carousels);
```
kittenswolf: This was probably accidentally committed :)

BlackDragon-B: You're right haha. Gonna rectify that this evening since it does nothing anyways.

src/renderer/features/albums/components/album-detail-content.tsx
jeffvli: There's some stylistic changes I've requested if you missed the previous comment!

BlackDragon-B: I think I included all the changes in the latest commit by using array filter like used in `album-artist-detail-content.tsx`

jeffvli (approved):

PR #193, opened 2023-07-29
Fix sidebar layout to show full album cover

### LINKED ISSUE #158: Album cover in the bottom left not square
## Expected Behavior

Снимок экрана 2023-06-23 162750

Album cover art in the bottom left should be square.

## Current Behavior

Alas, the album covers are not square.

Снимок экрана 2023-06-23 162713

It seems like they're missing about 10 pixels from the bottom.

- Application version : 0.2.0

gelaechter:
I had a closer look at the Sidebar layout to fix #158 and now understand how it works.
I don't know how the previous numbers came to be, I assume they're from a now changed previous layout?
If the Titlebar does not have any relevance for the sidebar (I've only ever seen it on the right-hand side of the window when using the “Web” Window bar style) this should be fixed.

Also, a question from a web-dev noob: Why not use something like Flexbox for this kind of layout, wouldn't that make it easier?

### DIFF
--- a/src/renderer/features/sidebar/components/sidebar.tsx
+++ b/src/renderer/features/sidebar/components/sidebar.tsx
@@ -57,10 +57,14 @@ import { Platform } from '/@/renderer/types';
 
 const SidebarContainer = styled.div<{ windowBarStyle: Platform }>`
     height: 100%;
-    max-height: ${(props) =>
-        props.windowBarStyle === Platform.WEB || props.windowBarStyle === Platform.LINUX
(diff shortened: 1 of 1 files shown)


jeffvli:
It was left over from when I initially implemented it while I was testing out framer-motion for the animations. I'm planning on updating the sidebar layout with grid so I may hold off on merging this.

jeffvli:
Actually going to merge this for now until I have time to rework the layout.

PR #222, opened 2023-08-20
Improve MPV initialization and restore

kgarner7:
- set mpv settings only after it has successfully started (at least on linux, settings were not taken)
- change timing of restore queue to behave properly


jeffvli:
Looks good, thanks!

PR #229, opened 2023-08-29
improved client detection

Kakoluz:
Improve reporting of Client name so jellyfin statictics are better calculated instead of gruping every one in a single 'PC' generic device.


jeffvli:
Looks good, thanks!

PR #231, opened 2023-08-31
Add transparency/opacity for queue sidebar

### LINKED ISSUE #233: Synchronized lyrics word wrap changes when active
## Expected Behavior
Words in lyrics do not move as they become highlighted

## Current Behavior
The synchronized lyrics change from font-weight: 400; to font-weight: 800; when they are set as active, this causes their white-space to change and wrap words to the next line.
The issue is observable on a maximized 1920x1080 screen or you can resize the window until the issue is present.
Feishin_3lDcSeTluB

## Steps to Reproduce

1. Open LYRICS tab on a song currently playing with synchronized lyrics

2. Wait for a longer line of lyrics to become active (highlighted).

Changing the default to font-weight: 600; and opacity: 0.3;
Only changing opacity as the lyrics become active would keep their white-space the same at all times and prevent them from wrapping.
Feishin_VzJfvxXkSN

## Context
Listening to music and reading lyrics becomes difficult and very annoying as the words move around the screen, this is a minor problem but still should not be the case.

## Your Environment

-   Application version: 0.3.0
-   Operating System and version: Windows 11
-   Server and version: Jellyfin 10.9.0

kgarner7:
This PR adds the ability to add background opacity for the fullscreen player to make the text more visible if needed.

### DIFF
--- a/src/renderer/features/lyrics/lyric-line.tsx
+++ b/src/renderer/features/lyrics/lyric-line.tsx
@@ -11,14 +11,13 @@ interface LyricLineProps extends ComponentPropsWithoutRef<'div'> {
 
 const StyledText = styled(TextTitle)<TitleProps & { $alignment: string; $fontSize: number }>`
     color: var(--main-fg);
-    font-weight: 400;
+    font-weight: 600;
     text-align: ${(props) => props.$alignment};
     font-size: ${(props) => props.$fontSize}px;
     opacity: 0.5;
     padding: 0 1rem;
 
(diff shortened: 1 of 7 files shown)


jeffvli:
Code looks fine. I pushed a couple adjustments to the border-radius and opacity to make the backgrounds look a bit more natural.

Also pushed a lyrics adjustment for #233.

kgarner7:
Looks great!

PR #239, opened 2023-09-09
Add ability to save/restore queue

kgarner7:
Uses `/rest/savePlayQueue` and `/rest/getPlayQueue`
Things that could be improved:
- putting the save/restore buttons somewhere more sensible?
- better warning/recommendation for too large uri (large queues will suffer)
- restore with current time as well (I couldn't figure out how to se the current time after adding to queue without setTimeout, which is not great)
- (maybe?) support remote save/restore from server, as well as local save/restore.


Programie:
I just came across this while searching for an already existing issue before requesting such a feature.

As far as I understand, the current implementation of this feature in this PR is to allow the user to manually store and restore the queue using buttons? Why not simply retrieve the queue on startup and sync it periodically so other clients can retrieve it from the server as well?

I just saw such a feature in [Tempo](https://github.com/CappielloAntonio/tempo) where the sync is automatically done every few seconds. Maybe that could be used for this feature in Feishin, too?

flightmansam:
> I just came across this while searching for an already existing issue before requesting such a feature.
> 
> As far as I understand, the current implementation of this feature in this PR is to allow the user to manually store and restore the queue using buttons? Why not simply retrieve the queue on startup and sync it periodically so other clients can retrieve it from the server as well?
> 
> I just saw such a feature in [Tempo](https://github.com/CappielloAntonio/tempo) where the sync is automatically done every few seconds. Maybe that could be used for this feature in Feishin, too?

I think this is a great idea!

Programie:
OK, there was a bit of misunderstanding from my side about the implementation in Tempo.

Actually, Tempo does automatically sync the queue from the server (`/rest/getPlayQueue`) in a configured interval. But it does not automatically push the queue to the server (`/rest/savePlayQueue`) for some reason.

I would expect to push the queue to the server periodically and pull it from the server on start of the app (probably also ask the user whether the pulled queue should be used in case it's a different one, i.e. created by a different application).

But in any case, it should be possible to configure whether the server queue should be used at all. Some users might want to use separate queues on different devices. I'm one of those users wanting to have one queue shared across different devices. For example, I want to listen to some music on my computer and continue listening to it on my phone.

And the best way to implement that would be to make it seamless user experience without requiring the user to first push a button to actually push the queue to the server and pull it on the other device. Maybe I forgot to push the queue on my computer.

PR #243, opened 2023-09-11
Replaygain support for Web Player

kgarner7:
This PR adds ReplayGain support for web player. Notes:
- uses WebAudio (which is [not supported in all browsers](https://caniuse.com/?search=audiocontext))
- WebAudio and MediaSession to not behave well, so image/title display may not behave well [Github Issue](https://github.com/w3c/audio-session-detach/issues/10)
- ~~**PLEASE** test this in Safari, when I added this to Navidrome there was an issue on MacOS~~ This _should_ be good now (I have a personal deployment of it which behaves properly now), but still worth checking 
- While this PR does support setting the sampling rate, I couldn't figure out how to cleanly support configuring that without a refresh
- It might also be nice to toggle RG from the full-screen player, but the menu is getting kinda full, so I'm going to refrain from adding anything more there


kgarner7:
A sample rate that can be configured without a refresh is unlikely (at least with `createMediaElementSource`) for a number of reasons:
- to change the sample rate, you would need to delete and create a new audio context
- this would then require connecting the audio element to this new context
- However, per `https://github.com/WebAudio/web-audio-api/issues/1202`, you cannot disconnect a media source from source node. This means that the one you first created is the one you're stuck with
- HOWEVER, you cannot attach nodes created from one context to another context.

It might be possible to do through the more roundabout way, but I'm not sure that makes sense.

PR #245, opened 2023-09-11
Added docker image build script

Kakoluz:
This PR adds the becesary files to build and serve a production ready docker image that uses the web version behind an nginx server. It is perfect for self hosted people like me. The build script is ready to be just ran and the image uploaded to as a "official" feishin docker image. the built image is around 200MB in size, 90MB compressed. Wich makes this a really light app.

It is note worthy that it might be necesary to change the port from 80 to something else as this settings are intended to be ran behind a proxy and not directly as a service, but some people might want to do it that way.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,17 @@
+# --- Builder stage
+FROM node:18-alpine as builder
+WORKDIR /app
+COPY . /app
+
+# Scripts include electron-specific dependencies, which we don't need
+RUN npm install --legacy-peer-deps --ignore-scripts
+RUN npm run build:web
+
+# --- Production stage
 FROM nginx:alpine-slim
-COPY release/app/dist/web /usr/share/nginx/html
+
+COPY --from=builder /app/release/app/dist/web /usr/share/nginx/html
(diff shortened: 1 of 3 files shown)


jeffvli:
Looks good! Have you tested any other base images (like alpine) to see if you can get it even smaller?

Kakoluz:
Not really, i will test it later and report back, i will also change the port to something else

BlackDragon-B:
>  i will also change the port to something else
Wouldn't you eliminate the need for this by using port bindings in docker?

Kakoluz:
> > i will also change the port to something else
> > Wouldn't you eliminate the need for this by using port bindings in docker?

Yes, but deploying a web app on port 80 by default and expect the user to change it. is not exactly a good practice

Kakoluz:
Using slim alpine nginx i managed to get the app down to around 20MB uncompressed. I also changed the port to something not reserved and somewhat easy to remember.

jeffvli:
I pushed a commit to update the dockerfile to use a multi-stage build so we can build the app directly, otherwise the dist files in `/release/app/dist/web`  would need to be built manually.

Issue #242, opened 2023-09-10
User MPV scripts may interfear with the program


137-Trimethylxanthin:
## Expected Behavior

it should play the music.

## Current Behavior

It is stuck at 0:00 and the terminal gives some errors if you have it open via termianl
from my limited understanding of mpv, i think it crashes at the start
(replaced username with %user%)
at start:
```powershell
Error in main process TypeError: Error processing argument at index 0, conversion failure from mediatracknext
    at IpcMainImpl.<anonymous> (C:\Users\%user%\AppData\Local\Programs\feishin\resources\app.asar\dist\main\main.js:14:522585)
    at IpcMainImpl.emit (node:events:513:28)
    at EventEmitter.<anonymous> (node:electron/js2c/browser_init:2:81053)
    at EventEmitter.emit (node:events:513:28)
Initializing MPV with data:  {
  extraParameters: [ '' ],
  properties: {
    'audio-exclusive': 'no',
    'gapless-audio': 'weak',
    replaygain: 'track',
    'replaygain-clip': true,
    'replaygain-preamp': 0
  }
}
Setting mpv params:  [ '--idle=yes', '--prefetch-playlist=yes', '' ]
Setting MPV properties:  {
  'audio-exclusive': 'no',
  'gapless-audio': 'weak',
  replaygain: 'track',
  'replaygain-clip': true,
  'replaygain-preamp': 0
}
(node:12904) UnhandledPromiseRejectionWarning: #<Object>
(Use `Feishin --trace-warnings ...` to show where the warning was created)
(node:12904) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). To terminate the node process on unhandled promise rejection, use the CLI flag `--unhandled-rejections=strict` (see https://nodejs.org/api/cli.html#cli_unhandled_rejections_mode). (rejection id: 1)
Error in main process Error: Socket is closed
    at new NodeError (node:internal/errors:387:5)
    at Socket._writeGeneric (node:net:864:8)
    at Socket._write (node:net:886:8)
    at writeOrBuffer (node:internal/streams/writable:391:12)
    at _write (node:internal/streams/writable:332:10)
    at Socket.Writable.write (node:internal/streams/writable:336:10)
    at C:\Users\%user%\AppData\Local\Programs\feishin\resources\app.asar\dist\main\main.js:14:343343
    at new Promise (<anonymous>)
[...]

Issue #234, opened 2023-09-01
User MPV configuration may interfere with Feishin


ghost:
## Expected Behavior

An MPV window doesn't appear and instead plays in the background

## Current Behavior

MPV window appears due to the configuration that I use for MPV when playing back media when using MPV standalone

## Steps to Reproduce

Place a file in ~/.config/mpv/mpv.conf with the contents 
```
force-window=immediate
```

## Possible Solution (Not obligatory)

Run mpv with --no-config or specify a custom configuration directory (--config-dir) so that users can create a custom configuration for MPV with Feishin if they like, or add their own scripts.

## Context

Spent too much time figuring out what was the issue

## Your Environment

Feishin with MPV on Linux

PR #247, opened 2023-09-15
Prevent MPV from loading user config/scripts (#234 & #242)

iiPythonx:
This is a very quick fix which just disables mpv loading the users configuration file and local scripts.  
Tested with the same scenario from #234.


jeffvli:
Thanks! In the future, this should probably be configurable so that users can choose to use their own configuration if they want.

iiPythonx:
@jeffvli Agreed! I just wanted to temporarily patch these issues for the time being. Thanks for merging it.

PR #268, opened 2023-09-28
Add Server Rescan

kgarner7:
Resolve #121.
Notes:
- Jellyfin won't report scan progress
- Navidrome/Subsonic will
- `fullScan` is not actually a formal subsonic parameter, Navidrome appears to have added it
- There's probably a better place to put the rescan button, but I don't know


jeffvli:
I'm going to keep this on hold for now.  Rescan being in the sidebar as an item feels a bit awkward, and I want to make sure any new features are added with `ServerFeatures` in mind and written without hardcoding server types for the implementation.

PR #283, opened 2023-10-07
add more emphasis to current song

### LINKED ISSUE #237: Highligh the current playing song more
Currently, only the text of the current song title is changed to blue. It should be more obvious which song is being played in the queue.

kgarner7:
closes #237; adds background and primary-color border

### DIFF
--- a/.erb/configs/webpack.config.remote.dev.ts
+++ b/.erb/configs/webpack.config.remote.dev.ts
@@ -9,112 +9,119 @@ import checkNodeEnv from '../scripts/check-node-env';
 import baseConfig from './webpack.config.base';
 import webpackPaths from './webpack.paths';
 
+const { version } = require('../../package.json');
+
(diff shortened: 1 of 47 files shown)


jeffvli:
I'm not a fan of the border personally. It conflicts a bit with some other app borders such as when the queue it's in the sidebar.

I also want to enhance this with an animated play icon similar to Spotify's.

kgarner7:
Fair enough! I was thinking about adding an indicator as well, but realized that would require another column. I can revert the border change and add a "now playing" column

jeffvli:
I would say instead of a new column, it could be added to either:
- `row index` (replaces the number of the currently playing track)
- `title` or `title (combined)` (added to the right of the song name).

kgarner7:
Row index _probably_ makes more sense, but if that is the choice, then it should probably be included by default in the full-screen play queue.

PR #284, opened 2023-10-07
[bugfix]: MPV do not next/prev on queue set

kgarner7:
When calling setQueue, do not successively call next()/prev(). Somehow, prev() works, but for next() there is a race condition where after the first skip you will skip two tracks.


jeffvli:
I re-added the await in https://github.com/jeffvli/feishin/commit/ca39409cc38be73bd7c5604ea4cb6f55234af67c which fixes the issue without having to alter the existing player controls.

PR #293, opened 2023-10-12
Replace hardcoded Jellyfin authentication DeviceId to include hostname

LilyRose2798:
Fixes bug preventing two separate users on the same Jellyfin instance from being logged in simultaneously due to conflicting DeviceIds. See [here](https://gist.github.com/nielsvanvelzen/ea047d9028f676185832e51ffaf12a6f#device-identifiers) for information on DeviceIds. If two clients try to authenticate using the same device id then the other will have their session revoked.

### DIFF
--- a/src/renderer/features/albums/components/navidrome-album-filters.tsx
+++ b/src/renderer/features/albums/components/navidrome-album-filters.tsx
@@ -152,11 +152,11 @@ export const NavidromeAlbumFilters = ({
             customFilters,
             data: {
                 _custom: {
+                    ...filter._custom,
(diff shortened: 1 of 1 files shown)


LilyRose2798:
It may also be worth including the username as well as hostname. Can submit an updated version if needed.

lucyrose39:
This is good fix

This exact issue presents itself on my own jellyfin server

jeffvli:
@LilyRose2798 Including username would be nice, especially if duplicate device ids affect existing sessions.

PR #304, opened 2023-10-14
[feature]: Support using system fonts

### LINKED ISSUE #270: [Feature Request] Support adding custom fonts
Apologies if this is already possible (wouldn't mind diving into files), but it would be cool if you could add a custom font for content, or if you could select from all fonts installed on your system.

### LINKED ISSUE #288: Cyrillic font support
Support for cyrillic fonts. Cyrillic texts revert to a standard serif font.

kgarner7:
Uses the **experimental** queryLocalFonts API, when prompted, to get the fonts and do CSS. Resolves #270 and also resolves #288 (by proxy)

Caveats/notes:
- This is experimental, and is only supported by Chrome/Chromium/Edgeium (see https://caniuse.com/?search=querylocalfonts)
- As far as I can tell, the only way to dynamically change the font (shown in https://wicg.github.io/local-font-access/#example-style-with-local-fonts) was by DOM manipulation; css variables did not seem to work
- This shows **all** fonts, including their variants (bold/italic/etc); given that the style names could be localized, not sure of a way to parse this (on my system, for instance, I had 859 different combinations)
- I made fonts a separate top-level setting because it was easier to manipulate without causing as many rerenders; feel free to put that back

### DIFF
--- a/src/main/main.ts
+++ b/src/main/main.ts
@@ -21,6 +21,8 @@ import {
     Menu,
     nativeImage,
     BrowserWindowConstructorOptions,
+    protocol,
+    net,
 } from 'electron';
 import electronLocalShortcut from 'electron-localshortcut';
 import log from 'electron-log';
@@ -42,6 +44,8 @@ export default class AppUpdater {
         autoUpdater.checkForUpdatesAndNotify();
     }
 }
+
+protocol.registerSchemesAsPrivileged([{ privileges: { bypassCSP: true }, scheme: 'feishin' }]);
 
 process.on('uncaughtException', (error: any) => {
     console.log('Error in main process', error);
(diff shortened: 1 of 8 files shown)


Nik-mmzd:
Is it possible to add option to use "System default" font instead of choosing from every font installed? Or this option is already included to this PR?
After all, most operating systems allow the user to select a default font, and this font usually contains all the necessary characters (e.g. Cyrillic characters).

kgarner7:
As far as I can tell, there's no way for the browser to tell this. Draft spec here: https://wicg.github.io/local-font-access/#dom-window-querylocalfonts.

Nik-mmzd:
Perhaps it would be better to give the user the option to "upload" a custom font(s) to the app rather than querying all installing fonts using unstable API?

kgarner7:
Since the font api has been in chrome since 103 (last year), I'd say it's stable there. I've added an **Electron-only** workaround that will allow providing a custom font by adding a custom protocol (`feishin`) that will proxy font files.

PR #307, opened 2023-10-14
[feature]: support running feishin on custom path

kgarner7:
Support serving Feishin on a subpath, and change webpack to autodetect the subpath as opposed to hardcoding `/`. This can be changed dynamically with an environment variable.
Note for users, that `/path/` and `/path` will see different results.


jeffvli:
Could you also add documentation regarding this on the README?

PR #312, opened 2023-10-17
[bugfix]: Replace hardcoded Jellyfin authentication DeviceId to include hostname and username

LilyRose2798:
Fixes bug preventing two separate users on the same Jellyfin instance from being logged in simultaneously due to conflicting DeviceIds. See [here](https://gist.github.com/nielsvanvelzen/ea047d9028f676185832e51ffaf12a6f#device-identifiers) for information on DeviceIds. If two clients try to authenticate using the same device id then the other will have their session revoked.

### DIFF
--- a/src/renderer/api/jellyfin/jellyfin-controller.ts
+++ b/src/renderer/api/jellyfin/jellyfin-controller.ts
@@ -100,7 +100,7 @@ const authenticate = async (
             Username: body.username,
         },
         headers: {
-            'x-emby-authorization': `MediaBrowser Client="Feishin", Device="${getHostname()}", DeviceId="Feishin", Version="${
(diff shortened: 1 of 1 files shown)


LilyRose2798:
Recreated #293 with added username in DeviceId

jeffvli:
Thanks!

Also just for the future -- you don't need to recreate the pull request when pushing changes to a branch that already has a pull request open. It will automatically update!

LilyRose2798:
Yeah I tried to just update the PR by pushing to the branch, but it seems like it automatically closed itself after I did a `git reset --hard upstream/development`, recreated the commit, and did a `git push -f origin` to overwrite the branch history. There was probably a better way to do it but I didn't want a messy commit history and I'm not the greatest at using git.

PR #314, opened 2023-10-19
[very rough draft]: add web visualizer

kgarner7:
Adds a audio visualizer for web player. This is a very rough draft; there are probably better visualizers/maybe multiple visualizers, but I personally am fond of this one.

Notes:
- Uses my fork, because https://github.com/foobar404/wave.js/pull/92 is necessary
- Only for web; I'm not sure of any way to capture the audio from MPV (and I don't believe you can embed it in Electron either, at least not easily)
- Not sure if it needs a flashing lights warning?


kgarner7:
I've elected to fall back into custom implementation, for the following reasons:
- the data returned from `getByteFrequencyData` represents [liner scale](https://developer.mozilla.org/en-US/docs/Web/API/AnalyserNode/getFloatFrequencyData). This uses exponential scaling
- The frequency bins depends on the current frequency. For example, if you set 96 kHz, the returned data is [0, 48 kHz]. This PR only shows between [16, 20k] (so higher sample rates means less bins)
- Reduce update rate (50-60 ms interval seem fine, others were using `requestAnimationFrame`, which happens more often). I guess you could use `requestAnimationFrame` and then using the timestamp

PR #317, opened 2023-10-20
[bugfix/feat]: always fetch artist image for Navidrome

### LINKED ISSUE #313: [Navidrome] No Artist images displayed on 0.3.0 and above
## Expected Behavior
Album Artist images should be displayed both on the Album Artist grid and Artist pages

## Current Behavior
Album Artist images are not displayed at all on versions before or after 0.2.0
on 0.2.0 they are only displayed on the individual artist page, and not the grid

## Steps to Reproduce
Connect any version other than 0.2.0 to a Navidrome server

Observe the issue either the Album Artist grid or individual Artist pages

## Context
Images are stored locally on the Navidrome Server (inside the artist folder if that matters)

Images work on: Sonixd, Navidrome WebUI and the Substreamer Android app

https://github.com/jeffvli/feishin/issues/90 might be relevant as the symptoms are identical on versions other than 0.2.0
(though there is a new placeholder image on the grid page)

## Your Environment

-   **Application version:** Tested on all versions between 0.0.1-alpha6 and 0.4.1
-   **Operating System:** ArchLinux x86_64 Kernel 6.5.7-arch1-1
-   **Desktop Environment:** SwayWM (Wayland)
-   **Server:** Navidrome Docker [deluan/navidrome:develop] [0.49.3-SNAPSHOT] (59f0c487)

kgarner7:
Resolves #313. Removes check for album artist image matching JWT, and always attempts to fetch album artist image in list (`ar-${artistId}` is a Navidrome-specific thing for `getCoverArt`). 

Unfortunately Navidrome doesn't expose whether the artist image is real or not using the native API; maybe it would be good to have a toggle to select whether to try and fetch all images?

### DIFF
--- a/src/renderer/api/navidrome/navidrome-normalize.ts
+++ b/src/renderer/api/navidrome/navidrome-normalize.ts
@@ -20,10 +20,6 @@ const getImageUrl = (args: { url: string | null }) => {
         return null;
     }
 
-    if (url?.match('eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9')) {
-        return null;
-    }
-
     return url;
 };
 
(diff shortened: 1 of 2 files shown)


jeffvli:
Adding a toggle just increases complexity in my opinion. 

I added an error check for the library header image so it won't just show an errored image.

PR #332, opened 2023-10-28
[bugfix/feature]: Improve ratings

kgarner7:
Fix: add preventDefault/stopPropagation to prevent scrolling to top in queue
Feat: instead of double click for clear, click on same value


src/renderer/components/rating/index.tsx
```diff
@@ -16,17 +13,37 @@ const StyledRating = styled(MantineRating)`
...
+        },
+        [onChange, props.value],
+    );
+
+    const debouncedOnChange = debounce(valueChange, 100);
 
     return (
         <Tooltip
-            label="Double click to clear"
+            label="Click on the same value to clear"
```
jeffvli: Label can be removed now since this is the typical functionality for the rating component. I only added the label previously because double click is non-standard.

PR #334, opened 2023-10-29
[bugfix]: prevent default 

kgarner7:
.-.; as it turns out that `preventDefault` was important with the `ToolTip`. This is necessary to prevent jumping in play queue. 
As far as I can tell, this is the only component with that issue, but if there are any other actions in the future, then `preventDefault`/`stopPropagation` is required. Sorry for second pr on this

### DIFF
--- a/package.json
+++ b/package.json
@@ -93,6 +93,39 @@
             ],
             "icon": "assets/icons/icon.ico"
         },
+        "deb": {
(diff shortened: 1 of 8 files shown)


jeffvli:
Oh yeah, didn't realize that line got removed.

PR #344, opened 2023-11-01
[bugfix]: Remove setting path in attempt to bypass bug

### LINKED ISSUE #343: Settings crash after setting custom system font
## Expected Behavior
Expected **Hamburger menu > Settings** to open the settings

## Current Behavior
Settings crashes, so I can't change any settings any more! Doesn't work on reload, either.
Oddly, the custom font I picked (SF Mono) _does_ load correctly.

Screenshot 2023-10-31 at 12 46 39

## Steps to Reproduce

1. Open the Settings menu.
2. In **General**, pick **System font**
3. Choose a font from the macOS Fonts folder.
4. Observe the app crash.
5. Try to open the settings menu again. Observe the app crash.
6. Try to reload the settings menu. Still an app crash.

## Context

I like a monospace font for my music player, so I tried picking SF Mono. It does look pretty nice, but I think I'd prefer being able to use my settings menu.

## Your Environment

-   Application version (e.g. v0.1.0) : `0.5.0`
-   Operating System and version (e.g. Windows 10) : macOS Sonoma
-   Server and version (e.g. Navidrome v0.48.0) : Jellyfin 10.8.9
-   Node version (if developing locally) : not developing locally

kgarner7:
Maybe resolves #343? I can't verify because the issue doesn't happen for me, but this appears to be the symptom at least...

### DIFF
--- a/src/renderer/features/settings/components/general/application-settings.tsx
+++ b/src/renderer/features/settings/components/general/application-settings.tsx
@@ -68,9 +68,9 @@ export const ApplicationSettings = () => {
 
     const fontList = useMemo(() => {
         if (fontSettings.custom) {
-            return new File([], fontSettings.custom.split(/(\\|\/)/g).pop()!);
+            return fontSettings.custom.split(/(\\|\/)/g).pop()!;
         }
-        return null;
+        return '';
     }, [fontSettings.custom]);
 
     const onFontError = useCallback(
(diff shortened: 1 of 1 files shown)


jeffvli:
Yeah, that seems to be it.

If this doesn't fix it, you could do what I did with the mpv executable path and just remove the `defaultValue` and set the raw path string as a placeholder.

kgarner7:
Placeholder is probably a better option anyway, so let's go with that

Issue #380, opened 2023-11-14
Configure a fixed server for the docker web app


haath:
Hey all,

To make it easier for my users, I'd like to set up a Feishin front-end that is configured to already have the URL of my Jellyfin server.
Then users visiting the Feishin web app would only be prompted to enter their Jellyfin credentials and log in.

In my opinion the current `Open Menu > Manage servers` flow is a bit redundant, and could be streamlined away when the intent is just to use one frontend with one backend.

On the Docker side, this could easily be configured by setting some `JELLYFIN_URL` or `NAVIDROME_URL` environment variable.

stefan1983:
+1 

I ran into the same as well, when connecting from different network devices.

jeffvli:
I did some preliminary work on getting this done in branch [feature/#380](https://github.com/jeffvli/feishin/tree/feature/%23380). I can't seem to get the environment variables to pass from docker into the production build though.

If someone is familiar with this process can they take a look?

kgarner7:
Given that Dockerized NGINX supports templating, what's probably the easiest solution is to just have the web version serve an additional JavaScript file, generated by NGINX template itself.
```nginx
location ${PUBLIC_PATH}/settings.js {
    add_header Content-Type application/javascript;
    return '"use strict";window.SERVER_URL="${SERVER_URL}";'
}
```
You could then tweak the `index.ejs` file to include this script first for web builds (and then use `process.env.SERVER_URL`, and fallback to `window.SERVER_URL`)

Kakoluz:
I started working on this feature on my own a short while ago and left it somewhat hung. I have days off work this week and i will look into it. But the solution @kgarner7 proposes seems reasonable. its just matter of setting the docker image to replace envs with those passed to the container. Might also look into possible alternatives. Expect a PR related to this soonish.

Angablade:
If you are using Nginx as a reverse proxy in front of everything, you can do something real simple to do this.

`	location = /renderer.js {
			sub_filter_once on;
			sub_filter_types *;
			sub_filter '({initialValues:{legacyAuth:!1,name:"",password:"",savePassword:!1,type:v.ServerType.JELLYFIN,url:"http://",username:""}})' '({initialValues:{legacyAuth:!1,name:"My Music",password:"",savePassword:!1,type:v.ServerType.JELLYFIN,url:"https://example.co.uk",username:""}})';
			proxy_set_header Forwarded $proxy_add_forwarded;
[...]

PR #413, opened 2023-12-18
Add a pre-defined server for the docker version

Kakoluz:
This PR solves #380. Allows for a pre defined server configuration to be setted as environments in the docker image. I also added a locked state for the UI so user won't be allowed to change the connection details to connect to other servers.

I also took the liberty to fix and optimize the docker build process for it to take advantage of node_modules caching and removed an unused compilation when calling npm run build:docker (i think this one was my fault)

My familiarity with React is nearly 0, so i might have fucked up something or may not compy with some standard. Feel free to modify any code that's off or incorrect. My development environment is not adequate for testing electron builds, only the web version. It might be broken I can't test.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -14,8 +14,10 @@ RUN npm run build:web
 FROM nginx:alpine-slim
 
 COPY --chown=nginx:nginx --from=builder /app/release/app/dist/web /usr/share/nginx/html
+COPY ./settings.js.template /settings.js.template
+RUN envsubst < /settings.js.template > /usr/share/nginx/html/settings.js
 COPY ng.conf.template /etc/nginx/templates/default.conf.template
 
 ENV PUBLIC_PATH="/"
 EXPOSE 9180
-CMD ["nginx", "-g", "daemon off;"]
+CMD ["/bin/sh", "-c", "envsubst < /settings.js.template > /usr/share/nginx/html/settings.js && envsubst '${PUBLIC_PATH}' < /etc/nginx/templates/default.conf.template > /etc/nginx/conf.d/default.conf && nginx -g 'daemon off;'"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -14,9 +14,9 @@ RUN npm run build:web
 FROM nginx:alpine-slim
 
 COPY --chown=nginx:nginx --from=builder /app/release/app/dist/web /usr/share/nginx/html
-COPY ./settings.js.template /settings.js.template
+COPY ./settings.js.template /etc/nginx/templates/settings.js.template
 COPY ng.conf.template /etc/nginx/templates/default.conf.template
 
 ENV PUBLIC_PATH="/"
 EXPOSE 9180
-CMD ["/bin/sh", "-c", "envsubst < /settings.js.template > /usr/share/nginx/html/settings.js && envsubst '${PUBLIC_PATH}' < /etc/nginx/templates/default.conf.template > /etc/nginx/conf.d/default.conf && nginx -g 'daemon off;'"]
+CMD ["nginx", "-g", "daemon off;"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -15,7 +15,6 @@ FROM nginx:alpine-slim
 
 COPY --chown=nginx:nginx --from=builder /app/release/app/dist/web /usr/share/nginx/html
 COPY ./settings.js.template /settings.js.template
-RUN envsubst < /settings.js.template > /usr/share/nginx/html/settings.js
 COPY ng.conf.template /etc/nginx/templates/default.conf.template
 
 ENV PUBLIC_PATH="/"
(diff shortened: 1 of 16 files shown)


src/renderer/features/servers/components/add-server-form.tsx
```diff
@@ -33,11 +33,11 @@ export const AddServerForm = ({ onCancel }: AddServerFormProps) => {
     const form = useForm({
         initialValues: {
             legacyAuth: false,
-            name: '',
+            name: !isElectron() ? window.SERVER_NAME : process?.env.SERVER_NAME?? '' ,
```
kgarner7: `process` is not available because `contextIsolation` is true. Consider instead exposing the environment variables in `preload.ts`, which can then be available to the renderer. As-is, this breaks the desktop app

src/renderer/index.ejs
```diff
@@ -6,6 +6,7 @@
   <meta http-equiv="Content-Security-Policy" />
   <meta name="viewport" content="width=device-width, initial-scale=1">
   <title>Feishin</title>
+  <script src="settings.js"></script>
```
kgarner7, line 10: This should only be loaded for `web` target (it's not useful for desktop Feishin). See https://github.com/jeffvli/feishin/pull/387/files for webpack changes you can do

src/renderer/features/servers/components/add-server-form.tsx
```diff
@@ -33,11 +33,11 @@ export const AddServerForm = ({ onCancel }: AddServerFormProps) => {
...
             password: '',
             savePassword: false,
-            type: ServerType.JELLYFIN,
-            url: 'http://',
+            type: !isElectron() ? window.SERVER_TYPE?.toLowerCase() : process?.env.SERVER_TYPE?.toLowerCase() ?? ServerType.JELLYFIN,
```
kgarner7: Rather than just accepting a string, should probably check is this string one of the valid types (currently just `ServerType.JELLYFIN` and `ServerType.NAVIDROME`). You can even define a function that takes a string and returns a valid server type (Navidrome/Jellyfin) if the string is valid, or defaults to Jellyfin, that way you don't need to type check
```typescript
const getServerType = (value?: string) => {
    switch (value?.toLowerCase()) {
        case ServerType.NAVIDROME:
            return ServerType.NAVIDROME;
        default:
            return ServerType.JELLYFIN;
    }
};
```
(an example, there are other ways). This prevents having to do `as ServerType` later

Dockerfile
```diff
@@ -1,18 +1,22 @@
...
 FROM nginx:alpine-slim
 
 COPY --chown=nginx:nginx --from=builder /app/release/app/dist/web /usr/share/nginx/html
+COPY ./settings.js.template /settings.js.template
 COPY ng.conf.template /etc/nginx/templates/default.conf.template
 
 ENV PUBLIC_PATH="/"
 EXPOSE 9180
-CMD ["nginx", "-g", "daemon off;"]
+CMD ["/bin/sh", "-c", "envsubst < /settings.js.template > /usr/share/nginx/html/settings.js && envsubst '${PUBLIC_PATH}' < /etc/nginx/templates/default.conf.template > /etc/nginx/conf.d/default.conf && nginx -g 'daemon off;'"]
```
kgarner7: The NGINX container already supports envsubst automatically (`NGINX_ENVSUBST_TEMPLATE_DIR`, `/etc/nginx/templates`), maybe that would be cleaner?

kgarner7 (changes requested):
Thanks for your work so far! Please ask if you have any questions about the comments.

Dockerfile
Kakoluz: I did this quite a time ago, but if i remember correctly i was having problems with multiple variables replacing at the same time and leaving a mess so it wouldn't work. This was the solution i found on the spot.

kgarner7: Ugh, wrong button press. Sorry for the likely three emails.

You can resolve this by adding something like 
```nginx
  location ${PUBLIC_PATH}settings.js {
    alias /etc/nginx/conf.d/settings.js;
  }
```
to the `ng.conf.template` and then in the Dockerfile doing
```Dockerfile
COPY settings.js.template /etc/nginx/templates/settings.js.template
```

Kakoluz: I will try, but if i recall correctly it was on the nginx.conf file where it was replacing the envs twice resulting in some gibberish. Will try anyway

kgarner7:
I've made the necessary changes to support `setting.js.template` automatically and fix environment variable.

Kakoluz:
Is there anything left to change? I had some uncommited changes on my pc that fixed most of the stuff but i haven't found the time to finish everything.

kgarner7:
I don't think so? Since I made some changes myself though, I'd like @jeffvli to review it as well (but it looks good from my point of view). Sorry for just coming in and making changes

PR #449, opened 2024-01-15
Use startIndex in Jellyfin getPlaylistSongList

SlyFabi:
Fixes Jellyfin playlists with more than 300 songs.


jeffvli:
Going to merge this, but there have been some changes to playlist fetching in #419 to support Subsonic which mostly makes this obsolete. Subsonic doesn't support paginated requests on playlists so the logic was changed to fetch the entire playlist on one request rather than using pagination.

PR #450, opened 2024-01-16
[feature]: Show album comment, Last.fm/MusicBrainz links

kgarner7:
A few small things to show some additional metadata for albums/artists. Inspired by Navidrome.

- Album pages will now show safely-escaped (only add links) comment (Navidrome). Comment parsing for Note cell is also updated to use safer parsing.
- Jellyfin will fetch ProviderIds (MusicBrainz ID) for album/artist page
- Jellyfin and Navidrome show lastfm and Musicbrainz (if tagged) links, if enabled

### DIFF
--- a/src/renderer/features/albums/components/album-detail-content.tsx
+++ b/src/renderer/features/albums/components/album-detail-content.tsx
@@ -1,7 +1,7 @@
 import { MutableRefObject, useCallback, useMemo } from 'react';
 import { RowDoubleClickedEvent, RowHeightParams, RowNode } from '@ag-grid-community/core';
 import type { AgGridReact as AgGridReactType } from '@ag-grid-community/react/lib/agGridReact';
-import { Box, Group, Spoiler, Stack } from '@mantine/core';
+import { Box, Group, Stack } from '@mantine/core';
 import { useSetState } from '@mantine/hooks';
 import { useTranslation } from 'react-i18next';
 import { FaLastfmSquare } from 'react-icons/fa';
@@ -12,7 +12,7 @@ import { Link } from 'react-router-dom';
 import styled from 'styled-components';
 import { queryKeys } from '/@/renderer/api/query-keys';
 import { AlbumListSort, LibraryItem, QueueSong, SortOrder } from '/@/renderer/api/types';
-import { Button, Popover } from '/@/renderer/components';
+import { Button, Popover, Spoiler } from '/@/renderer/components';
 import { MemoizedSwiperGridCarousel } from '/@/renderer/components/grid-carousel';
 import {
     TableConfigDropdown,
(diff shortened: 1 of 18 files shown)


src/renderer/features/albums/components/album-detail-content.tsx
```diff
@@ -395,6 +403,58 @@ export const AlbumDetailContent = ({ tableRef, background }: AlbumDetailContentP
...
+                                </Button>
+                            ) : null}
+                        </Group>
+                    </Box>
+                ) : null}
+                {comment && (
+                    <Box component="section">
+                        <Spoiler
+                            hideLabel={t('common.collapse')}
+                            maxHeight={60}
```
jeffvli: Let's change `maxHeight` to 75. At 60, it cuts off at 2.5 lines so it shows half of the next line which looks a bit off. With 75 it'll cut off at 3 full lines.

src/renderer/features/albums/components/album-detail-content.tsx
```diff
@@ -395,6 +403,58 @@ export const AlbumDetailContent = ({ tableRef, background }: AlbumDetailContentP
...
+                            mb={20}
```
jeffvli: Let's adjust styling of all components that are using inline margin. Right now if an album has external links but no comment, the external links won't have any margin between it and the table header.

Remove `py="1rem"` from the play bar and genres sections, and then just add a `gap: 1rem` to the `DetailContainer` component so that we have proper spacing between all sections regardless of how many sections there are.

src/renderer/store/settings.store.ts
```diff
@@ -281,6 +282,7 @@ const initialState: SettingsState = {
     general: {
         accent: 'rgb(53, 116, 252)',
         defaultFullPlaylist: true,
+        externalLinks: false,
```
jeffvli: Default this to `true`. I'd say it's better to have users have to opt-out here rather than opt-in since people may not look through every setting.

src/renderer/features/artists/components/album-artist-detail-content.tsx
```diff
@@ -411,6 +417,50 @@ export const AlbumArtistDetailContent = ({ background }: AlbumArtistDetailConten
                         </Group>
                     </Box>
                 ) : null}
+                {externalLinks ? (
+                    <Box
+                        component="section"
+                        mb={-20}
+                        mt={-20}
```
jeffvli: Same thing as the AlbumDetailContent comments. Let's normalize the styling to use flex gap between all the sections instead of using inline padding/margin.

isle9:
Is the "Read more on Last.fm" link at the end of the artist description still necessary if we now have buttons for both last.fm and musicbrainz? That link just opens the artist page on last.fm.

PR #480, opened 2024-01-31
[bugfix]: Fix repeated track scrobbling

### LINKED ISSUE #364: Doesn't scrobble songs when they are on repeat
## Expected Behavior

When the song ends and repeat, it should be scrobbled.

## Current Behavior

Songs restart but doesn't scrobbles, the numbers of plays stays the same.
image

## Steps to Reproduce

1. Play a song.

2. Put it on repeat (repeat one).

3. When the songs repeat, it won't scrobble to the media server.

N/A

## Context

Trying to scrobble my plays to my media server when I'm repeating songs.

## Your Environment

-   Application version (e.g. v0.1.0) : 0.5.0
-   Operating System and version (e.g. Windows 10) : Linux (docker)
-   Server and version (e.g. Navidrome v0.48.0) : Navidrome v.0.49.3
-   Node version (if developing locally) :

kgarner7:
Resolves #364

This commit fixes scrobbling tracks that are repeated in the queue (either due to a REPEAT_ONE or just enqueued multiple times). There are two extra cases:

1. The current player changes. The player alternates from 1 -> 2 -> 1 whenever playback finishes normally, so this is a sort of epoch that can be used to detect both repeat one and track being enqueued multiple times.
2. The current index changes. The player gets set back to 1 in a variety of cases (most often previous/next), so the check in (1) is not enough. However, the index changing will help identify that this is a new song.

### DIFF
--- a/src/renderer/features/player/hooks/use-scrobble.ts
+++ b/src/renderer/features/player/hooks/use-scrobble.ts
@@ -315,29 +315,19 @@ export const useScrobble = () => {
 
     useEffect(() => {
         const unsubSongChange = usePlayerStore.subscribe(
-            (state) => [
-                state.current.song,
-                state.current.time,
-                state.current.player,
-                state.current.index,
-            ],
+            (state) => [state.current.song, state.current.time, state.current.player],
             handleScrobbleFromSongChange,
             {
                 // We need the current time to check the scrobble condition, but we only want to
                 // trigger the callback when the song changes
-                // There are three conditions where this should trigger:
+                // There are two conditions where this should trigger:
                 // 1. The song actually changes (the common case)
                 // 2. The song does not change, but the player dows. This would either be
                 //    a single track on repeat one, or one track added to the queue
                 //    multiple times in a row and playback goes normally (no next/previous)
-                // 3. The song does not change, but the queue position does. This would happen if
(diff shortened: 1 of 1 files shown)


src/renderer/features/player/hooks/use-scrobble.ts
```diff
@@ -315,12 +315,29 @@ export const useScrobble = () => {
...
-                equalityFn: (a, b) => (a[0] as QueueSong)?.id === (b[0] as QueueSong)?.id,
+                // There are three conditions where this should trigger:
+                // 1. The song actually changes (the common case)
+                // 2. The song does not change, but the player dows. This would either be
+                //    a single track on repeat one, or one track added to the queue
+                //    multiple times in a row and playback goes normally (no next/previous)
+                // 3. The song does not change, but the queue position does. This would happen if
+                //    the same song has been enqueued multiple times in a row, and the prev/next
+                //    button is pressed. In this case, the player is forced to 1, which may not cause
+                //    condition 2 to fire. Thus, comparing the queue index will catch this case
```
jeffvli: I think the better fix for condition 3 is changing the equalityFn to check the QueueSong by `uniqueId` instead of `id`. Each song added to the queue has a unique identifier that's appended onto the Song object so we wouldn't have to check if the index changed or not.

PR #486, opened 2024-02-02
[enhancement] Add a button size setting

iiPythonx:
Exactly what it says on the tin. The limits of 15px-30px were chosen from what seemed to actually fit within the player bar container on my monitor. 10px was the original lower limit but it looked wayyyy too small for anybody to actually use so I bumped it up to 15px.

First typescript related change, so if something isn't formatted right or if it just does something wrong in general, feel free to fix up or just ignore the PR.

### DIFF
--- a/src/i18n/locales/en.json
+++ b/src/i18n/locales/en.json
@@ -400,7 +400,7 @@
         "audioExclusiveMode_description": "enable exclusive output mode. In this mode, the system is usually locked out, and only mpv will be able to output audio",
         "audioPlayer": "audio player",
         "audioPlayer_description": "select the audio player to use for playback",
-        "buttonSize": "Button size",
+        "buttonSize": "Player bar button size",
         "buttonSize_description": "The size of the player bar buttons",
         "clearCache": "Clear browser cache",
         "clearCache_description": "A 'hard clear' of Feishin. In addition to clearing Feishin's cache, empty the browser cache (saved images and other assets). Server credentials and settings are preserved.",
--- a/src/i18n/locales/en.json
+++ b/src/i18n/locales/en.json
@@ -400,6 +400,8 @@
         "audioExclusiveMode_description": "enable exclusive output mode. In this mode, the system is usually locked out, and only mpv will be able to output audio",
         "audioPlayer": "audio player",
         "audioPlayer_description": "select the audio player to use for playback",
+        "buttonSize": "Button size",
+        "buttonSize_description": "The size of the player bar buttons",
         "clearCache": "Clear browser cache",
         "clearCache_description": "A 'hard clear' of Feishin. In addition to clearing Feishin's cache, empty the browser cache (saved images and other assets). Server credentials and settings are preserved.",
         "clearQueryCache": "Clear Feishin cache",
--- a/src/renderer/features/settings/components/general/control-settings.tsx
+++ b/src/renderer/features/settings/components/general/control-settings.tsx
@@ -39,7 +39,7 @@ export const ControlSettings = () => {
         {
             control: (
                 <NumberInput
-                    defaultValue={settings.buttonSize || 20}
+                    defaultValue={settings.buttonSize}
                     max={30}
                     min={15}
                     rightSection="px"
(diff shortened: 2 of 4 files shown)


src/renderer/features/settings/components/general/control-settings.tsx
```diff
@@ -36,6 +36,33 @@ export const ControlSettings = () => {
     const { setSettings } = useSettingsStoreActions();
 
     const controlOptions = [
+        {
+            control: (
+                <NumberInput
```
kgarner7, line 41: Maybe make the number input smaller (and also add `Px`) or something to the right? See `Sample rate`

src/renderer/store/settings.store.ts
```diff
@@ -282,6 +283,7 @@ const initialState: SettingsState = {
     },
     general: {
         accent: 'rgb(53, 116, 252)',
+        buttonSize: 20,
```
kgarner7, line 286: I _believe_ you should also update the version of the store, so that the changes take effect

src/i18n/locales/en.json
```diff
@@ -400,6 +400,8 @@
         "audioExclusiveMode_description": "enable exclusive output mode. In this mode, the system is usually locked out, and only mpv will be able to output audio",
         "audioPlayer": "audio player",
         "audioPlayer_description": "select the audio player to use for playback",
+        "buttonSize": "Button size",
+        "buttonSize_description": "The size of the player bar buttons",
```
kgarner7, line 404: If the units are not listed in the number input, they should probably be listed here

kgarner7 (changes requested):
@jeffvli may disagree with me, but I feel like it might also make sense to have the button sizes take effect for `right-controls`

src/renderer/features/settings/components/general/control-settings.tsx
iiPythonx: Sure, I'll see about adding it there rather then in the language file.

src/renderer/store/settings.store.ts
iiPythonx: ~~The changes do seem to take effect live and still persist after reload.~~

src/i18n/locales/en.json
iiPythonx: Added them to the input.

src/renderer/store/settings.store.ts
iiPythonx: Nevermind, I just now realized what you meant but I'm also not sure how much I should bump the version by. Am I meant to go with 6.1 or 7? Also, building it locally and running it seemed to work even with my existing 
no-button-size Feishin config.

iiPythonx:
> but I feel like it might also make sense to have the button sizes take effect for right-controls

I would do this, but it seems like everything in right-controls is a lot more precisely sized. Not sure I feel comfortable trying to change all of those to respect the button size especially when for the playback speed element it's just raw text (so pretty sure I would have to deal with font sizing there as well).

Quite honestly, I think the right controls are somewhat fine as-is. If you guys want them to respect the button size then I'll leave that up to you I really don't trust myself with scaling of all them.

src/renderer/store/settings.store.ts
kgarner7: I do see it change, but for zustand persist, I believe the correct approach to make migrations happen is bump the version (to 7)

kgarner7:
> > but I feel like it might also make sense to have the button sizes take effect for right-controls
> 
> I would do this, but it seems like everything in right-controls is a lot more precisely sized. Not sure I feel comfortable trying to change all of those to respect the button size especially when for the playback speed element it's just raw text (so pretty sure I would have to deal with font sizing there as well).
> 
> Quite honestly, I think the right controls are somewhat fine as-is. If you guys want them to respect the button size then I'll leave that up to you I really don't trust myself with scaling of all them.

Honestly, I think that's fine as-is. Maybe last thing is change the title `Button size` to `Player bar button size` (so that it's clear from just the title), since it does **not** change all buttons. Thanks!

Issue #477, opened 2024-01-29
Global hotkeys don't work.


turkeyphant:
## Expected Behavior

Pressing global hotkeys such as Media_Prev, Media_Play_Pause, Media_Next performs the relevant actions

## Current Behavior

These keys don't have any effect in Feishin. They work perfectly in Winamp and Spotify. Global hotkeys is enabled.

## Steps to Reproduce

1. Turn on global hotkeys in settings

2. Press any of the media playback hotkeys

## Context

Can't use the keyboard to change tracks or play/pause.

## Your Environment

-   Application version (e.g. v0.1.0) : 0.5.3
-   Operating System and version (e.g. Windows 10) : Windows 10
-   Server and version (e.g. Navidrome v0.48.0) : [0.51.0](https://github.com/navidrome/navidrome/releases/tag/v0.51.0) (fd61b29a)
-   Node version (if developing locally) :

patrick-motard:
It works for me.

- app version: 0.5.3
- os: windows 11 & OSX Ventura 13.4 m1 chip
- server: navidrome [0.50.1-SNAPSHOT](https://github.com/navidrome/navidrome/compare/v0.50.1...9220d8f5) (9220d8f5)

I would add however that even though play/pause and next/previous work, the Feishin UI often reports the wrong state of playback. For example, if i have another app focused and I pause a song playing in Feishin, and then focus feishin, it still shows that the song is playing even though its not. Probably unrelated to what you're experiencing though.

The only time I've seen the playback buttons not work is if some other audio is taking precedence. Like if audio is playing in a tab in my browser and that was the last app i had focused that played audio. I can go to a different tab, or different app, and the media keys affect that browser tab still, not feishin.

patrick-motard:
#476 is the bug I was describing above.

iiPythonx:
Should be fixed as part of https://github.com/jeffvli/feishin/pull/488, didn't see this duplicate.

kgarner7:
Marking this as resolved. Until a new release, you can use https://feishin.vercel.app

Issue #404, opened 2023-12-07
Global media hotkeys don't work until toggled


Insprill:
## Expected Behavior

Global media hotkeys should be functional out of the box as the default state of the toggle indicates.

## Current Behavior

After a fresh install of the application, or the deletion of the config file, global media hotkeys are not functional until they're toggled off and back on.

## Steps to Reproduce

1. Remove the `global_media_hotkeys` from your config.
2. Start Feishin and start playing a song.
3. Try to use the media hotkeys. They won't be functional.
4. Open the settings and toggle global media hotkeys.
5. Try using the media hotkeys again. They will now be functional.

## Your Environment

-   Application version (e.g. v0.1.0): `v0.5.2`
-   Operating System and version (e.g. Windows 10): `Windows 11`
-   Server and version (e.g. Navidrome v0.48.0): `Navidrome 0.50.1`

auyylaso:
I'm having the same issue on windows 10
On V0.5.3

iiPythonx:
Should be fixed as part of #488.

kgarner7:
Closing this as resolved by #488 (thanks!)

PR #488, opened 2024-02-03
[bugfix] Fix global hotkeys not working on startup

### LINKED ISSUE #318: Global Media Keys not working
## Current Behavior

Global Media Keys not working until global media key option is toggled off and back on. 

## Steps to Reproduce

1.  Launch Feishin - try to use media keys - nothing happens
 
2.  Go into options and toggle global media key off and back on

Toggling the setting off and on seems to work but only for 1 launch.

## Context

Trying to set up  my media keys to work with Feishin.

## Your Environment

-   Application version (e.g. v0.1.0) : 0.4.1
-   Operating System and version (e.g. Windows 10) : Win 11
-   Server and version (e.g. Navidrome v0.48.0) : Navidrome  0.49.3
-   Node version (if developing locally) :

iiPythonx:
Fixes #318, #404, and #477 which was caused just by a broken comparison. Not sure why the value is `undefined` when it's set to `true`, but    at least now the statements are consistent and global hotkeys work on startup.

(tested and working, if it doesn't work then maybe theres a deeper issue, not sure)

### DIFF
--- a/src/main/main.ts
+++ b/src/main/main.ts
@@ -635,7 +635,7 @@ ipcMain.on(
 
         const globalMediaKeysEnabled = store.get('global_media_hotkeys') as boolean;
 
-        if (globalMediaKeysEnabled) {
+        if (globalMediaKeysEnabled !== false) {
             enableMediaKeys(mainWindow);
         }
     },
--- a/src/main/main.ts
+++ b/src/main/main.ts
@@ -317,9 +317,9 @@ const createWindow = async () => {
         }
     });
 
-    const globalMediaKeysEnabled = store.get('global_media_hotkeys') as boolean;
+    const globalMediaKeysEnabled = store.get('global_media_hotkeys', true) as boolean;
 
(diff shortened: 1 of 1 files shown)


src/main/main.ts
```diff
@@ -635,7 +635,7 @@ ipcMain.on(
 
         const globalMediaKeysEnabled = store.get('global_media_hotkeys') as boolean;
 
-        if (globalMediaKeysEnabled) {
+        if (globalMediaKeysEnabled !== false) {
```
kgarner7: Rather than changing the condition, a potentially more clear option would be `store.get('global_media_hotkeys', true) as boolean` (`true` being the default when not defined).

iiPythonx: I didn't know that was a thing, I'll definitely use that instead to clear up the code.

iiPythonx:
That should do it, testing shows that it still works and I'm not sure of any other improvements to make to it.

kgarner7:
Thanks!

Issue #490, opened 2024-02-04
German translations sentence-case


lymnyx:
Even though the translation file for german (`i18n/locales/de.json`) contains strings with properly capitalized words, those same strings are sentence-case when displayed in Feishin.

(The other languages' strings all seem to be lowercase anyway, that's why this is somewhat german-specific.)

lymnyx:
Fixed in 29df2a6.

PR #491, opened 2024-02-04
[bugfix/enhancement] Fixes german translations being sentence-case

lymnyx:
To me the sentenceCasePostProcessor doesn't seem to do much apart from cause #490 as it is applied on complete strings with capitalization, I am somewhat inexperienced with i18n though, so I might be mistaken here.

Additionally, I made minor changes to the `de.json` translation file.


src/i18n/i18n.ts
```diff
@@ -125,25 +125,9 @@ const titleCasePostProcessor: PostProcessorModule = {
...
-                    sentence.charAt(0).toLocaleUpperCase() + sentence.slice(1).toLocaleLowerCase()
-                );
-            })
-            .join('. ');
-    },
-};
 i18n.use(lowerCasePostProcessor)
     .use(upperCasePostProcessor)
     .use(titleCasePostProcessor)
-    .use(sentenceCasePostProcessor)
```
jeffvli: Removing `sentenceCasePostProcessor` affects every other translation so this change needs to be reverted.

Take a look at the [i18next post processor documentation](https://www.i18next.com/misc/creating-own-plugins#post-processor). The `translator` argument in the callback function includes the language key so you can modify each of the postprocessors specifically for the `de` language.

jeffvli (changes requested):

src/i18n/i18n.ts
lymnyx: Okay, will do, thank you.

lymnyx: (see changes in 1d8af79)

src/i18n/i18n.ts
```diff
@@ -128,13 +128,13 @@ const titleCasePostProcessor: PostProcessorModule = {
 const sentenceCasePostProcessor: PostProcessorModule = {
     type: 'postProcessor',
     name: 'sentenceCase',
-    process: (value: string) => {
+    process: (value: string, _: string, __: TOptions<StringMap>, translator: any) => {
```
jeffvli: I'm not a fan of the `_` and `__` variables. Could you rename them to `_key` and `_options` so it's more clear what they're intended to be?

src/i18n/i18n.ts
```diff
@@ -128,13 +128,13 @@ const titleCasePostProcessor: PostProcessorModule = {
...
         const sentences = value.split('. ');
 
         return sentences
             .map((sentence) => {
                 return (
-                    sentence.charAt(0).toLocaleUpperCase() + sentence.slice(1).toLocaleLowerCase()
+                    sentence.charAt(0).toLocaleUpperCase() + (translator.language !== 'de' ? sentence.slice(1).toLocaleLowerCase() : sentence.slice(1))
```
jeffvli: Since you don't want to perform any postprocessing on the string at all for `de`, you should move the conditional logic to the start of the function for the sake of readability and performance. Having the ternary operator inside the loop would affect performance (albeit very little) on `de` and other languages when it doesn't need to.

Like:
```typescript
    process: (value, _key, _options, translator) => {
        if (translator.language === 'de') {
            return sentence;
        }

        # rest of the function
    },
```

jeffvli (changes requested):

src/i18n/i18n.ts
lymnyx: I did initially write it like that, but when testing this caused the english-language strings (strings without existing `de`-translation) to be lowercase as sentenceCase didn't make their first character uppercase and `translator.language` was `de` no matter what.

 That is why I decided on this approach. Should I still change it?

jeffvli: No you should keep the logic as is then. I forgot about the fallback language case.

I do want to make the logic more re-usable though.

Something like
```
const ignoreSentenceCaseLanguages = ['de'];

# And then update the ternary condition
!ignoreSentenceCaseLanguages.includes(translator.language) ? ... : ....
```

jeffvli (approved):

PR #492, opened 2024-02-04
Image Resolution Setting

pyxfluff:
Title. Does not patch Electron's downscaling, but it will send the server the request for higher resolution images. Also changes the range from 300 to 175 if desired on low data devices. 350 is still the default.


kgarner7 (changes requested):
As it currently stands, this changes the cover art resolution for **all** images, including places where I'd argue that doesn't really make sense (namely, tiny cover thumbnails in track/queue list). The idea of changing the size for large views (fullscreen player, left queue?, card/poster view? not sure the best list...) is nice, but I don't think it is appropriate to change *every* image size.

pyxfluff:
Thanks, will look to change it some time later on this week. I did find a fix for the fullscreen player's bad resolution with a friend which I'll have him merge into my fork, will find other ones where the low quality is really bad and remove it where it's not.

iiPythonx:
Ignoring the very unclean push (git doesn't like me today), this PR now only adjusts the image resolution for the fullscreen player image. No other images are touched, nor are the resolution fallbacks removed.

src/i18n/locales/en.json
```diff
@@ -518,6 +518,8 @@
         "playButtonBehavior_optionAddLast": "$t(player.addLast)",
         "playButtonBehavior_optionAddNext": "$t(player.addNext)",
         "playButtonBehavior_optionPlay": "$t(player.play)",
+        "playerAlbumArtResolution": "Player Album Art Resolution",
+        "playerAlbumArtResolution_description": "The resolution for the large player's album art preview. Larger makes it look more crisp, but may slow loading down. Defaults to 0, meaning auto",
```
jeffvli: These need to be in lowercase similar to the other translation strings since we're applying postprocessing on them in the components.

pyxfluff:
Forgot about this, will finish the job later tonight.

src/i18n/locales/en.json
iiPythonx: Never realized that the rest of them were actually in all lowercase when I was making it. Also adjusted my old button size PR to be lowercase as well as kgarner's cache clearing strings (which I can revert if those changes are unwanted).

jeffvli (approved):

PR #501, opened 2024-02-11
[enhancement]: Audio player enhancements

### LINKED ISSUE #177: Default to the web-backend instead of mpv, or provide an easy way to disable mpv on first start.

Change the default behaviour of asking the user to select mpv executable to defaulting the Web backend without asking.

Current behaviour, the very first launch will ask for the path to `mpv.exe`:

image

It could be simply a button, `[ Use web-backend for music playback ]`, with some explanation why MPV is recommended.

## Additional context

To me, the web backend works perfectly fine as I had all my library pre-encoded in `m4a/aac-lc`. Though I am not sure about the reasoning behind the idea to default to mpv.

It would also be a good idea to search the `PATH` environment variable and populate it when found. I had it installed via [`scoop`](https://scoop.sh/#/apps?q=mpv&id=b05b47128464d8969416289383fbfc69a47353e3) and the binary directory was available in my PATH environment - `%USERPROFILE%\scoop\apps\mpv\current\mpv.exe`. It would be nice if I can just click "next" to continue.

### LINKED ISSUE #227: Automatically detect MPV executable path on Linux

Automatically detect MPV executable path on Linux by checking `PATH` environment variable.

## Additional context

N/A

kgarner7:
Supports running Feishin solely using web audio (useful for clients with problems with MPV). Also moves save/restore queue to utils, as MPV object is now optional.

Resolves #177


jeffvli:
I was actually testing this recently too.

I removed the `MPV required` blocker screen instead and added error notifications if mpv failed to initialize.

I also changed the binary path on the player initialization to fall back to `undefined` instead of `''` which lets node-mpv to access mpv based off the executable on the system env path.  The node-mpv documentation states that the binary path is required for Windows which is why I originally wanted to require all users to enter the path, but it works properly on my machine so it may resolve a lot of issues people were running into especially when trying to use mpv installed via package managers.

I'll push some of my commits over and refactor based on this approach since I'm not a fan of adding a setting explicitly to disable mpv when we already have the audio player setting.

PR #511, opened 2024-02-17
[enhancement] Allow hiding disc labels on albums with only one disc

iiPythonx:
This was a pretty quick PR based off of thatzkid's question in the discord server. As said in the title, it adds an option to the table setting component allowing you to turn off disc labels if the album you're viewing only has one disc in it.

In my opinion, this declutters the album view a bit. The feature is very useful for multi-disc albums, don't get me wrong; but I do prefer having them not be listed unless there's multiple.

As with the rest of my PRs, I'm not a TS dev. If I made some weird decision somewhere, let me know.

### DIFF
--- a/src/renderer/components/virtual-table/table-config-dropdown.tsx
+++ b/src/renderer/components/virtual-table/table-config-dropdown.tsx
@@ -367,6 +367,18 @@ export const TableConfigDropdown = ({ type }: TableConfigDropdownProps) => {
         });
     };
 
+    const handleUpdateDiscLabel = (e: ChangeEvent<HTMLInputElement>) => {
+        setSettings({
+            tables: {
+                ...useSettingsStore.getState().tables,
+                [type]: {
+                    ...useSettingsStore.getState().tables[type],
+                    hide1DiscLabel: e.currentTarget.checked,
+                },
+            },
+        });
+    };
+
     return (
         <>
             <Option>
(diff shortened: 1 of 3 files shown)


kgarner7 (changes requested):
I'm not too keen to take this in as-is, as the disc label checkbox allows quick select/deselect of all songs and this PR would remove that capability. Not sure of the best approach, for preserving this option though

iiPythonx:
Not sure of the best way to proceed, although I can make it default to `false` which won't remove the disc labels unless people explicitly wish to enable it.

iiPythonx:
Couple of ideas:
- If this PR is enabled in the table settings, the "select all" box could be moved into the table header instead to keep that functionality
- Alternatively, is there any way to make CTRL+A select all the songs in an album? Right now it just selects everything in the app, but if it selected the tracks in the album that could be another option.
- (or) if this PR is more an annoyance then an enhancement, I can just close this. I don't exactly have a use for this PR as-is... I'm just doing it since both my friend on discord as well as thatzkid expressed the want for this feature (or something similar).

kgarner7:
> Couple of ideas:
> 
>     * If this PR is enabled in the table settings, the "select all" box could be moved into the table header instead to keep that functionality
> 
>     * Alternatively, is there any way to make CTRL+A select all the songs in an album? Right now it just selects everything in the app, but if it selected the tracks in the album that could be another option.
> 
>     * (or) if this PR is more an annoyance then an enhancement, I can just close this. I don't exactly have a use for this PR as-is... I'm just doing it since both my friend on discord as well as thatzkid expressed the want for this feature (or something similar).

1 if possible I think would be the best approach (that's how Navidrome handles select in general). I'm not sure about select all keyboard shortcut, because I'm not sure how easy it is to tell which context you are in (in particular, if you have both the queue and album page open).

iiPythonx:
Well, I guess I bit off a bit more then I can chew with those ideas. 100% agree with your statement on the keyboard shortcuts.

My main issue right now is I would have to try and add a select all box into the header which seems like it would be impossible without creating another table column for it. Granted, that shouldn't be too hard, but I'm pretty sure it would shift every other element in the table which is NOT something I consider acceptable. Additionally, even if I do get the select all box rendering inside the table header, I have no idea how I would reimplement the select all functionality since it would no longer be inside of the album detail component.

In short, my non-react brain definitely can't get this implemented.  
Changes like this are wayyyyy outside my comfort zone with React/TS, so I'm going to close this PR.

PR #515, opened 2024-02-18
[enhancement] DRPC Image Support 

thororen1234:
I changed the discord rpc to have navidrome and jellyfin using the currentSong?.imageUrl and I changed the interval from 15 to 10.

### DIFF
--- a/src/renderer/features/discord-rpc/use-discord-rpc.ts
+++ b/src/renderer/features/discord-rpc/use-discord-rpc.ts
@@ -36,7 +36,7 @@ export const useDiscordRpc = () => {
         const activity: SetActivity = {
             details: currentSong?.name.padEnd(2, ' ') || 'Idle',
             instance: false,
-            largeImageKey: undefined,
+            largeImageKey: currentSong?.imageUrl || 'icon',
             largeImageText: currentSong?.album || 'Unknown album',
             smallImageKey: undefined,
             smallImageText: currentStatus,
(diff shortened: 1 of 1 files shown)


iiPythonx:
Assuming I'm reading it right, this no longer respects `discordSettings.showServerImage`, which it seems like it should; some people may not wish to have images on their status.

Additionally, this will still only work if the server URL Feishin is using is publicly accessible, so out of curiosity, what's the point of always sending it to discord? For anybody using a local media server (me included) no image will show up at all, not even the fallback icon.

jeffvli:
Navidrome (and Subsonic) requires your credentials on the image URL so exposing that to Discord will potentially leak your username and password.

thororen1234:
> Assuming I'm reading it right, this no longer respects `discordSettings.showServerImage`, which it seems like it should; some people may not wish to have images on their status.
> 
> Additionally, this will still only work if the server URL Feishin is using is publicly accessible, so out of curiosity, what's the point of always sending it to discord? For anybody using a local media server (me included) no image will show up at all, not even the fallback icon.

That was my bad I accidentally deleted it because mobile select sucks.

It should work on local servers because it would use the url of the image through your pc discord just uses the url it's given.

thororen1234:
> Navidrome (and Subsonic) requires your credentials on the image URL so exposing that to Discord will potentially leak your username and password.

This is a very valid point I think some sort of tinkering is in place.

iiPythonx:
> discord just uses the url it's given.

Last time I checked, all images that go through discord (rpc included) have to go through discord's proxy layer first. Because of that, sending a local URL to discord will just result in the image not loading, regardless of who tries to view it.

Pretty sure they do it this way to prevent things such as getting ip logged just from receiving an image.
Regardless, Feishin would need to upload the images to a service such as imgur in order to get them to display.

thororen1234:
> > discord just uses the url it's given.
> 
> Last time I checked, all images that go through discord (rpc included) have to go through discord's proxy layer first. Because of that, sending a local URL to discord will just result in the image not loading, regardless of who tries to view it.
> 
> Pretty sure they do it this way to prevent things such as getting ip logged just from receiving an image. Regardless, Feishin would need to upload the images to a service such as imgur in order to get them to display.

That's dumb if so because the RPC server itself is locally hosted so says discord.

thororen1234:
Anyways I'll come back when this is completely secure and such.

kgarner7:
Consider following (or even contributing) on https://github.com/opensubsonic/open-subsonic-api/discussions/25 for better auth

PR #558, opened 2024-03-23
Implement Navidrome sharing

### LINKED ISSUE #80: Share from context menu
Navidrome has recently implemented sharing! It would be convenient to interact with this feature through Feishin, rather than navigating to my Navidrome instance.

## Additional context
https://github.com/navidrome/navidrome/pull/2106

iiPythonx:
PR that adds an option for sharing an album or songs via Navidrome (nothing else yet), resolves #80.  
There are more then likely *a lot* of issues with how I did this, if you have any feedback on it please lmk (considering I've still barely touched React in my life).

Two nitpicks:
- Right now, this PR selects `resourceType` based on the first item you've selected. This *should* be fine since I'm assuming you can't magically select two items of different types within Feishin (ie. selecting an album <ins>and</ins> a song at the same time).
- Additionally, there seems to be a lack of margin on TextInput fields (not that bad but it's a bit squished):

image

Anyways, as stated in the beginning, any feedback would be very much appreciated.

### DIFF
--- a/src/renderer/api/controller.ts
+++ b/src/renderer/api/controller.ts
@@ -8,6 +8,7 @@ import type {
     AlbumArtistDetailArgs,
     AlbumArtistListArgs,
     SetRatingArgs,
+    ShareItemArgs,
     GenreListArgs,
     CreatePlaylistArgs,
     DeletePlaylistArgs,
@@ -55,6 +56,7 @@ import type {
     SimilarSongsArgs,
     Song,
     ServerType,
+    ShareItemResponse,
 } from '/@/renderer/api/types';
 import { DeletePlaylistResponse, RandomSongListArgs } from './types';
 import { ndController } from '/@/renderer/api/navidrome/navidrome-controller';
@@ -102,6 +104,7 @@ export type ControllerEndpoint = Partial<{
     scrobble: (args: ScrobbleArgs) => Promise<ScrobbleResponse>;
     search: (args: SearchArgs) => Promise<SearchResponse>;
     setRating: (args: SetRatingArgs) => Promise<RatingResponse>;
+    shareItem: (args: ShareItemArgs) => Promise<ShareItemResponse>;
     updatePlaylist: (args: UpdatePlaylistArgs) => Promise<UpdatePlaylistResponse>;
 }>;
 
@@ -149,6 +152,7 @@ const endpoints: ApiController = {
         scrobble: jfController.scrobble,
         search: jfController.search,
         setRating: undefined,
+        shareItem: undefined,
         updatePlaylist: jfController.updatePlaylist,
     },
     navidrome: {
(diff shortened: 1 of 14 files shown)


src/renderer/api/controller.ts
```diff
@@ -223,6 +228,7 @@ const endpoints: ApiController = {
         scrobble: ssController.scrobble,
         search: ssController.search3,
         setRating: undefined,
+        shareItem: undefined,
```
kgarner7, line 231: Strictly speaking (even though it wouldn't be used for now) Subsonic API also supports sharing https://subsonic.org/pages/api.jsp#createShare

src/renderer/api/navidrome/navidrome-types.ts
```diff
@@ -343,6 +343,17 @@ const removeFromPlaylistParameters = z.object({
     id: z.array(z.string()),
 });
 
+const shareItem = z.object({
+    id: z.string(),
+});
+
+const shareItemParameters = z.object({
+    description: z.string(),
+    downloadable: z.boolean(),
```
kgarner7, line 352: Navidrome (and subsonic) also supports `expires` value.

src/renderer/api/types.ts
```diff
@@ -766,6 +766,18 @@ export type RatingQuery = {
 
 export type SetRatingArgs = { query: RatingQuery; serverId?: string } & BaseEndpointArgs;
 
+// Sharing
+export type ShareItemResponse = { id: string } | undefined;
+
+export type ShareItemBody = {
+    description: string;
+    downloadable: boolean;
+    resourceIds: string;
```
kgarner7, line 776: Similar, should probably also have expiry here as well (optional, either date or unix timestamp)

kgarner7 (changes requested):
Sharing is not available in every Navidrome. Make sure to add it as a feature in `feature-types.ts` (ServerFeature), modify `navidrome-controllers.ts#getServerInfo` to check for the correct version, and check if that feature exists.

src/renderer/api/navidrome/navidrome-types.ts
iiPythonx: I'll get to work on implementing this right now along with the server version check.

kgarner7 (changes requested):
I added a few changes. It might be nice to have the ability to view existing shares, but this is a lower priority (I _may_ take a look at implementing this).

I'm not particularly a fan of having the share only copied to clipboard, as that might not work especially in browser environments where clipboard access is limited. It would be good to have a way to open in browser (maybe when clicking on the notification?)

iiPythonx:
> It would be good to have a way to open in browser (maybe when clicking on the notification?)

Went ahead and made it so clicking on the success notification does indeed open the share in browser.  

> It might be nice to have the ability to view existing shares, but this is a lower priority (I may take a look at implementing this).

I'll leave that up to you then (if you decide to work on it).

kgarner7 (approved):
...ngl, adding share listing is more of a pain than I'd like. It can come as a feature later.

iiPythonx:
Replacement is at https://github.com/jeffvli/feishin/pull/575 (sorry!).

Issue #556, opened 2024-03-22
[Feature Request] Add current Song to Playlist Hotkey


gittremblehub:
I think it would be really cool if you had a fast way to add the current playing Song to a  playlist.

My suggestion would be:
Add like 5 new hotkey options (or make it variable) which are not set by default(just like how other hotkeys are right now).
And then making it possible to choose a playlist for that hotkey where the song should be added.

I am not that familiar with the code-base and I hope it is a feasable suggestion.

PR #560, opened 2024-03-27
[feat]  Create "Add current track" button in playlist view

will-hut:
I'm enjoying using Feishin as my music player for my library. I thought it would be nice to add a "Add current track" button to the playlist sidebar. That way, when a song is playing, I can quickly add it to as many playlists as I feel like it belongs in.

demo

Currently, I have it hiding the button if the playlist has rules (ie, a smart playlist). You can add a song to a shared playlist, but if you don't have (admin) permission, it throws an error toast. 

This also allows adding duplicates of a song to a playlist, so a toggle somewhere to disable duplicates could be a nice feature.

I'm not a React dev and haven't really contributed to open source, so if there's any issues please let me know! I think I did a decent job, but I'm always looking for ways to make my code better.

### DIFF
--- a/src/renderer/features/sidebar/components/sidebar-playlist-list.tsx
+++ b/src/renderer/features/sidebar/components/sidebar-playlist-list.tsx
@@ -202,16 +202,17 @@ export const SidebarPlaylistList = ({ data }: SidebarPlaylistListProps) => {
                             title: t('error.genericError', { postProcess: 'sentenceCase' }),
                         });
                     },
+                    onSuccess: () => {
+                        toast.success({
+                            message: t('form.addToPlaylist.success', {
+                                message: 1,
+                                numOfPlaylists: 1,
+                                postProcess: 'sentenceCase',
+                            }),
+                        });
+                    },
                 },
             );
-
-            toast.success({
-                message: t('form.addToPlaylist.success', {
-                    message: 1,
-                    numOfPlaylists: 1,
-                    postProcess: 'sentenceCase',
-                }),
-            });
         },
         [addToPlaylistMutation, currentSong, serverID, t],
     );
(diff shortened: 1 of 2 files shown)


will-hut:
This adds a feature similar to what is mentioned in #556

src/renderer/features/sidebar/components/sidebar-playlist-list.tsx
```diff
@@ -161,8 +185,43 @@ export const SidebarPlaylistList = ({ data }: SidebarPlaylistListProps) => {
...
+                    serverId: serverID,
+                },
+                {
+                    onError: (err) => {
+                        toast.error({
+                            message: `[${playlist?.name}] ${err.message}`,
+                            title: t('error.genericError', { postProcess: 'sentenceCase' }),
+                        });
+                    },
+                },
```
kgarner7, line 214: You should move the `toast.success` to `onSuccess` here

src/renderer/features/sidebar/components/sidebar-playlist-list.tsx
```diff
@@ -161,8 +185,43 @@ export const SidebarPlaylistList = ({ data }: SidebarPlaylistListProps) => {
...
+            });
+        },
+        [addToPlaylistMutation, currentSong, serverID, t],
+    );
+
     const memoizedItemData = useMemo(() => {
-        const base = { defaultFullPlaylist, handlePlay: handlePlayPlaylist };
+        const base = {
+            defaultFullPlaylist,
+            handleAdd: handleAddCurrent,
```
kgarner7, line 223: This should not be enabled for Navidrome shared playlists (only the owner can change them)

kgarner7 (changes requested):
Thanks for the commit! A few other thoughts:
- This doesn't consider duplicates (e.g. adding the same song multiple times)
- The new icon isn't quite intuitive, maybe it'd be worthwhile to remove the delay for tooltip (or another icon)

I'm also not sure if this is necessarily appropriate period, but I'll defer to @jeffvli for that.

PR #567, opened 2024-03-31
Mac control center support (Fixes #549)

### LINKED ISSUE #549: MacOS Control Center for Web player
## Expected Behavior

As described in https://github.com/jeffvli/feishin/issues/204, the currently playing song should appear in the control center on the mac so the system understand media is playing from Feishin. Without this, controlling the media from a bluetooth device like ear pods does not work, even though keyboard media keys work. 

## Current Behavior

Control the media, like playing or pausing from bluetooth headset does not control Feishin.

## Steps to Reproduce

1. Play media in Feishin

2. Observe the media playing in Feishin does not appear in the ▶️ status menu bar item or in the Control Center.

3. Playing or pausing from the bluetooth audio device does not control media in Feishin. 

## Your Environment

-   Application version: 0.6.1
-   Operating System and version: macOS 14
-   Server: Navidrome

Explosion-Scratch:
Currently only working in the web version


kgarner7 (changes requested):
Thanks for the PR, but I'm not really sure what this is trying to do:
1. `navigator.mediaSession` **already** is in use, but in `use-center-controls.ts`. While there is some room for improvement there, it does at least work in web browsers on MacOS (Safari, Firefox, Chrome)
2. The change as proposed is also broken: 
    1. `Cannot read properties of undefined (reading 'name')` when the queue is empty (most likely artist metadata)
    2. More severely, I see `Invalid hook call. Hooks can only be called inside the body of a function component` about every second.
    3. At least on Feishin Desktop on macOS, this doesn't seem to help either (I don't personally understand why, even with mediaSession set properly MacOS doesn't show)
    4. This removes the ability for mediaSession  to play/pause/etc.
    
I appreciate the time you spent, I would also like this to be fixed on desktop (and less janky elsewhere as well) [I did try and look why mediaSession isn't updating control center, but I haven't found any good leads thus far). But I'd much rather fix the existing implementation (or find out something macOS desktop-only that we're missing in the first place)

Explosion-Scratch:
@kgarner7 I've fixed 2.i and 2.ii just now (will push in a few), also mediaSession play pause seems to work fine for me. I'll push in just a sec

Explosion-Scratch:
🤦🏻 Didn't even see use-center-controls.ts lol

Could it maybe not be working due to the `isElectron()` call?

Explosion-Scratch:
The issue with electron vs web might be due to BrowserWindow instead of BrowserView:
https://github.com/electron/electron/issues/31448#issuecomment-1002980034

Explosion-Scratch:
Using use-center-controls isn't setting `navigator.mediaSession` at all =/

(I commented out my code when testing this)

kgarner7:
It does look like there is an issue where the **first** track doesn't get metadata, but it definitely sets it otherwise. And I having the actionHandler without `isElectron()` doesn't seem to help either

Explosion-Scratch:
Yeah I've been looking through it - It only seems to update when you click pause and play buttons not on skipping songs. I don't have time to work on it more right now but I can later. Some improvements I want to make though:
- Update mediaSession on skip back and forward / debug
- Figure out why electron isn't showing it (it shows MPV when MPV player is selected but no art or anything) - I'm assuming you're using a modernish electron
- Use Blobs for the images - This fixed a lot of errors for people on stackoverflow 🤷
- Make sure the handlers don't conflict with media keys - We already sense media keys through global shortcuts but these also are applied to the mac control center so double pause/seek would be bad

Explosion-Scratch:
@kgarner7 I found it!! I was looking through MPV to see how they did it (take it from the pros):
https://github.com/mpv-player/mpv/blob/7f3ca6c524d01d92f756909ccdf877bccd627808/osdep/mac/remote_command_center.swift#L148-L158

We probably need to use this on mac:
https://developer.apple.com/documentation/mediaplayer/mpnowplayinginfocenter

Could use this:
https://github.com/MarshallOfSound/electron-media-service

PR #582, opened 2024-04-11
Fix portrait mode detection

Kaydax:
For some reason it was hard coded to only switch to portrait mode when the window width is less-than or equal to 768px.

This was a very simple fix and seems to work perfectly fine after testing it for a bit with multiple window sizes.

### DIFF
--- a/src/renderer/features/player/components/center-controls.tsx
+++ b/src/renderer/features/player/components/center-controls.tsx
@@ -63,7 +63,7 @@ const SliderValueWrapper = styled.div<{ $position: 'left' | 'right' }>`
     justify-content: center;
     max-width: 50px;
 
-    @media (width <= 768px) {
+    @media (orientation: portrait) {
         display: none;
     }
 `;
@@ -81,7 +81,7 @@ const ControlsContainer = styled.div`
     justify-content: center;
     height: 35px;
 
-    @media (width <= 768px) {
+    @media (orientation: portrait) {
         ${ButtonsContainer} {
             gap: 0;
         }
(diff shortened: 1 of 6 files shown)


src/renderer/features/player/components/center-controls.tsx
```diff
@@ -63,7 +63,7 @@ const SliderValueWrapper = styled.div<{ $position: 'left' | 'right' }>`
     justify-content: center;
     max-width: 50px;
 
-    @media (width <= 768px) {
```
kgarner7: This should be based on width alone (not orientation)

src/renderer/features/player/components/center-controls.tsx
```diff
@@ -81,7 +81,7 @@ const ControlsContainer = styled.div`
     justify-content: center;
     height: 35px;
 
-    @media (width <= 768px) {
+    @media (orientation: portrait) {
```
kgarner7: Same here

src/renderer/features/player/components/playerbar.tsx
```diff
@@ -30,7 +30,7 @@ const PlayerbarControlsGrid = styled.div`
     gap: 1rem;
     height: 100%;
 
-    @media (width <= 768px) {
+    @media (orientation: portrait) {
```
kgarner7: Same here

src/renderer/features/shared/components/library-header.module.scss
```diff
@@ -58,7 +58,7 @@
         }
     }
 
-    @container (min-width: 768px) {
+    @container (orientation: portrait) {
```
kgarner7: And here

kgarner7 (changes requested):
Thanks for the PR! While detecting orientation for the full screen player and related components probably makes sense, I'd disagree that you should _also_ apply the same changes to the playbar, which is **just** width-based

PR #598, opened 2024-04-25
Allow smaller album covers in card/poster display

Dylancyclone:
Firstly thank you for your awesome work on this project!

Something I noticed on my screen was that even at the smallest setting the Album Covers on the Albums page felt a little too big. It would be awesome to be able to make them a bit smaller, especially with how small they're allowed to get in the table/list mode

Old minimum:
image

New minimum:
image

I couldn't find any contribution guidelines in this repo so I hope this is ok :D


jeffvli:
Thanks! I went ahead and changed the other grids as well.

PR #603, opened 2024-05-02
Fix LrcLib fetcher

vukanovics:
[LrcLib expects durations in seconds](https://lrclib.net/docs), while the duration field in Feishin is [normalized to milliseconds](https://github.com/jeffvli/feishin/blob/9cd8807a75e5acb158ac711cd053e13bd33523c6/src/renderer/api/navidrome/navidrome-normalize.ts#L97).
This causes the lyrics fetch to fail.

Convert the normalized duration into seconds.


kgarner7:
Thanks!

Issue #565, opened 2024-03-30
[Feature request] View and play similar songs


Explosion-Scratch:
I know you can do this in full screen but there should be a context menu item for it and a route to a song's "radio".

I hate feature requesting without contributing to a project so I'm trying to implement this myself but I can't figure out how to navigate to a page from `context-menu-provider.tsx` or where to put the new route (`app-router.tsx`? - How do I pass data to routes, like song ID?)

I'll figure it out if I get some time

kgarner7:
If you actually want to take a look at this (would be great!), you could look at the following sections:
1. Add a new event for `ContextMenuItemType` in `events.ts`
2. Add that event where appropriate in `context-menu-items.tsx` (probably just  queue/song/playlist)
3. Implement that feature in `context-menu-provider.tsx`. There is already a controller for this, so you could call that. Of note, while Jellyfin instant mix should _always_ work, Navidrome (really, Subsonic) similar tracks may not work always

Explosion-Scratch:
@kgarner7 I would love to! Thanks for your advice on which sections, I'm very new to react 😅 (more of a Svelte/Vue guy). I have a busy week this week and next due to some travel but likely after that things will settle down.

Navidrome/subsonic similar songs seem to fail for me so I'll need to read some docs first. My ultimate goal would be something like spotify's song radios, so:
Corner click -> List of similar songs is displayed (like an album layout) -> Play from there

kgarner7:
Yeah, `gettSimilarSongs` relies on last.fm so it isn't as reliable. I've added fallbacks to both Navidrome (just get random songs from the artist) and Jellyfin (InstantMix; not as good, but at least you'll get a playlist). There are still cases where there will be no tracks at all (e.g. album artist with only one track, no similar), but I _think_ that's about as good as it'll get.

BobWs:
`gettSimilarSongs` works great on the ios clients **play:sub** and **substreamer**. This feature would be nice to see in Feishin and then add the ability to save the result into a playlist, this would make it a great playlist management for Navidrome!

kgarner7:
Given #650 and the ability to see similar songs in full screen player, I'll mark this as resolved

PR #650, opened 2024-06-24
Add option to play similar tracks from the context menu

isaiahfuller:
I noticed that you can get similar songs in the fullscreen view and I also noticed some conversation about it at #565, so I tried to follow what was said there.


jeffvli:
Looks good, thanks for the PR!

PR #659, opened 2024-07-03
[bugfix]: use tsx instead of ts-node, update @electron/rebuild

kgarner7:
Replace ts-node with tsx for transpile, and upgrade `@electron/rebuild`. It _might_ be possible to fix ts-node to work, but I did not have any luck with that. This requires node >= 18 now

### DIFF
--- a/package.json
+++ b/package.json
@@ -277,6 +277,7 @@
         "terser-webpack-plugin": "^5.3.1",
         "ts-jest": "^27.1.4",
         "ts-loader": "^9.2.8",
+        "ts-node": "^10.9.2",
         "tsconfig-paths-webpack-plugin": "^4.0.0",
         "typescript": "^5.2.2",
         "typescript-plugin-styled-components": "^3.0.0",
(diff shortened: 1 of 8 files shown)


package.json
```diff
@@ -295,6 +293,7 @@
         "@ag-grid-community/infinite-row-model": "^28.2.1",
         "@ag-grid-community/react": "^28.2.1",
         "@ag-grid-community/styles": "^28.2.1",
+        "@electron/rebuild": "^3.6.0",
```
sel10ut: Shouldn't it be in `devDependencies`?

package.json
```diff
@@ -350,14 +349,15 @@
         "semver": "^7.5.4",
         "styled-components": "^6.0.8",
         "swiper": "^9.3.1",
+        "tsx": "^4.16.2",
```
sel10ut: and this

jeffvli:
I might just hold off on the v31 upgrade for now. I can't quite get the ESM build to work on [my branch](https://github.com/jeffvli/feishin/tree/fix/esm) either. I got most of it working except for preload which is still breaking.

sel10ut:
it works for me on development with tsx

sel10ut:
try keeping `electron -r ts-node/register/transpile-only` with tsx

jeffvli:
Yep.. working fine for me now. I think when I initially tested, it broke because ts-node wasn't installed. I tried on my own branch and started messing with some ESM stuff but this is much easier with just tsx.

PR #660, opened 2024-07-03
(Jellyfin) Allow multiple sessions from the same user with different instances

### LINKED ISSUE #644: Jellyfin - Can only have one feishin instance logged in at a time
### Expected Behavior

I should be able to have multiple feishin installations on various computers/locations logged in at once.

### Current Behavior

As soon as a second feishin instance logs in, the first will log out.  Occurs irrespective of network (internal vs over internet) or OS (windows vs linux).  I've tried with Jellyfin's "Maximum number of simultaneous user sessions" set to 0 (disabled) and 20 with no change in behavior.  Using Jellyfin 10.9.5.

### Steps to Reproduce

From a primary computer, log into Jellyfin with Feishin
From a secondary computer, log into Jellyfin with Feishin.  Note the primary computer's Feishin has been logged out.

### Possible Solution

I don't know enough about how the auth works, but I did find it interesting that if you reproduce this with a web hosted Feishin, the same thing applies but only to the hosted version - i.e. if you sign into the web version, it will sign out other web versions but not the desktop apps, and vice versa.  

### Context

_No response_

### Application version

0.7.1

### Operating System and version

Arch Linux, Windows 10, Firefox web app

### Server and Version

Jellyfin 10.9.5

### Node Version (if developing locally)

_No response_

### LINKED ISSUE #653: "Your session has expired" with Jellyfin 10.9.6
### Expected Behavior

Opening the app it should login normally

### Current Behavior

Every now and then Feishin says "Your session has expired, please relogin"

I need to fill in my password

### Steps to Reproduce

.

### Possible Solution

It would be cool, if Feishin could use the stored credentials and tries to relogin itself, if this fails, then it should throw the error. 

### Context

_No response_

### Application version

0.7.1

### Operating System and version

Fedora, Windows 11 happens on both

### Server and Version

Jellyfin 10.9.6

### Node Version (if developing locally)

_No response_

sel10ut:
#### Changes
 * Allow multiple sessions from a single user with different instances of the same client. Although [it's said](https://gist.github.com/nielsvanvelzen/ea047d9028f676185832e51ffaf12a6f) that you cannot have multiple users signed in to the client with the same `deviceId`, this limitation has already been removed [here](https://github.com/jellyfin/jellyfin/pull/10197).

* Migrate from deprecated api auth methods. Additionally, send more information, as the official web client [does](https://github.com/jellyfin/jellyfin-sdk-typescript/blob/e44fa0cf548b586ab489f3e22bf3323c26b94115/src/api.ts), which will, for example, allow Feishin to signal its version immediately after an update, not only after a re-login.

#### Issues
Fixes #644, maybe fixes #653


jeffvli:
Looks good, thanks!

PR #707, opened 2024-08-14
[slightly less scuffed bugfix]: Update table rating/favorite when updated anywhere …

kgarner7:
…else

Modify player store to have temporary state for favorite/rating update Add effect handler for `virtual-table` to update rating/favorite for players

Note that this does not handle song grid view.
Using a similar handler for gird view did not work, as it appeared to result in inconsistent state.

Finally, this is probably not the optimal solution. Performance appears fine for ~20k items, but no guarantees.


src/renderer/features/playlists/components/playlist-detail-song-list-content.tsx
```diff
@@ -248,7 +249,7 @@ export const PlaylistDetailSongListContent = ({ tableRef }: PlaylistDetailConten
                         onCellContextMenu: handleContextMenu,
                         status,
                     }}
-                    getRowId={(data) => data.data.uniqueId}
+                    getRowId={(data) => data.data.id}
```
jeffvli: Duplicate songs can exist in playlists which is why uniqueId is used here.

kgarner7:
Performance:
1. Feishin chugs when opening an album with ~270 tracks. But that's _probably_ more a problem of not paginating albums. Favoriting/starring items (max ~20, before Navidrome errors out) is pretty fast after queue favorite. Playcount/playdate update is also quick.
2. The worst case, Navidrome playlist with 20K tracks (Jellyfin doesn't allow for duplicates, apparently) is surprisingly fast as well.

src/renderer/features/playlists/components/playlist-detail-content.tsx
```diff
@@ -156,6 +157,18 @@ export const PlaylistDetailContent = ({ tableRef }: PlaylistDetailContentProps)
...
 
+    // Duplicates are only present if on Navidrome
+    const getId = useCallback(
+        (data: GetRowIdParams<any>): string => {
+            return server?.type === ServerType.JELLYFIN
+                ? data.data.id
+                : `${data.data.id}-${data.data.pageIndex}`;
+        },
+        [server?.type],
+    );
```
jeffvli: Does it affect performance to just use `${data.data.id}-${data.data.pageIndex}` as the global id instead of the conditional? If not, might be cleaner to just have one implementation rather than two.

kgarner7: Looks like `getRowNode` is generally just a linear search underneath, so using I was hoping it'd be a more efficient implementation, but no. So I'll just reimplement using `forEachNode`, which is equally (or more, for multiple items) efficient currently.

src/renderer/features/playlists/components/playlist-detail-content.tsx
```diff
@@ -221,14 +234,12 @@ export const PlaylistDetailContent = ({ tableRef }: PlaylistDetailContentProps)
...
-                    getRowId={(data) => {
-                        // It's possible that there are duplicate song ids in a playlist
-                        return `${data.data.id}-${data.data.pageIndex}`;
-                    }}
+                    getRowId={getId}
                     rowClassRules={rowClassRules}
                     rowData={playlistSongData}
                     rowHeight={60}
                     rowSelection="multiple"
+                    shouldUpdateSong={server?.type === ServerType.JELLYFIN}
```
jeffvli: Does this not work for Navidrome?

kgarner7: Outdated, but this was because the Jellyfin implementation can rely on the fact there were no duplicates, whereas Navidrome cannot

jeffvli: Oh my bad, I was still reviewing without refreshing your changes :)

PR #717, opened 2024-08-29
Album blur, allow clicking the playerbar to toggle the player, misc changes

pyxfluff:
Has a couple of changes.

- Adds a blur to the album page like so:
image
image
This change respects settings in the player ("Enable background image") and will disable itself if the option is not enabled. Migrate to a global setting later possibly(?)

- Allows clicking the bottom player bar to minimize the player:
https://github.com/user-attachments/assets/37ccc199-253f-47b8-b4e7-beede21775cd

- Updates the translations to use proper plural words only if anything is more than one (1 song instead of "1 songs") and updates Spanish a little bit


pyxfluff:
Forgot to add translations to the albums header. Not adding due to #720 but I'll merge [this commit in](https://github.com/iiPythonx/feishin/commit/e6180ea7a37b49181d93e400d20591548b4405c4) to here when both of my PRs are reviewed

jeffvli:
The styling is broken when I'm running it from your branch (doesn't match your screenshots). Can you check directly from your PR branch?

Live:
image
PR:
image


image

Felitendo:
Please merge this. It looks soo clean

package-lock.json
kgarner7, line 1: I don't really see a reason for this PR to have changes to package-lock.json. Please reverte this

release/app/package-lock.json
kgarner7, line 1: Same here

src/renderer/features/player/components/playerbar.tsx
```diff
@@ -70,14 +72,21 @@ export const Playerbar = () => {
...
+    };
 
     const autoNextFn = useCallback(() => {
         const playerData = autoNext();
         updateSong(playerData.current.song);
     }, [autoNext]);
 
     return (
-        <PlayerbarContainer>
+        <PlayerbarContainer onClick={handleToggleFullScreenPlayer}>
```
kgarner7: Currently, this has weird behavior when you click on any of the metadata items in the bottom left. Specifically, it will try to open full screen player before doing some other operation. This includes the title (open to queue), artist (open to album page), album, and context menu.
Please either fix this behavior, or fix this behavior.
Also, personally, but this behavior doesn't make sense regardless, as there is no indicator that clicking anywhere else has any action.

kgarner7 (changes requested):

src/renderer/store/settings.store.ts
```diff
@@ -333,6 +335,8 @@ const initialState: SettingsState = {
     general: {
         accent: 'rgb(53, 116, 252)',
         albumArtRes: undefined,
+        albumBackground: false,
+        albumBackgroundBlur: 1.5,
```
kgarner7: One more, since I forgot to add this as part of review. I think it would make sense to default to 6 to preserve the similar behavior with just a single color. I do appreciate that this is configurable though.

src/renderer/features/artists/components/album-artist-detail-header.tsx
```diff
@@ -26,12 +28,16 @@ export const AlbumArtistDetailHeader = forwardRef(
             {
                 id: 'albumCount',
                 secondary: false,
-                value: detailQuery?.data?.albumCount && `${detailQuery?.data?.albumCount} albums`,
+                value: `${detailQuery?.data?.albumCount} ${t('entity.album', {
```
kgarner7: Please do `t("entity.albumWithCount", { count: detailQuery?.data?.albumCount })` instead.

src/renderer/features/artists/components/album-artist-detail-header.tsx
```diff
@@ -26,12 +28,16 @@ export const AlbumArtistDetailHeader = forwardRef(
...
+                    count: detailQuery?.data?.albumCount as number,
+                })}`,
             },
             {
                 id: 'songCount',
                 secondary: false,
-                value: detailQuery?.data?.songCount && `${detailQuery?.data?.songCount} songs`,
+                value: `${detailQuery?.data?.songCount} ${t('entity.track', {
+                    count: detailQuery?.data?.songCount as number,
```
kgarner7: Similarly, please do `t("entity.trackWithCount", { count: detailQuery?.data?.songCount })` instead.

pyxfluff:
Git just totally messed everything up, will let you know when its safe to merge bc everything is screwed up now

src/renderer/layouts/default-layout/player-bar.tsx
```diff
@@ -1,15 +1,38 @@
...
+    const { playerbarOpenDrawer } = useGeneralSettings();
+    const [barState, setBarState] = useState('playerbar-bg');
+
+    const handleMouseEnter = () => {
+        setBarState('playerbar-bg-active');
+    };
+
+    const handleMouseLeave = () => {
+        setBarState('playerbar-bg');
+    };
```
jeffvli: You should be using a pseudo hover style here rather than capturing mouseEnter and mouseLeave events to set the state.

For an example, check the [styled-components](https://styled-components.com/docs/api#supported-css) docs for how to implement hover styles.

src/renderer/layouts/default-layout/player-bar.tsx
```diff
@@ -1,15 +1,38 @@
+import { useState } from 'react';
 import styled from 'styled-components';
 import { Playerbar } from '/@/renderer/features/player';
+import { useGeneralSettings } from '/@/renderer/store/settings.store';
 
-const PlayerbarContainer = styled.footer`
+interface PlayerbarContainerProps {
+    barState: string;
```
jeffvli: To accomodate the hover style, you can send a boolean value rather than the css variable string, and then conditionally choose either `playerbar-bg-active` or `playerbar-bg` depending on the boolean value.

This also needs to be a styled-components `transient` prop which uses the dollar sign `$`. If you don't add the `$`, the prop will be passed on as a HTML DOM element which for custom prop names would be invalid.

[Reference on the docs](https://styled-components.com/docs/api#using-custom-props)

src/renderer/features/shared/components/library-header.module.scss
```diff
@@ -133,6 +135,7 @@
     width: 100%;
     height: 100%;
     background: var(--bg-header-overlay);
+    opacity: 0.6;
```
jeffvli: Adding `opacity: 0.6` conflicts with the original `var(--bg-header-overlay)` when the background image is disabled. Either remove the opacity css value or make it conditional based on album image being enabled.

kgarner7 (approved):

src/renderer/features/shared/components/library-header.tsx
```diff
@@ -64,7 +66,9 @@ export const LibraryHeader = forwardRef(
                     className={styles.background}
                     style={{ background, filter: `blur(${blur ?? 0}rem)` }}
                 />
-                <div className={styles.backgroundOverlay} />
+                <div
+                    className={`${styles.backgroundOverlay} ${(albumBackground && styles.opaqueOverlay) || ''}`}
+                />
```
jeffvli: Can you use the `clsx` library for this instead? Search the repo for clsx for an example.

pyxfluff: Will do when I get the time again which might not be today possibly so unless this is a critical issue I wouldn't let it hold this from being merged

jeffvli: All good, I just edited your branch from my side.

PR #743, opened 2024-09-12
[bugfix]: switch to UTC for release dates

iiPythonx:
This commit switches DayJS to use UTC rather then local time when processing absolute dates, this fixes a bug where albums would show as releasing a day before they really did.

Some quick tests I did:
- Album release (and original) dates are correct.
- `createdAt` and `dateAdded` fields are still correct.

If you see anything wrong let me know, but I've tested every file that uses this function and they all seem to work correctly.

### DIFF
--- a/src/renderer/components/virtual-table/index.tsx
+++ b/src/renderer/components/virtual-table/index.tsx
@@ -42,6 +42,7 @@ import { NoteCell } from '/@/renderer/components/virtual-table/cells/note-cell';
 import { RowIndexCell } from '/@/renderer/components/virtual-table/cells/row-index-cell';
 import i18n from '/@/i18n/i18n';
 import {
+    formatDateAbsolute,
     formatDateAbsoluteUTC,
     formatDateRelative,
     formatSizeString,
@@ -183,7 +184,7 @@ const tableColumns: { [key: string]: ColDef } = {
             GenericTableHeader(params, { position: 'center' }),
         headerName: i18n.t('table.column.dateAdded'),
         suppressSizeToFit: true,
-        valueFormatter: (params: ValueFormatterParams) => formatDateAbsoluteUTC(params.value),
+        valueFormatter: (params: ValueFormatterParams) => formatDateAbsolute(params.value),
         valueGetter: (params: ValueGetterParams) =>
             params.data ? params.data.createdAt : undefined,
         width: 130,
--- a/src/renderer/components/virtual-table/index.tsx
+++ b/src/renderer/components/virtual-table/index.tsx
@@ -41,7 +41,11 @@ import { useFixedTableHeader } from '/@/renderer/components/virtual-table/hooks/
 import { NoteCell } from '/@/renderer/components/virtual-table/cells/note-cell';
 import { RowIndexCell } from '/@/renderer/components/virtual-table/cells/row-index-cell';
 import i18n from '/@/i18n/i18n';
-import { formatDateAbsolute, formatDateRelative, formatSizeString } from '/@/renderer/utils/format';
+import {
+    formatDateAbsoluteUTC,
+    formatDateRelative,
+    formatSizeString,
+} from '/@/renderer/utils/format';
 import { useTableChange } from '/@/renderer/hooks/use-song-change';
 
 export * from './table-config-dropdown';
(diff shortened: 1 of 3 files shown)


kgarner7 (changes requested):
While this is fine for release date, it is _not_ necessarily fine for things like createdAt/updatedAt (date added), which from Navidrome at least include timezone. It would be more correct to conditionally call `local()` for certain locations

iiPythonx:
In that case, would you prefer the default being left as local and then adding a UTC version?
Edit: did that instead.

src/renderer/utils/format.tsx
```diff
@@ -1,14 +1,19 @@
...
 import relativeTime from 'dayjs/plugin/relativeTime';
+import utc from 'dayjs/plugin/utc';
 import formatDuration from 'format-duration';
 import type { Album, AlbumArtist, Song } from '/@/renderer/api/types';
 import { Rating } from '/@/renderer/components/rating';
 
 dayjs.extend(relativeTime);
+dayjs.extend(utc);
 
 export const formatDateAbsolute = (key: string | null) =>
```
kgarner7, line 13: Might as well make the string a const and use it in both rather than duplicate

kgarner7 (changes requested):

src/renderer/utils/format.tsx
iiPythonx: Alright, cleaned that up.

src/renderer/components/virtual-table/index.tsx
```diff
@@ -179,7 +183,7 @@ const tableColumns: { [key: string]: ColDef } = {
             GenericTableHeader(params, { position: 'center' }),
         headerName: i18n.t('table.column.dateAdded'),
         suppressSizeToFit: true,
-        valueFormatter: (params: ValueFormatterParams) => formatDateAbsolute(params.value),
+        valueFormatter: (params: ValueFormatterParams) => formatDateAbsoluteUTC(params.value),
```
kgarner7: This should be absolute, not absolute UTC.

iiPythonx: I thought `dateAdded` only specified a date, with no time information? In that case shouldn't it be processed as UTC?

kgarner7: No. `dateAdded` has time zone (at least for Navidrome)

iiPythonx: Switched it out with the non-UTC version

Optekah:
Thank you for fixing this! I have entries tagged `1983-09` showing up as `Aug 31, 1983`

When this is implemented, will that `1983-09` release now display as `Sept, 1983` or  `Sept 1, 1983` on an album page?

kgarner7:
> Thank you for fixing this! I have entries tagged `1983-09` showing up as `Aug 31, 1983`
> 
> When this is implemented, will that `1983-09` release now display as `Sept, 1983` or `Sept 1, 1983` on an album page?

You can test this live on https://feishin.vercel.app, but in my testing it would be `Sep 1, 1983`. Basically, take the string, treat it as UTC+0 and then display it that way (without time offsets).

Optekah:
> You can test this live on https://feishin.vercel.app

Just tried it and there's indeed a bug present where it incorrectly states it was released `Sept 1, 1983` instead of just `Sept, 1983`.

> Basically, take the string, treat it as UTC+0 and then display it that way (without time offsets).

As someone who has no dev experience, I have no idea what any of this means :P Is it a way for users to fix the above problem?

iiPythonx:
> Just tried it and there's indeed a bug present where it incorrectly states it was released Sept 1, 1983 instead of just Sept, 1983.

I don't recall there ever being an option for it to say `Sept, 1983` rather then including the day. That *could* be added after-the-fact, but nothing you would be able to do (inside Feishin) without writing some more changes.

Optekah:
Checking my Navidrome GUI, it correctly says `Sept, 1983`.

Just took a look at a different release that is tagged with only `YYYY` instead of one that is tagged `YYYY-MM` like my above example, and the way it's displayed is particularly egregious in its incorrectness. It displays as `Jan 1, 2000` even though the band didn't even _exist_ for at least another six or seven months after that— they met and formed sometime around June 2000.

> I don't recall there ever being an option for it to say `Sept, 1983` rather then including the day.

Would it be appropriate if I submit a separate issue for this to be changed / implemented? 

I suppose it comes down to the concept that displaying _no information_ is better than displaying _wrong information_; one I would assume most of us would agree with :P

kgarner7:
https://github.com/jeffvli/feishin/commit/730683fe2542318135c4fb2ce4c78f2f7b81ba41

PR #747, opened 2024-09-16
Lyrics Translation and Romaji (Fulfill #732) [Translation Part]

### LINKED ISSUE #732: Lyrics Translation and Romaji (Phonetic symbol for Japanese songs)

Lyrics Translation and Romaji (Phonetic symbol for Japanese songs)

### Some trivia thought I could think of:
1. I've used NetEase Music. It has translation and romaji. Thus probably possible to scrap translation/romaji at the same time of the lyrics which is already written in the code, but I'm not quite sure whether it's possible to get it with the interface http://music.163.com/api/song/lyric)/api/song/lyric used in the code now. I searched the Internet, but only found information about the interface http://openapi.music.163.com/openapi/music/basic/song/lyric/get/v2.
2. For a much more general solution (since NetEase Music is only a Chinese Music Service Provider, I don't think other language is provided), probably need integration of popular translation APIs, which if user want to use the translation feature needs to enter their API in the settings. If the API is given, then it could be adapted to translate to the language selected by the user.
3. For the Romaji part, I found a library called WanaKana or Kuroshiro, is precisely a convertor for Japanese to Romaji and other notations. There could be two buttons on the lyrics page meaning enable/disable light/dark translation and romaji.
### Reference
1. [The new and the only NetEase API I could find about get lyrics on the official developer website](https://developer.music.163.com/st/developer/document?docId=13ddbaab86854b18903bf73dadf84c0d)
2. [An already done solution for Lyrics Get, API Translation, Romaji Converter, etc. But in C#](https://github.com/jitwxs/163MusicLyrics)
3. [Repo of WanaKana, Javascript library for detecting and transforming between Hiragana, Katakana, and Romaji](https://github.com/WaniKani/WanaKana)
Preview
A coarse draft UI preview I'm thinking of about the lyrics page

### Is this a server-specific feature? (e.g. Jellyfin only)

- [ ] Yes

ENDlezZenith:
Fulfill #732:
- [x] Lyrics Translation
    - [x] Microsoft Azure Translator API for Initial Support (2M Chars Free per Month + Pay as You Go)
    - [x] Google Cloud Translation API for Wider Range of Users (0.5M Chars Free per Month + Pay as You Go)
- [x] Lyrics Romaji (TOO MANY ISSUES, POSTPONED, SPLIT PR)
    - [x] Try Kuroshiro Library from Forks which at Least has Initial TypeScript Support


ENDlezZenith:
Getting a bunch of "Not rewriting GET /@sglkc/kuromoji/dict/filename.gz because the path includes a dot (.) character." in my dev env. I have no idea on fixing this. :(

ENDlezZenith:
> Getting a bunch of "Not rewriting GET /@sglkc/kuromoji/dict/filename.gz because the path includes a dot (.) character." in my dev env. I have no idea on fixing this. :(

This is fixed by adding `disableDotRule: true` in webpack dev server config, but again we got another error. It's like solving an error and getting another one in its dependency, and on, and on in the Romaji Part. I'm thinking of splitting this PR into two halves regarding that the translation part is almost done.

DrENDzZ:
Yep, solve an error and get another one in its dependency, and on, and on with the Romaji Part. I think I'll split this PR into two half. Regarding that the translation part is good now, I'm now on appending AWS, and Google Cloud Translator API for a broader support.

src/main/features/core/lyrics/translate.ts
```diff
@@ -0,0 +1,50 @@
...
+                url: `https://translation.googleapis.com/language/translate/v2?target=${targetLanguage as string}&key=${apiKey}`,
+            });
+            TranslatedText = response.data.data.translations[0].translatedText;
+        } catch (e) {
+            console.error('Google Cloud translate request got an error!', e);
+            return null;
+        }
+    }
+    return TranslatedText;
+};
```
jeffvli: This file should be moved over into the `renderer` since it isn't referenced anywhere in the main process.. Probably under `src/renderer/features/lyrics/queries`.

src/main/features/core/lyrics/translate.ts
```diff
@@ -0,0 +1,50 @@
...
+        try {
+            const response = await axios({
+                data: [
+                    {
+                        Text: originalLyrics,
+                    },
+                ],
+                headers: {
+                    'Content-Type': 'application/json',
+                    'Ocp-Apim-Subscription-Key': apiKey,
```
jeffvli: One issue users might run into is that if the translator resource [is created using a selected region instead of global](https://learn.microsoft.com/en-us/azure/ai-services/translator/reference/v3-0-reference#authenticating-with-a-regional-resource), it will require an additional header value for `Ocp-Apim-Subscription-Region` or the API will return HTTP 401.

This can be handled by either:
- Adding an additional setting when using Azure translator to select the region
- Or -- adding a note in the description that the resource needs to be created using a global region scope rather than regional

I'd be ok with either.

ENDlezZenith: > One issue users might run into is that if the translator resource [is created using a selected region instead of global](https://learn.microsoft.com/en-us/azure/ai-services/translator/reference/v3-0-reference#authenticating-with-a-regional-resource), it will require an additional header value for `Ocp-Apim-Subscription-Region` or the API will return HTTP 401.
> 
> This can be handled by either:
> 
> * Adding an additional setting when using Azure translator to select the region
> * Or -- adding a note in the description that the resource needs to be created using a global region scope rather than regional
> 
> I'd be ok with either.

Thanks for the review, I would probably choose to just make a note about selecting the global endpoint in the apiKey Setting. Both Google and Microsoft recommend using the global scope as it automatically chooses the closed data center. Thus, I think it might be just more beneficially for us common users to use global scope as the connection could be more stable all around the globe? By the way, I've just rechecked docs on specific region scopes of both Google and Microsoft, there just like hundreds of scopes and various second-level domain based on different scope (e.g. Microsoft: America Scopes: api-nam., Asia Pacific Scopes: api-ap., Europe Scopes: api-eur., and for Switzerland, even change in top-level domain. Google: Europe: eu.)

I'll submit a new commit real quick.

src/main/features/core/lyrics/translate.ts
ENDlezZenith: > This file should be moved over into the `renderer` since it isn't referenced anywhere in the main process.. Probably under `src/renderer/features/lyrics/queries`.

Done. `src/renderer/features/lyrics/queries/lyric-translate.ts`

src/renderer/features/settings/components/playback/lyric-settings.tsx
```diff
@@ -116,6 +124,47 @@ export const LyricSettings = () => {
...
+        {
+            control: (
+                <TextInput
+                    value={settings.apiKey}
+                    onChange={(e) => {
+                        setSettings({ lyrics: { ...settings, apiKey: e.currentTarget.value } });
+                    }}
+                />
+            ),
+            description: t('Api key for translation (Support global service endpoint only)'),
```
ENDlezZenith: Modified

jeffvli, line 164: This should be using a translation key instead of the raw string.

jeffvli: This should be using a translation key instead of the raw string.

ENDlezZenith: > This should be using a translation key instead of the raw string.

Ops sorry, forgot to change the description one, committed.

PR #758, opened 2024-09-23
Subsonic 2, general rework

### LINKED ISSUE #306: [Essential] Subsonic API Support
**Why Feishin does need the Subsonic API Now?**
As previously mentioned [over here](https://github.com/jeffvli/feishin/issues/289) there is an essential need to enable the Subsonic API. Navidrome relies upon it heavly all the time, there are basically no clients that dont rely on navidrome's own api due to compability reasons i assume.

**Why there is no reason that we need to wait for Subsonic's API changes?**
After researching the Link that got mentioned on the Githubs Readme turnts out there is no eta or any development on new subsonic api changes so there is basically no need to wait for future subsonic api changes.
Even if someone rebuilds it, it takes up months if not years due to how old subsonic is in some regards.
So i hope this gets fixed as fast as possible.

kgarner7:
Adapted from and supersedes #419. Resolves #306.


kgarner7:
Main weird thing that I haven't quite resolved is updating playlist when an item is removed. Other things _should_ work fine (although fetching large playlists will be. long)

jeffvli:
I might not have the time to go through every change, but as long as it's functional I'm fine with merging.

I'm not seeing the issue with playlist updates when a song is removed (it's working fine for me?), but the playlist list page itself is erroring out due to missing filters for subsonic.

I'll update as I go through it.

jeffvli:
Couple more issues:
- The table view doesn't respect the itemCount in regards to being able to infinitely scroll through all the data (happens on Album and Track list pages from what I can see)
- Double clicking on a track in the track list (table view) seems to be adding an incorrect subset of songs

danjrwalsh:
With this merged, might be a good idea to update the [What music servers does Feishin support](https://github.com/jeffvli/feishin?tab=readme-ov-file#what-music-servers-does-feishin-support) FAQ part of the readme

PR #772, opened 2024-10-01
Add play button to song table album cover, like it is in grid

Egor3f:
Issue: #770

How it looks like now:

image

PR #784, opened 2024-10-09
Some fixes to #772 (Add play button to song table)

Egor3f:
Some fixes to #772

After my pull request was merged, I have noticed that not all Virtual Tables use hook, so context was not passed in some cases. Now fixed that, but cannot edit previous pull request so making a new one instead


jeffvli:
I noticed this as well and was going to tag you, but you got to it first. Thanks!

Issue #841, opened 2024-11-12
Streaming, cover art and other images don't show when the username has URL reserved characters in it


gravelld:
### Expected Behavior

My username on a Subsonic server includes a "+" character. If I can log into a Subsonic server and display my library, the cover art should also display.

I should also be able to stream music.

### Current Behavior

I am able to login to the server and browse my library. However, no cover art displays. Streaming music does not work.

### Steps to Reproduce

1. Create a user on the Subsonic server with a "+" in the username, e.g. `my+test@example.com`
2. Login to the server with Feishin
3. Observe artwork isn't shown
4. Observe streaming doesn't work.

### Possible Solution

The image and streaming query parameters should be URL encoded.

### Context

I consulted the logs in Astiga and observed that most library browsing calls to `getAlbum` et al do have the username URL encoded (I think because Axios is used; this is default behaviour).

However, the calls to `getCoverArt` do not have the query parameters URL encoded. I think that's because the result of [`getCoverArt()`](https://github.com/jeffvli/feishin/blob/21f4a78dd77c72ce532260a1cb484df6363957fb/src/renderer/api/subsonic/subsonic-normalize.ts#L15) is used verbatim in an HTML element.

### Application version

0.11.1

### Operating System and version

Linux

### Server and Version

Astiga

Cambi0nn:
A little add-on (I'm the idiot with + in my email) that streaming itself also doesn't work. @gravelld can better tell if it's caused by only the image call or also by the stream call.

gravelld:
> (I'm the idiot with + in my email)

😄

Ah yes! I forgot to mention the streaming - it also looks like that isn't encoded.

jeffvli:
Will be fixed in the next version

Cambi0nn:
Can it be re-opened? It indeed didn't work after the fix (I had asked @gravelld to look into it as I don't know if it was on Astiga's side or general and I hadn't heard back yet, hence not posting here), but with + not being encoded it also doesn't work.

I would think it would be possible to encode the username only, instead of the whole credentials string?

jeffvli:
@Cambi0nn The `+` should be being encoded properly, at least my initial tests.

Are you able to open the browser devtools to check the actual request URL being sent during authentication?

Cambi0nn:
Yes, the stream call gives me the following response:
```
<subsonic-response status="failed" version="1.16.0" serverVersion="Astiga/production" xmlns="http://subsonic.org/restapi">
[...]

PR #850, opened 2024-11-20
Fix URL encoding on Subsonic

mytja:
Previous commit (https://github.com/jeffvli/feishin/commit/8ec4551b46ff532d3790f23faa63f7ee472ffa33) has been reverted, as it has encoded even equal signs (=), and and signs (&), which should not have been encoded. Nextcloud Music has subsequently failed to receive separate username and password and has therefore failed whilst authenticating the user.

Example of URL beforehand:
https://cloud.example.com/index.php/apps/music/subsonic/rest/stream.view?id=track-4936&v=1.13.0&c=feishin_&u%3Dtest-test%40example.com%26p%3Dpassword

Example of URL now:
https://cloud.example.com/index.php/apps/music/subsonic/rest/stream.view?id=track-4936&v=1.13.0&c=feishin_&u=test-test%40example.com&p=password

Related to #841

### DIFF
--- a/src/renderer/api/subsonic/subsonic-controller.ts
+++ b/src/renderer/api/subsonic/subsonic-controller.ts
@@ -65,7 +65,7 @@ export const SubsonicController: ControllerEndpoint = {
         const cleanServerUrl = `${url.replace(/\/$/, '')}/rest`;
 
         if (body.legacy) {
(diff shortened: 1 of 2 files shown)


jeffvli:
Thanks!

PR #852, opened 2024-11-21
Shutdown handling for windows

Criveratech:
i have no idea how to properly test this outside of windows
lets be honest i dont really know how to properly test this there either
"works on my machine" ¯\_(ツ)_/¯
i was getting annoyed at shutdowns not saving the queue
this is really just the basic example from the electron shutdown plugin, i also had to add the node loader.
i have 0 experience with webpack so what i did is probably not the right way, but it built.


hansyulian:
hey, appreciate the effort, but i also facing the same issue and i already create the fix: https://github.com/jeffvli/feishin/pull/942

if you want to use my branch instead, make a build from this:
https://github.com/hansyulian/feishin/tree/hans-0.14

PR #941, opened 2025-06-04
Feature: Shuffle Button

hansyulian:
I always want to shuffle my playlist song list when i click it, so i make a button to do so

### DIFF
--- a/src/renderer/features/search/components/library-command-item.tsx
+++ b/src/renderer/features/search/components/library-command-item.tsx
@@ -7,6 +7,7 @@ import {
     RiAlbumFill,
     RiPlayFill,
     RiPlayListFill,
+    RiShuffleFill,
     RiUserVoiceFill,
 } from 'react-icons/ri';
 import styled from 'styled-components';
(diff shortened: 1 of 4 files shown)


src/renderer/features/search/components/library-command-item.tsx
```diff
@@ -168,6 +169,19 @@ export const LibraryCommandItem = ({
                 >
                     <RiPlayFill />
                 </Button>
+                <Button
+                    compact
+                    disabled={disabled}
+                    onClick={(e) => handlePlay(e, id, Play.SHUFFLE)}
+                    size="md"
+                    tooltip={{
+                        label: t('player.play', { postProcess: 'sentenceCase' }),
```
jeffvli: The i18n key should be `player.shuffle`, similar to the other ones.

You should also hide the shuffle button for tracks. The shuffle button only makes sense for albums / artists in this menu.

hansyulian: ok

src/renderer/features/search/components/library-command-item.tsx
```diff
@@ -169,19 +168,6 @@ export const LibraryCommandItem = ({
...
-                    onClick={(e) => handlePlay(e, id, Play.SHUFFLE)}
-                    size="md"
-                    tooltip={{
-                        label: t('player.play', { postProcess: 'sentenceCase' }),
-                        openDelay: 500,
-                    }}
-                    variant="default"
-                >
-                    <RiShuffleFill />
-                </Button>
```
jeffvli: You don't need to remove this.

Just do a filter it out for `songs` only. Like `itemType !== LibraryItem.SONG ...`

Issue #934, opened 2025-05-27
Ungraceful exit causes the last play state to be not saved


hansyulian:
### Expected Behavior

For any case of ungraceful exit such as windows shutdown, task manager kill task, and anything else, the last state of the play should at least be preserved to some extent

### Current Behavior

The restored state back to the very start of state of the player because the ungraceful exit didn't manage to save the state

### Steps to Reproduce

1. Open Feishin
2. Play the song for like several minutes or go next song in the list
3. Restart windows
4. Open Feishin again
5. Back to the initial state before the song get played

### Possible Solution

either:
- more robust towards ungraceful exit
- some option to periodically save state, maybe no need to be the very last state to be saved. If we have like periodic 1 minute or 30 seconds, or any configurable time to save the state periodically by calling the function in the `mainWindow.on('close', .... if(!saved && store.get('resume'){      all these codes  }` periodically 

### Context

I shutdown turned on my PC, then play feishin, i heard the song A in my playlist that is currently playing. after several hours i shutdown it after listening to like 30 songs and currently in song XYZ. Then when i open again the feishin after i turn on my PC, i go back to the song A in my playlist again, instead of XYZ or maybe the TUV that was played before XYZ. because of my ungraceful exit the feishin didn't preserve some recent state

### Application version

0.12.7

### Operating System and version

Windows

### Server and Version

Navidrome

mihawk90:
Still relevant (also on Linux)

PR #942, opened 2025-06-04
Fix: Ungraceful shutdown not saving state

hansyulian:
issue details:
https://github.com/jeffvli/feishin/issues/934


hansyulian:
hi @jeffvli , i think this fix would be needed to handle computer shutdown / force close from command line etc

jeffvli:
I'm going to hold off on this for now.

Handling all the edge cases is difficult if we want to actually target 'exit' states across all operating systems.

I may be reworking the queue in the near future which will include persisting the state of the queue on every change rather than just on exit.

hansyulian:
ok i dont mind as i already have my own build. Actually my changes here just do that, to persisting the queue on change rather than exit.

PR #949, opened 2025-06-09
disable visualizer background

pyxfluff:
this is a simple fix made on the fork i help with, i didn't know it wasn't in main until now. does exactly what it says on the tin.


jeffvli:
Can you provide a screenshot of the before and after of what this change does?

I'm testing it out and I don't see a difference.

pyxfluff:
After:

image


Before:
image

PR #951, opened 2025-06-10
lyrics: add translation lyrics for netease.ts

et21ff:
### What is the motivation for this change?

Currently, the NetEase lyrics provider (`netease.ts`) only fetches the original lyrics (`lrc`) and ignores available translated lyrics (`tlyric`). This prevents users from seeing translations for many songs, especially CJK (Chinese, Japanese, Korean) pop music, even when the translations exist on the NetEase platform.

This PR aims to resolve this issue by enabling support for translated lyrics from NetEase.

### What is the new behavior?

- The `getLyricsBySongId` function in `netease.ts` now fetches both original (`lrc`) and translated (`tlyric`) lyrics from the NetEase API.
- If a translation is found, it is parsed and injected as new, fully-formatted LRC lines directly following their corresponding original lines.
- For example, a single original line `[00:12.34]Original Text` becomes two lines in the output string:
  ```lrc
  [00:12.34]Original Text
  [00:12.34]Translated Text
  ```
 This approach was chosen to keep the function's public contract (Promise<string | null>) unchanged, ensuring minimal and non-breaking changes for higher chances of acceptance.

How to test the new behavior?

Check out this branch.

Run the app in production preview mode (pnpm run start).

Play a song that has a known translation on NetEase (e.g., "Lemon" by Kenshi Yonezu, ID: 536622451).

图片

### DIFF
--- a/src/i18n/locales/en.json
+++ b/src/i18n/locales/en.json
@@ -609,6 +609,8 @@
         "mpvExtraParameters_help": "one per line",
         "musicbrainz": "show musicbrainz links",
         "musicbrainz_description": "show links to musicbrainz on artist/album pages, where mbid exists",
+        "neteaseTranslation": "Enable NetEase translations",
+        "neteaseTranslation_description": "When enabled, fetches and displays translated lyrics from NetEase if available.",
         "passwordStore": "passwords/secret store",
         "passwordStore_description": "what password/secret store to use. change this if you are having issues storing passwords.",
         "playbackStyle": "playback style",
--- a/src/i18n/locales/zh-Hans.json
+++ b/src/i18n/locales/zh-Hans.json
@@ -180,6 +180,8 @@
         "followLyric_description": "滚动歌词到当前播放位置",
         "audioExclusiveMode": "音频独占模式",
         "font": "字体",
+        "neteaseTranslation": "启用网易云歌词翻译",
+        "neteaseTranslation_description": "启用后，在获取歌词时将包含并显示网易云音乐提供的翻译（如果存在）。",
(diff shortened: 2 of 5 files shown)


et21ff:
Considering users using netease for lyric search only need a original version of lyric , adding netease_translate.ts may be a probable option

jeffvli:
First of all, thank you for the PR.

This functionality is useful, but it needs to be configurable to be enabled/disabled since there are users who *will not* want to see the included translations.

Please add a settings option in the UI under `lyric-settings`, which includes storing the settings value under `localSettings` as well to access the value in the main process.

et21ff:
> First of all, thank you for the PR.
> 
> This functionality is useful, but it needs to be configurable to be enabled/disabled since there are users who _will not_ want to see the included translations.
> 
> Please add a settings option in the UI under `lyric-settings`, which includes storing the settings value under `localSettings` as well to access the value in the main process.

the setting option is easpecially for netease lyric(cuz only netease has translation lyric) and will it be confilicting with the  existing api-translate option? So i wonder weather its suitable to add it globally in the lyric-settings (or merge it). And maybe we should ask @kgarner7 for further information.

jeffvli:
You just need to make the new setting conditional based on if `netease` is selected as a lyric provider or not.

It should not conflict with any of the existing settings.

et21ff:
> You just need to make the new setting conditional based on if `netease` is selected as a lyric provider or not.您只需要根据是否选择 `netease` 作为歌词提供商来制定新的条件即可。
> 
> It should not conflict with any of the existing settings.它不应该与任何现有设置冲突。

图片
change enable netease translation for test，set false by default

PR #960, opened 2025-06-17
update readme because subsonic is supported now

pyxfluff:
changes as title states, was discussed in the discord a few days prior to this


jeffvli:
Can you update the text and links to OpenSubsonic page instead of the original Subsonic?

PR #973, opened 2025-06-27
fix: discord presence not clearing after pausing player

Lyall-A:
Discord presence was not getting cleared when pausing the player unless it was the beginning of the track, it now clears if:
* No current song
* Player is paused
* Beginning of track

Before:
* No current song
* Player is paused and beginning of track

### DIFF
--- a/src/renderer/features/discord-rpc/use-discord-rpc.ts
+++ b/src/renderer/features/discord-rpc/use-discord-rpc.ts
@@ -25,7 +25,7 @@ export const useDiscordRpc = () => {
             previous: (number | PlayerStatus | QueueSong | undefined)[],
         ) => {
             // No current song, or we switched to a new track and the player was paused (end of album, etc.)
-            if (!current[0] || (current[0] && current[2] === 'paused' && current[1] === 0))
+            if (!current[0] || (current[0] && current[2] === 'paused') || current[1] === 0)
                 return discordRpc?.clearActivity();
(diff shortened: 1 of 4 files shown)


pyxfluff:
There is a "paused" state in the RPC for a reason, you do not need to be clearing it. This should be an option if anything, not something I would really want forced on me

Lyall-A:
I guess it makes more sense when its set to show as "playing Feishin" on Discord, but I don't think it really makes sense when `Show status as listening` is enabled. Maybe it should depend on whether that setting is enabled or not

Lyall-A:
I've updated it to be an option instead, like the show idle option that was removed not long ago. I didn't actually realize that the paused state was intended, I just seen that it said I was listening to a track for 2 hours and assumed there was an issue so my bad. I still think this option should get added even if implemented some different way

src/renderer/features/discord-rpc/use-discord-rpc.ts
jeffvli, line 1: `discordSettings.showPaused` is missing in the dependency array of `setActivity` which is causing issues with the new conditions.

Lyall-A: whoops, fixed

jeffvli: Thanks!

PR #996, opened 2025-07-05
Automatically reconnect to Discord RPC at interval

Lyall-A:
Currently, if Discord is launched after Feishin or the connection was to suddenly be disconnected, you would have to disable and re-enable the integration or restart Feishin for it to connect again

This creates an interval every 10 seconds to check the connection and reconnects if needed

### DIFF
--- a/src/renderer/features/discord-rpc/use-discord-rpc.ts
+++ b/src/renderer/features/discord-rpc/use-discord-rpc.ts
@@ -136,8 +136,17 @@ export const useDiscordRpc = () => {
     useEffect(() => {
         if (!discordSettings.enabled) return discordRpc?.quit();
 
-        discordRpc?.initialize(discordSettings.clientId);
+        let timeout: any;
+        (async function initialize() {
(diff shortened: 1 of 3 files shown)


src/renderer/features/discord-rpc/use-discord-rpc.ts
```diff
@@ -136,8 +136,17 @@ export const useDiscordRpc = () => {
...
+            // Checks RPC status every 10 seconds, initializes if needed
+            const isConnected = await discordRpc?.isConnected();
+            if (!isConnected) {
+                discordRpc?.initialize(discordSettings.clientId);
+            }
+            timeout = setTimeout(initialize, 10000);
+        })();
         return () => {
+            clearTimeout(timeout);
             discordRpc?.quit();
```
jeffvli: I'm personally not a fan of the interval approach.

I would rather have it so that it's checked at the top of `setActivity` so that it validates and initializes (if needed) before every call.

jeffvli (changes requested):

PR #1020, opened 2025-07-23
Added a Linux Notes section below MacOS for desktop.

flamesage:
This adds a section to the README file indicating the AppImage is the recommended way to use Feishin on Linux.

It also provides an example `.desktop` file.

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,7 +63,7 @@ If you're using a Linux device, a `.desktop` file is recommended for easy launch
 
 Download the [latest release (AppImage)](https://github.com/jeffvli/feishin/releases) and [application icon](https://github.com/jeffvli/feishin/blob/development/resources/icon.png?raw=true) to your `~/applications/` folder. This folder may need to be created if it does not already exist.
 
-Rename the AppImage to `Feishin-linux-x86_64.AppImage` and the icon to `Feishin-linux-x86_64.png`.
+Rename the icon to `Feishin-linux-x86_64.png`.
 
 Save the [example desktop file](https://raw.githubusercontent.com/jeffvli/feishin/refs/heads/development/feishin.desktop) as `~/.local/share/applications/feishin.desktop`.
 
--- a/README.md
+++ b/README.md
@@ -57,6 +57,24 @@ If you're using a device running macOS 12 (Monterey) or higher, [check here](htt
 
 For media keys to work, you will be prompted to allow Feishin to be a Trusted Accessibility Client. After allowing, you will need to restart Feishin for the privacy settings to take effect.
 
+#### Linux Notes
+
+If you're using a Linux device, a `.desktop` file is recommended for easy launching of Feishin.
+
+The below file can be saved as `~/.local/share/applications/feishin.desktop` and assumes Feishin is downloaded to the `~/.applications/` folder.
+
+```yaml
+[Desktop Entry]
+Name=Feishin
+Comment=An Electron-based music streaming app
+Exec=/home/username/.applications/Feishin-linux-x86_64.AppImage
+Icon=/home/username/.applications/Feishin-linux-x86_64.png
+Terminal=false
+Type=Application
+Categories=AudioVideo;Music;Player;
+StartupNotify=true
+```
+
 ### Web and Docker
 
 Visit [https://feishin.vercel.app](https://feishin.vercel.app) to use the hosted web version of Feishin. The web client only supports the web player backend.
--- a/README.md
+++ b/README.md
@@ -61,19 +61,13 @@ For media keys to work, you will be prompted to allow Feishin to be a Trusted Ac
 
 If you're using a Linux device, a `.desktop` file is recommended for easy launching of Feishin.
 
-The below file can be saved as `~/.local/share/applications/feishin.desktop` and assumes Feishin is downloaded to the `~/.applications/` folder.
+Download the [latest release (AppImage)](https://github.com/jeffvli/feishin/releases) and [application icon](https://github.com/jeffvli/feishin/blob/development/resources/icon.png?raw=true) to your `~/applications/` folder. This folder may need to be created if it does not already exist.
 
-```yaml
-[Desktop Entry]
-Name=Feishin
-Comment=An Electron-based music streaming app
-Exec=/home/username/.applications/Feishin-linux-x86_64.AppImage
-Icon=/home/username/.applications/Feishin-linux-x86_64.png
-Terminal=false
-Type=Application
-Categories=AudioVideo;Music;Player;
-StartupNotify=true
-```
+Rename the AppImage to `Feishin-linux-x86_64.AppImage` and the icon to `Feishin-linux-x86_64.png`.
+
+Save the [example desktop file](https://raw.githubusercontent.com/jeffvli/feishin/refs/heads/development/feishin.desktop) as `~/.local/share/applications/feishin.desktop`.
+
+You will now see Feishin show up in your menu.
 
 ### Web and Docker
 
--- a/jeffvli-feishin.desktop
+++ b/jeffvli-feishin.desktop
@@ -5,5 +5,5 @@ Exec=/home/username/.applications/Feishin-linux-x86_64.AppImage
 Icon=/home/username/.applications/Feishin-linux-x86_64.png
 Terminal=false
 Type=Application
-Categories=AudioVideo;Music;Player;
+Categories=AudioVideo;Audio;Music;Player;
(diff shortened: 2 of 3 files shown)


README.md
```diff
@@ -57,6 +57,24 @@ If you're using a device running macOS 12 (Monterey) or higher, [check here](htt
 
 For media keys to work, you will be prompted to allow Feishin to be a Trusted Accessibility Client. After allowing, you will need to restart Feishin for the privacy settings to take effect.
 
+#### Linux Notes
+
+If you're using a Linux device, a `.desktop` file is recommended for easy launching of Feishin.
+
+The below file can be saved as `~/.local/share/applications/feishin.desktop` and assumes Feishin is downloaded to the `~/.applications/` folder.
```
kgarner7: Probably worth noting that you should download the release and save it _somewhere_ **without** the version number (as apparently that's important for autoupdate)

README.md
```diff
@@ -57,6 +57,24 @@ If you're using a device running macOS 12 (Monterey) or higher, [check here](htt
...
+
+```yaml
+[Desktop Entry]
+Name=Feishin
+Comment=An Electron-based music streaming app
+Exec=/home/username/.applications/Feishin-linux-x86_64.AppImage
+Icon=/home/username/.applications/Feishin-linux-x86_64.png
+Terminal=false
```
kgarner7: What's the source for the icon? Downloading from `assets/release`?

README.md
```diff
@@ -57,6 +57,24 @@ If you're using a device running macOS 12 (Monterey) or higher, [check here](htt
...
+
+If you're using a Linux device, a `.desktop` file is recommended for easy launching of Feishin.
+
+The below file can be saved as `~/.local/share/applications/feishin.desktop` and assumes Feishin is downloaded to the `~/.applications/` folder.
+
+```yaml
+[Desktop Entry]
+Name=Feishin
+Comment=An Electron-based music streaming app
+Exec=/home/username/.applications/Feishin-linux-x86_64.AppImage
```
kgarner7: For anyone wanting to force Wayland, should probably note adding `--enable-features=UseOzonePlatform,WaylandWindowDecorations --ozone-platform-hint=auto`

README.md
```diff
@@ -57,6 +57,24 @@ If you're using a device running macOS 12 (Monterey) or higher, [check here](htt
 
 For media keys to work, you will be prompted to allow Feishin to be a Trusted Accessibility Client. After allowing, you will need to restart Feishin for the privacy settings to take effect.
 
+#### Linux Notes
+
+If you're using a Linux device, a `.desktop` file is recommended for easy launching of Feishin.
```
kgarner7, line 62: Maybe provide a sample `.desktop` in the root?

kgarner7 (changes requested):
Thanks for starting this. A few thoughts

README.md
flamesage: Yes. Added a link to that in the readme.

README.md
flamesage: Updated the README with the above changes.

README.md
flamesage: I have moved the example from being inline to a separate file `feishin.desktop` at the root of the repository.

flamesage:
Made the above changed as per requested.

README.md
flamesage: Fixed. Decided to add those parameters as Wayland is often default now.

mihawk90: That's probably not a good idea.
It will run on Wayland regardless, this flag doesn't change that.

What it does is run it in native Wayland mode (instead of XWayland). There is likely a reason why Electron doesn't do this by default and requires a flag for it. So adding it to the Desktop file by default is probably not ideal and should be an explicit choice by the user.

flamesage: It's my understanding `--ozone-platform-hint=auto` which is in the current revision will automatically use Wayland if wayland is available, and fallback to Xorg if not present.

mihawk90: Yes? That's not what I said though... I never said anything about the behaviour on X11.

Electron's and therefore Feishin's current default behaviour is running on XWayland under Wayland. This flag as you pointed out makes it so it runs in Wayland native mode.

kgarner7: I am in agreement with @mihawk90, it would be better to have this as a separate `feishin-force-wayland.desktop` (or just note the flags)

flamesage: Believe I've updated this as per the above requirements. If there need to be any additional changes please let me know.

jeffvli-feishin.desktop
```diff
@@ -0,0 +1,9 @@
+[Desktop Entry]
+Name=Feishin
+Comment=An Electron-based music streaming app
+Exec=/home/username/.applications/Feishin-linux-x86_64.AppImage
+Icon=/home/username/.applications/Feishin-linux-x86_64.png
+Terminal=false
+Type=Application
+Categories=AudioVideo;Music;Player;
```
mihawk90: Should probably also include the [`Audio` Main Category](https://specifications.freedesktop.org/menu-spec/latest/category-registry.html#:~:text=multimedia%20(audio/video)-,Audio,-An%20audio%20application) for better compatibility like here:
https://github.com/jeffvli/feishin/blob/d28054cc7f95624c983a2ce1efd1e60650ab4a45/electron-builder.yml#L69

You can keep the additional `Music` category though.

README.md
```diff
@@ -57,6 +57,18 @@ If you're using a device running macOS 12 (Monterey) or higher, [check here](htt
...
+
+Download the [latest release (AppImage)](https://github.com/jeffvli/feishin/releases) and [application icon](https://github.com/jeffvli/feishin/blob/development/resources/icon.png?raw=true) to your `~/applications/` folder. This folder may need to be created if it does not already exist.
+
+Rename the AppImage to `Feishin-linux-x86_64.AppImage` and the icon to `Feishin-linux-x86_64.png`.
```
jeffvli: I've removed the version number from the production AppImage filename so this step can probably be removed.

flamesage: Updated.

jeffvli-feishin.desktop
```diff
@@ -0,0 +1,9 @@
+[Desktop Entry]
+Name=Feishin
+Comment=An Electron-based music streaming app
+Exec=/home/username/.applications/Feishin-linux-x86_64.AppImage
```
mihawk90, line 4: This path uses `~/.applications`, but the readme references `~/applications`.

It should be the same, or better yet, tell the user to adjust their path themselves. They can't use the provided desktop file anyway unless their home-dir happens to be `username`.

I'm not sure if there is a better way to do this. I'm looking at the XDG spec and it's not entirely clear to me whether or not a variable like `$USER` is allowed, which would circumvent this issue.

The [Spec](https://specifications.freedesktop.org/desktop-entry-spec/latest/exec-variables.html) says that `$` is a reserved character and therefore needs to be quoted, which would lead me to believe that using a variable would work, but it doesn't explicitly say whether or not they are allowed.

However, I know from experience that KDE at least has no issue with this and takes pretty much whatever you throw at it, but `desktop-file-validate` vomits errors over it:
```
❯ head ~/.local/share/applications/mpv.desktop -n3             
[Desktop Action open-from-clipboard]
Exec=mpv --player-operation-mode=pseudo-gui -- "$(xsel -ob)"
Name=Open Clipboard URL

      ~                                                                                            at 06:37:04   
❯ desktop-file-validate ~/.local/share/applications/mpv.desktop
/home/tarulia/.local/share/applications/mpv.desktop: error: first group is not "Desktop Entry"
/home/tarulia/.local/share/applications/mpv.desktop: error: value "mpv --player-operation-mode=pseudo-gui -- "$(xsel -ob)"" for key "Exec" in group "Desktop Action open-from-clipboard" contains a non-escaped character '$' in a quote, but it should be escaped with two backslashes ("\\$")
```

When I do escape the `$`, this doesn't work anymore, so that would indicate it's not allowed.

But then again, this whole thing also doesn't work when I fix the second error it complains about so... probably just quirky KDE behaviour that is maybe technically a bug that helps me.

---

As discussed on Discord before if we wanted a "ready to go" type of desktop file I don't see a way around a shell script that downloads the AppImage and writes a desktop file from scratch (or from a template), inserting the username, possibly giving users the option to provide their own directory (for example my AppImages are in `~/Applications`, not `~/applications` or `~/.applications`), and the option to use the wayland-native switches.

The readme would then just have a `curl <link> | sh` in it.

mihawk90 (changes requested):
I don't think it's that easy unfortunately :melting_face:

Issue #1022, opened 2025-07-25
[Feature]: Tab support for action in search


Der-Penz:
### I have already checked through the existing feature requests and found no duplicates

- [x] Yes

### Is this a server-specific feature?

Not server-specific

### What do you want to be added?

When opening the search window i can search for a song and select it with the arrow keys up and down but I can play it with a hotkey or select the play, add last or add next button with tab. Only when the mouse hovers a specific search result those keys appear but not via keyboard navigation.

It would be nice if keyboard navigation triggers the play, add last or add next button to appear which would allow me to select the buttons via the tab key or make an hotkey to play, ... the selected item. Pressing enter currently opens the selected item but does not play it directly

<img width="803" height="829" alt="Image" src="https://github.com/user-attachments/assets/8e62bdb5-e3e7-4e6f-aa39-4adc0c19b06b" />
Track selected via keyboard navigation but buttons do not appear

<img width="791" height="845" alt="Image" src="https://github.com/user-attachments/assets/6141840d-17d9-4c7d-b6ef-1728ee6b62d3" />
with the mouse hovering over it the buttons appear (mouse is not visible because of the screenshot but the mouse is on the "Time (extended)" item.

PR #1024, opened 2025-07-26
Support tab navigation on ActionIcons in command palette

Der-Penz:
Solution for #1022 

### Added
When selecting an item with the arrow keys in the command palette the action keys will now become visible to. By using tab you are able to focus them and enter presses them.

image
selected Album "The Search" with the arrow keys and by pressing tab the play button gets selecte (see the slightly brighter background)


jeffvli:
Looks good, thanks!

PR #1034, opened 2025-07-31
add private mode toggle to app menu

Lyall-A:
This adds a new button to the app menu that toggles a private mode to disable scrobbling and discord activity, like Spotify's private session mode

image


jeffvli:
I like the feature, but one thing I would recommend doing is making it more obvious within the app whenver Private Mode is turned on.

For example, adding it to the title bar text, changing the icon color (red/green) in the app menu depending on whether it's on/off, etc.

Lyall-A:
> I like the feature, but one thing I would recommend doing is making it more obvious within the app whenver Private Mode is turned on.
> 
> For example, adding it to the title bar text, changing the icon color (red/green) in the app menu depending on whether it's on/off, etc.

I've made a few changes to make it more obvious

* It now shows in the title bar if its enabled
* It shows red in the app menu if its enabled
* It shows a toast message on enable/disable to make it more obvious what the function actually does

let me know if anything else needs changed

Issue #649, opened 2024-06-22
Windows 10 & 11 SMTC integration


EvilSquirrelGuy:
### What do you want to be added?

It would be useful if Feishin was able to communicate with Windows' System Media Transfer Controls (SMTC) API for controlling music playback, like other apps that can play media (e.g. Spotify, Firefox), which is good for having a visual indicator whenever tracks are changed.

### Is this a server-specific feature? (e.g. Jellyfin only)

- [ ] Yes

bmarinov:
As an added note, the impact of not supporting this is not purely visual. I often change between listening on Spotify and Feishin throughout the day. 
Using the media keys would often trigger spotify and start playing it in the background while Feishin is also playing music. Fixing this involves switching between apps, exiting Spotify etc. While listening to two tracks at the same time :D

PikaDude:
+1 I really hope this is added, it's my only gripe with Feishin.

AAMBIENTT:
Please fix this, May be the only reason i move to something else.

yakoodev:
+1

PR #1040, opened 2025-08-02
enable mediaSession api

Der-Penz:
This PR enables the navigator.mediaSession api so windows will display information about the current song next to the volume bar.

What changed: 
HardwareMediaKeyHandling, MediaSessionService features are not disabled anymore to make the mediaSession api work.

I plan to implement more events for the mediaSession api like next previous. 
But I am not sure what the reason was to disable those 2 features. Maybe there is a good reason for it and enabling it would break stuff. In that case I would look into it, on how it could be resolved since the mediaSession is a nice to have Feature.


Der-Penz:
This should also close #649 when implemented

mihawk90:
This will have to be handled platform specific as enabling this outright will conflict with MPRIS on Linux at least. Don't know about Mac.

That aside there is probably a reason why `HardwareMediaKeyHandling` is on the disabled list as well, and this PR also removes that without explanation.

Leaving the line in the file as a comment also serves no purpose. If you just want to keep it for reference as how it used to be, that is what Git is for. If you're going to remove a line, remove it completely.

kgarner7:
This was discussed in the Discord (/matrix bridge).
The _correct_ mechanism would be to just use MediaSession API for Electron (when not using MPV player).

Der-Penz:
> This will have to be handled platform specific as enabling this outright will conflict with MPRIS on Linux at least. Don't know about Mac.

One simple approach would be to enable it only on Windows via a process.platform check.
This would at least make the feature work on Windows without touching the Linux/Mac flows.

Alternatively, we could make this configurable for users who want navigator.mediaSession integration, while keeping it off by default on platforms where conflicts are known or unverified.

jeffvli:
I would prefer you add a toggle in the settings to Enable/Disable this feature, as there are some (including myself) that wouldn't want this enabled.

Toggling the setting should prompt the user to restart the app (similar to some other settings) since it's configured within the electron main process.

PR #1078, opened 2025-08-17
[feature]: Add Equalizer

0PandaDEV:
This is a revived PR from https://github.com/jeffvli/feishin/pull/308 I have attempted to fix as many merge conflicts as possible and could get it working in dev.

Values from [this message](https://discord.com/channels/922656312888811530/922656312888811535/1162427713567596666).
Works for both Electron and [modern browsers](https://caniuse.com/?search=createBiquadFilter)

Notes:
- it might be nice to have more filters, but I'm not touching that. Getting the Q value wrong leads to bad distortion
- it would also be nice to have low/highshelf for web/mpv, but when I tried the filter was considerably stronger in web audio
- better visualizer/UI? Getting the number input to behave was frustrating enough...
- *please* test this. I am not confident in audio processing.


package.json
```diff
@@ -120,6 +121,7 @@
         "react-window": "^1.8.9",
         "react-window-infinite-loader": "^1.0.9",
         "semver": "^7.5.4",
+        "styled-components": "^6.1.19",
```
kgarner7, line 124: Do not bring in `styled-components`. This was removed and should stay removed.

bun.lock
kgarner7, line 1: Remove this file. `bun` is not used anymore

kgarner7 (changes requested):
Initial requirements. I have not reviewed whether the other functionality is ok

pyxfluff:
I checked functionality myself and it did not seem to work aside from a few typos and other mistakes I had to fix. I applied the settings below and my audio still sounds the exact same from the web player.

image

0PandaDEV:
Yes the problem os it only works with mpv at the moment

kgarner7:
For web audio, see inspiration from #308. This will _also_ be required before approval

PR #1084, opened 2025-08-22
Feature: Search on playlist

hansyulian:
## Description
I have some playlist with too many songs and tends to keep growing:
- Worship
- Praise
As this religious song playlist tends to keep growing and very big, I'm having difficulty to find out whether a certain song is already in the playlist or not. 

So I create this feature to be able to search the song based on `name`, `artist`, and `album` within a playlist. 

The functionality itself will not filter the songs up added to the play queue. 

## Proof Of Work
feishin

### DIFF
--- a/src/main/features/core/lyrics/lrclib.ts
+++ b/src/main/features/core/lyrics/lrclib.ts
@@ -10,7 +10,7 @@ import {
 import { orderSearchResults } from './shared';
 
 const FETCH_URL = 'https://lrclib.net/api/get';
-const SEEARCH_URL = 'https://lrclib.net/api/search';
+const SEARCH_URL = 'https://lrclib.net/api/search';
 
 const TIMEOUT_MS = 5000;
 
@@ -59,7 +59,7 @@ export async function getSearchResults(
     }
 
     try {
-        result = await axios.get<LrcLibSearchResponse[]>(SEEARCH_URL, {
+        result = await axios.get<LrcLibSearchResponse[]>(SEARCH_URL, {
             params: {
                 q: params.name,
             },
--- a/src/renderer/features/playlists/components/playlist-detail-song-list-header-filters.tsx
+++ b/src/renderer/features/playlists/components/playlist-detail-song-list-header-filters.tsx
@@ -4,7 +4,7 @@ import { IDatasource } from '@ag-grid-community/core';
 import { closeAllModals, openModal } from '@mantine/modals';
 import { useQueryClient } from '@tanstack/react-query';
 import debounce from 'lodash/debounce';
-import { MouseEvent, MutableRefObject, useCallback } from 'react';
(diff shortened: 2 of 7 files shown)


src/renderer/hooks/use-deferrer.ts
jeffvli, line 1: Instead of a custom function/hook for this, just wrap the [useDebouncedCallback](https://mantine.dev/hooks/use-debounced-callback/) function from Mantine. I'm attempting to utilize as many of the built in hooks as possible. With this change, the `wait` function can also be removed.

src/shared/utils/is-any-fields-contains-string.ts
jeffvli, line 1: Since you're doing a client-side search, you should take advantage of [fuse.js](https://www.fusejs.io/) to enable fuzzy matching on the selected fields. Fuse.js is already utilized in other parts of the app (such as lyrics matching).

src/renderer/features/playlists/components/playlist-detail-song-list-header-filters.tsx
```diff
@@ -328,16 +338,26 @@ export const PlaylistDetailSongListHeaderFilters = ({
...
+                        const lowerCaseSearchTerm = searchTerm?.toLowerCase() || '';
+                        const filteredItems =
+                            songsRes?.items.filter((record) =>
+                                isAnyFieldsContainsString(
+                                    record,
+                                    SEARCH_FIELDS,
+                                    lowerCaseSearchTerm,
+                                ),
+                            ) || [];
+                        // HY: not optimal if the data is paginated. if anyone knows how to do server side filtering, then please let me know
```
jeffvli, line 355: This might actually be a blocker. The workaround would be to remove server-side pagination and fetch the entire playlist at once so that client-side pagination works more effectively. I was already thinking about reworking the API to handle that, but I hadn't gotten around to it yet. I may put this PR on hold until that is implemented.

src/main/features/core/lyrics/lrclib.ts
```diff
@@ -10,7 +10,7 @@ import {
 import { orderSearchResults } from './shared';
 
 const FETCH_URL = 'https://lrclib.net/api/get';
-const SEEARCH_URL = 'https://lrclib.net/api/search';
+const SEARCH_URL = 'https://lrclib.net/api/search';
```
kgarner7, line 13: This is a separate issue, and should probably be done in a separate PR (or I can just fix this)

kgarner7 (changes requested):
Given that playlists are now always client-side, a standard implementation should be much easier to do now.

kgarner7:
Playlist search is now implemented (and other playlist refactoring has been done). Superseded by 306167fee300a55188bd230490cf0b156086cdc5

PR #1085, opened 2025-08-22
fix: save state upon song change to prevent playlist state missing upon ungraceful termination

hansyulian:
## Description
Currently, when the app is terminated by any non exit of the app such as:
- computer shut down
- lose of power
- force termination by task kill
- any other external app termination event
it will lose the last played song as the current implementation only `save-queue` when the app exit. To fix this behaviour, I moved the `save-queue` on  change play queue which includes:
- adding/removing/replacing queue list
- change song

## How to replicate issue
1. open app, pay attention on what is the current song, for example X
2. change song by changing the play queue or play any current song, for example Y
3. terminate using task manager, task kill, or anything
4. open again the app
5. you will see X instead of Y

In this comparison i kill using task manager and reopen again and see how the difference of the song after reopen on the before and after

## Before
before

## After
after


src/renderer/app.tsx
```diff
@@ -60,6 +59,44 @@ export const App = () => {
     const { clearQueue, restoreQueue } = useQueueControls();
     const remoteSettings = useRemoteSettings();
     const cssRef = useRef<HTMLStyleElement | null>(null);
+    const reloadStateHandled = useRef(false);
+    const playerQueue = usePlayerStore((state) => state.queue);
+    const playerCurrent = usePlayerStore((state) => state.current);
```
jeffvli: There's a major performance impact from using `usePlayerStore((state) => state.current);` since the `current` state includes `currentTime` which is updated very frequently. This results in the entire `App` tree re-rendering every 0.5-1 seconds.

If you want to use values from the `current` state, you will need to use a custom selector with only the values you need.

I have included an example with react-scan showing the re-renders when the player is running.

https://github.com/user-attachments/assets/04ed159e-a367-404c-ab99-7b86a34c6b55

kgarner7: I've been holding off on pushing this change to `develop` (1: because there are other changes coming in, and 2: because Subsonic `savePlayQueue` isn't great, and unfortunately the extension will probably never get fixed .-.) but what I personally do is something like https://github.com/kgarner7/feishin/commit/2cf2adaee1e671acf1f43500f148999eea084190 (autosave every X songs, configurable by user)

hansyulian: wow @jeffvli that's very helpful! i'm glad i learned something new, the react-scan is going to be a very helpful tool. right now i always do manual performance check but ofc it will have some limitation. i have updated the PR with the performance fix

hansyulian: pow

jeffvli: I'm thinking that additionally a good idea would be to extract this functionality to a custom hook and put it further down the application tree so that we can prevent re-renders for the full app on every song change.

jeffvli: Sorry, I might not have been clear enough.

When I said "extract this functionality to a custom hook", I was expecting to extract all of the functionality added here into a hook for example `useSaveQueue`. Then you would add the hook to a component/file that's deeper in the tree that is always visible in the app (for example the PlayerBar). That way only the playerbar would re-render instead of the entire app whenever the `useSaveQueue` runs.

jeffvli:
Player state is now saved using indexeddb on all changes.

Issue #1041, opened 2025-08-02
[Feature]: Remove Electron Menu Bar


YaBoiMax:
### I have already checked through the existing feature requests and found no duplicates

- [x] Yes

### Is this a server-specific feature?

Not server-specific

### What do you want to be added?

I can't find anyone else raising this, maybe githubs search sucks, but please remove the Electron Menu bar that pops up when you press Alt.
It looks very annoying when I have global shortcuts that use Alt and a menu bar appears and disappears.

<img width="679" height="120" alt="Image" src="https://github.com/user-attachments/assets/ab1bc893-b8dd-4d09-b385-10bde7b628b6" />

<img width="686" height="166" alt="Image" src="https://github.com/user-attachments/assets/9ab14962-f967-4db6-8ba9-0cf87d9c8832" />

PR #1113, opened 2025-09-09
fix: electron menu bar for darwin devices

maximelafarie:
Since @jeffvli's commit 20debb0 in the v0.20.0, it completely disabled any `cmd`+`key` shortcut in the entire app, including quit, copy, paste, select all, and so on : 

image
image

(screenshots without and with the fix)

Since the menu bar is not included in the electron window on Darwin devices, issue #1041 does not apply and is therefore a regression.

I fixed this by applying menu bar masking only to non-Darwin devices.

~~I used `setMenuBarVisibility` to allow non-Darwin users who wish to do so to use the menu bar if necessary. By default, it will be invisible and can be redisplayed using the `Alt` key.~~

This restores all shortcuts and benefits in terms of convenience and accessibility.

Not to mention: it allows you to paste extremely complicated passwords from password managers when entering server information, for example.

### DIFF
--- a/src/main/index.ts
+++ b/src/main/index.ts
@@ -489,7 +489,10 @@ async function createWindow(first = true): Promise<void> {
 
     const menuBuilder = new MenuBuilder(mainWindow);
     menuBuilder.buildMenu();
-    Menu.setApplicationMenu(null);
+
+    if (process.platform !== 'darwin') {
+        Menu.setApplicationMenu(null);
+    }
 
     // Open URLs in the user's browser
     mainWindow.webContents.setWindowOpenHandler((edata) => {


src/main/index.ts
```diff
@@ -489,7 +489,10 @@ async function createWindow(first = true): Promise<void> {
 
     const menuBuilder = new MenuBuilder(mainWindow);
     menuBuilder.buildMenu();
-    Menu.setApplicationMenu(null);
+
+    if (process.platform !== 'darwin') {
+        mainWindow?.setMenuBarVisibility(false);
```
jeffvli: [setMenuBarVisibility](https://www.electronjs.org/docs/latest/api/browser-window#winsetmenubarvisibilityvisible-windows-linux) only affects the auto-hide functionality of the menu bar, so you'll need to replace it with `Menu.setApplicationMenu(null)`.

jeffvli (changes requested):

src/main/index.ts
maximelafarie: You're absolutely right, it's done.

PR #1121, opened 2025-09-10
Clicking outside a context menu closes it

### LINKED ISSUE #1120: [Bug]: Context Menu doesn't close when clicked away

- [x] Yes

### App Version

0.20.0

### Music Server and Version

Jellyfin v.10.10.7

Desktop Linux

### What happened?

Context menus persist even after clicking outside the menu. I would personally expect the menu to disappear, but the menu doesn't go away unless I click something in the menu. If I want to cancel out of the context menu without taking any action, I have to click "Get Info" and then escape.

Image

### Steps to reproduce

1. Open any context menu (right click, or left click any of the vertical/horizontal ... symbols).
2. Try clicking outside of the context menu.
3. Context menu should close (at least, in my opinion).

### Relevant log output

```shell

```

iparks99:
Fixes https://github.com/jeffvli/feishin/issues/1120.

As title says, clicking outside of any context menu now closes the context menu.

### DIFF
--- a/src/renderer/features/context-menu/context-menu-provider.tsx
+++ b/src/renderer/features/context-menu/context-menu-provider.tsx
@@ -114,7 +114,6 @@ export const ContextMenuProvider = ({ children }: ContextMenuProviderProps) => {
     const clickOutsideRef = useClickOutside(
         () => setOpened(false),
         ['mousedown', 'touchstart'],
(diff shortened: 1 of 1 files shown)


jeffvli:
Thanks for the PR. We decided to handle it (with a couple additional changes) in #1122.

PR #1125, opened 2025-09-13
client-side only sort for all playlists

### LINKED ISSUE #829: Sorting returns the same order no matter what
### Expected Behavior

When sorting a playlist, the items should appear in a specific order, sorted by that property.

### Current Behavior

The order of the playlist is the same, no matter what I sort by, it simply returns it sorted by ID.

### Steps to Reproduce

https://github.com/user-attachments/assets/7943e911-d475-49d2-a2a6-2cddacec323a

### Possible Solution

I tried installing a Navidrome server and the issue did not occur, but transferring my playlist(s) over is a huge pain, and I generally prefer Jellyfin, since it also hosts my other media.

### Context

This bug happened with the playlist in the video, and a separate one used for testing, with just 2 songs in it. (On jellyfin)
I tried using Sonixd and it didn't work either, I did get it to sort properly on Supersonic, but the rest of that client is unusable.

### Application version

v0.11.1

### Operating System and version

Windows 11 Desktop

### Server and Version

Jellyfin v10.10.1

### Node Version (if developing locally)

_No response_

### LINKED ISSUE #1015: [Bug]: Playlist only uses Id position to play songs

- [x] Yes

### App Version

0.18.0

### Music Server and Version

navidrome v0.57.0

Desktop Windows

### What happened?

Image

### Steps to reproduce

go to a playlist 
sort by somthing that changes postions of song
press play

### Relevant log output

```shell

```

kgarner7:
Resolves #829 
Resolves #1015 
Use client side sorting for all applications.
TODO:
- [x] handle reordering items better (maybe don't invalidate the entire query and just reorder locally? idk)
- [x] use the client side sort when adding playlist to queue?
- [ ] other stuff I definitely missed
- [ ] Subsonic support for reordering items in playlist [maybe]

### DIFF
--- a/src/renderer/api/jellyfin/jellyfin-controller.ts
+++ b/src/renderer/api/jellyfin/jellyfin-controller.ts
@@ -542,10 +542,6 @@ export const JellyfinController: ControllerEndpoint = {
             query: {
                 Fields: 'Genres, DateCreated, MediaSources, UserData, ParentId, People, Tags',
                 IncludeItemTypes: 'Audio',
-                Limit: query.limit,
-                SortBy: query.sortBy ? songListSortMap.jellyfin[query.sortBy] : undefined,
-                SortOrder: query.sortOrder ? sortOrderMap.jellyfin[query.sortOrder] : undefined,
-                StartIndex: query.startIndex,
                 UserId: apiClientProps.server?.userId,
             },
         });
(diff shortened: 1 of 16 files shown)


jeffvli (approved):

jeffvli:
Looks good to me, unless you want to put in the effort to add the subsonic re-ordering. Definitely not needed though.

kgarner7:
> Looks good to me, unless you want to put in the effort to add the subsonic re-ordering. Definitely not needed though.

If the preference is to instead _always_ just override the playlist (to deal with Jellyfin currently being broken, and allow dragging multiple items), then that should probably come at the same time.
But also, for Subsonic you'd _really_ `POST` support

PR #1129, opened 2025-09-15
Feature: added silent song notification setting

ilgabbo:
I added this setting to mute the default sound that the system notification emits when another song is played.
The setting is only visible when the “Enable song notification” setting is enabled, under the Playback section.

https://github.com/user-attachments/assets/30763328-5575-46d3-8e0f-b2bde652b5df

### DIFF
--- a/src/i18n/locales/it.json
+++ b/src/i18n/locales/it.json
@@ -403,8 +403,6 @@
         "hotkey_navigateHome": "vai alla schermata iniziale",
         "notify": "abilita notifiche delle tracce",
         "notify_description": "mostra una notifica quando cambia la traccia riprodotta",
-        "silentNotification": "Notifica silenziosa",
-        "silentNotification_description": "non riprodurre il suono della notifica",
         "preventSleepOnPlayback": "non sospendere in riproduzione",
         "preventSleepOnPlayback_description": "non sospendere il sistema quando la riproduzione è attiva",
         "discordDisplayType": "stile dello stato su {{discord}}"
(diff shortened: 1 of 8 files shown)


jeffvli:
No need for the additional toggle setting. 

Just set it to silent by default - I don't see a use case where having a notification sound played on song change is necessary. This also removes the need to add localization.

src/renderer/features/shared/components/library-header-bar.tsx
```diff
@@ -43,6 +44,16 @@ const HeaderBadge = ({ children, isLoading, ...props }: HeaderBadgeProps) => {
...
 
+interface PlayListDurationBadgeProps {
+    time?: null | number;
+}
+
+const PlayListDurationBadge = ({ time }: PlayListDurationBadgeProps) => {
+    if (time === null || time === undefined) return <SpinnerIcon />;
+    return <Badge>{dayjs(time).format('HH:mm:ss')}</Badge>;
+};
+
```
jeffvli: Please open a separate PR for this, since it's unrelated to the notifications.

src/i18n/locales/it.json
```diff
@@ -322,6 +322,7 @@
         "albumBackground_description": "aggiunge un'immagine di sfondo per le pagine degli album contenenti l'album art",
         "albumBackgroundBlur": "intensità sfocatura immagine di sfondo dell'album",
         "albumBackgroundBlur_description": "regola la quantità di sfocatura applicata all'immagine di sfondo dell'album",
+
```
jeffvli: Remove this extra newline.

src/renderer/features/shared/components/library-header-bar.tsx
ilgabbo: you're right, I made a little mistake, sorry 😅

mihawk90:
This should probably be rebased/squashed to remove the 8 unnecessary commits.

ilgabbo:
> This should probably be rebased/squashed to remove the 8 unnecessary commits.

I could close this one and open another one with a single commit

jeffvli:
No need, it can be squashed via the GitHub merge process.

PR #1130, opened 2025-09-16
Feature: added playlist duration badge

ilgabbo:
Adds a badge with the playlist duration next to the playlist title.

image

### DIFF
--- a/src/renderer/features/playlists/components/playlist-detail-song-list-header.tsx
+++ b/src/renderer/features/playlists/components/playlist-detail-song-list-header.tsx
@@ -11,6 +11,7 @@ import { usePlaylistDetail } from '/@/renderer/features/playlists/queries/playli
 import { FilterBar, LibraryHeaderBar } from '/@/renderer/features/shared';
 import { useCurrentServer } from '/@/renderer/store';
 import { usePlayButtonBehavior } from '/@/renderer/store/settings.store';
+import { formatDurationString } from '/@/renderer/utils';
 import { Badge } from '/@/shared/components/badge/badge';
 import { SpinnerIcon } from '/@/shared/components/spinner/spinner';
 import { Stack } from '/@/shared/components/stack/stack';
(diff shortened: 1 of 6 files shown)


src/renderer/features/shared/components/library-header-bar.tsx
```diff
@@ -43,6 +47,16 @@ const HeaderBadge = ({ children, isLoading, ...props }: HeaderBadgeProps) => {
     return <Badge {...props}>{isLoading ? <SpinnerIcon /> : children}</Badge>;
 };
 
+interface PlayListDurationBadgeProps {
```
kgarner7: There is already `format-duration` (and `formatDurationString`), it would be more appropriate to use one of the existing tools to get the duration.

src/renderer/features/shared/components/library-header-bar.tsx
```diff
@@ -43,6 +47,16 @@ const HeaderBadge = ({ children, isLoading, ...props }: HeaderBadgeProps) => {
...
+    time?: null | number;
+}
+
+const PlayListDurationBadge = ({ time }: PlayListDurationBadgeProps) => {
```
kgarner7: Not sure why this is a separate component? This could be easily inlined to the playlist component.

kgarner7 (changes requested):

src/renderer/features/shared/components/library-header-bar.tsx
ilgabbo: Hi, the idea was to make the display of the playlist duration optional via a setting, which was subsequently removed because it was effectively useless.
I pushed the new version without the component.

src/renderer/features/shared/components/library-header-bar.tsx
ilgabbo: Hi, I made the change with formatDurationString, but in my opinion it would be more useful to view #1131 as it already incorporates my change and close my pr.

jeffvli: I think you misunderstood? #1131 is a feature request, not a PR.

ilgabbo: Yeah i know #1131 is a feature request, my proposal was to ignore my PR if #1131 was accepted, since it already includes my feature

src/renderer/features/playlists/components/playlist-detail-song-list-header.tsx
```diff
@@ -52,6 +53,9 @@ export const PlaylistDetailSongListHeader = ({
                 <LibraryHeaderBar>
                     <LibraryHeaderBar.PlayButton onClick={() => handlePlay(playButtonBehavior)} />
                     <LibraryHeaderBar.Title>{detailQuery?.data?.name}</LibraryHeaderBar.Title>
+                    {detailQuery?.data?.duration && (
+                        <Badge>{formatDurationString(detailQuery.data!.duration!)}</Badge>
+                    )}
```
jeffvli: [Short-circuit](https://stackoverflow.com/questions/53048037/react-showing-0-instead-of-nothing-with-short-circuit-conditional-component) issue here when `detailQuery.data.duration` is 0.

image

PR #1138, opened 2025-09-19
feat: add semantic selectors for now-playing media

inverse:
This change adds unique class names to the elements that display the currently playing media information. This makes it easier for extension developers to parse the DOM and understand what media is playing.

It'll allow us to do things like https://github.com/web-scrobbler/web-scrobbler/pull/5612 easier since elements will be semantically available. 

The following classes have been added:
- `media-player`: The main player container.
- `player-cover-art`: The cover art of the playing track.
- `song-title`: The title of the playing track.
- `song-artist`: The artist of the playing track.
- `song-album`: The album of the playing track.
- `player-state-playing`/`player-state-paused`: The state of the player.
- `elapsed-time`: The elapsed time of the playing track.
- `total-duration`: The total duration of the playing track.

I did test the changes locally against the extension using the tags and it did work. However not sure if what was changed is inline with how it should be structured.

### DIFF
--- a/src/renderer/features/player/components/center-controls.tsx
+++ b/src/renderer/features/player/components/center-controls.tsx
@@ -5,6 +5,7 @@ import isElectron from 'is-electron';
 import { useEffect, useState } from 'react';
 import { useTranslation } from 'react-i18next';
 
+import { PlaybackSelectors } from '../../../../shared/constants/playback-selectors';
 import styles from './center-controls.module.css';
 
 import { PlayButton, PlayerButton } from '/@/renderer/features/player/components/player-button';
@@ -258,7 +259,13 @@ export const CenterControls = ({ playersRef }: CenterControlsProps) => {
             </div>
             <div className={styles.sliderContainer}>
                 <div className={styles.sliderValueWrapper}>
-                    <Text className="elapsed-time" fw={600} isMuted isNoSelect size="xs">
+                    <Text
+                        className={PlaybackSelectors.elapsedTime}
+                        fw={600}
+                        isMuted
+                        isNoSelect
+                        size="xs"
+                    >
                         {formattedTime}
                     </Text>
                 </div>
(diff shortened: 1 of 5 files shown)


jeffvli:
I'm not opposed to this, but adding plain classes is a bit prone to breakage if these components ever get refactored.

At the very least I would say add them to an enum with comments on how they're being used, and then reference them in the components.

inverse:
Thanks for the feedback . I moved the selectors to consts and added a node. Would there be anything else I should change?

src/renderer/features/player/components/center-controls.tsx
```diff
@@ -172,6 +173,11 @@ export const CenterControls = ({ playersRef }: CenterControlsProps) => {
                         />
                     )}
                     <PlayButton
+                        className={
+                            status === PlayerStatus.PAUSED
+                                ? 'player-state-paused'
+                                : 'player-state-playing'
```
jeffvli: This also needs to be updated to use PlaybackSelectors.

src/renderer/features/player/components/center-controls.tsx
```diff
@@ -5,6 +5,7 @@ import isElectron from 'is-electron';
 import { useEffect, useState } from 'react';
 import { useTranslation } from 'react-i18next';
 
+import { PlaybackSelectors } from '../../../../shared/constants/playback-selectors';
```
jeffvli: Can you update the import to use the absolute path similar to the others? `/@/shared/...`

src/renderer/features/player/components/left-controls.tsx
```diff
@@ -5,6 +5,7 @@ import React, { MouseEvent } from 'react';
 import { useTranslation } from 'react-i18next';
 import { generatePath, Link } from 'react-router-dom';
 
+import { PlaybackSelectors } from '../../../../shared/constants/playback-selectors';
```
jeffvli: Same thing here for absolute path import.

src/renderer/features/player/components/playerbar.tsx
```diff
@@ -56,7 +57,7 @@ export const Playerbar = () => {
 
     return (
         <div
-            className={styles.container}
+            className={`${styles.container} ${PlaybackSelectors.mediaPlayer}`}
```
jeffvli: Nitpicking again, but use `clsx` here similar to the other files.

inverse: Updated and moved PlayButton class into component itself

PR #1146, opened 2025-09-23
Add autodiscovery for Jellyfin servers

hedgehog1029:
I was disappointed that Feishin didn't auto-detect my jellyfin server on launch, so I implemented it :3

- Autodiscovery fires automatically and results are streamed back to the UI as they arrive
- Only implemented for Jellyfin, but space left in the implementation if the other servers secretly have autodiscovery methods too (I couldn't find any from a cursory look)
- Click a result to autofill the name and URL into the setup form
- UI could maybe do with a touch up

### DIFF
--- a/src/main/features/core/autodiscover/index.ts
+++ b/src/main/features/core/autodiscover/index.ts
@@ -1,11 +1,16 @@
 import { createSocket } from 'dgram';
 import { ipcMain } from 'electron';
+
 import { DiscoveredServerItem, ServerType } from '/@/shared/types/types';
 
 type JellyfinResponse = {
     Address: string;
     Id: string;
     Name: string;
+};
+
+function discoverAll(reply: (server: DiscoveredServerItem) => void) {
+    return Promise.all([discoverJellyfin(reply)]);
 }
 
 function discoverJellyfin(reply: (server: DiscoveredServerItem) => void) {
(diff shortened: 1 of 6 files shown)


kgarner7 (changes requested):
1. Multiple linter errors are present. Run `pnpm lint:fix`, and ignore the warning in `context-menu-provider.tsx`
2. This does not appear to work if Jellyfin is on the same system as Feishin (and also if there are VPNs/alternative routing)

hedgehog1029:
Linter errors resolved :saluting_face: 

I think it not working for local machine or weird routing setups is normal and in-line with how official Jellyfin clients would expect to work. Discovery works by broadcasting a UDP packet to `255.255.255.255:7359`, and packets with that destination address are never routed, only switched. So you have to be on the same L2 broadcast domain.

Also, all-high-bits broadcast is kind of weird and if you have multiple suitable interfaces it might just pick a random interface each time to send the packet out of. We could do slightly better if we scanned all the interfaces and used the subnet broadcast address to send a packet out of each interface, but that sounds complicated and platform-specific so I didn't bother.

hedgehog1029:
Per some discussion in the discord, fixed localhost discovery by sending another discovery packet to the loopback broadcast address!

kgarner7 (approved):

PR #1154, opened 2025-09-28
Remove unnecessary sections from Genius lyrics

angelolz:
When I was fetching lyrics from genius, I was noticing that the title displayed had the artist(s) names duplicated twice. I've added some adjustments to the artists and titles returned from the api response to try and remove that duplication, removing the artist name from the `full_title` field. 

I've also noticed that some unwanted text such as "143 Contributors" and "Translations" are appearing in the lyrics, those have been removed too.

Here's some screenshots to showcase the change:
**Before**
image

**After**
image

### DIFF
--- a/src/main/features/core/lyrics/genius.ts
+++ b/src/main/features/core/lyrics/genius.ts
@@ -204,16 +204,16 @@ async function getSongId(
         return null;
     }
 
-    const artist_normalized = hit.artist_names.replace(/\u00A0/g, ' ').trim();
-    const full_title_normalized = hit.full_title
+    const artistNormalized = hit.artist_names.replace(/\u00A0/g, ' ').trim();
+    const fullTitleNormalized = hit.full_title
         .replace(/\u00A0/g, ' ')
-        .replace(`by ${artist_normalized}`, '')
+        .replace(`by ${artistNormalized}`, '')
         .trim();
 
     return {
-        artist: artist_normalized,
+        artist: artistNormalized,
         id: hit.url,
-        name: full_title_normalized,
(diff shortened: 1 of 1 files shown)


src/main/features/core/lyrics/genius.ts
```diff
@@ -201,10 +204,16 @@ async function getSongId(
         return null;
     }
 
+    const artistNormalized = hit.artist_names.replace(/\u00A0/g, ' ').trim();
+    const fullTitleNormalized = hit.full_title
+        .replace(/\u00A0/g, ' ')
+        .replace(`by ${artistNormalized}`, '')
+        .trim();
+
```
kgarner7: Do you have a specific example where this is relevant? I'm not seeing examples of NBSP. I'm also not entirely keen on removing `by {artist}` if that's in the title

kgarner7 (changes requested):

src/main/features/core/lyrics/genius.ts
angelolz: I couldn't properly filter out the artist out of the title because of the NBSP being in the `hit.full_title` as shown below:
image

This is irrelevant now since I took a second look at this and realized that the response has a regular `title` value, which is being used instead.

Please let me know if there's a certain field you'd prefer to be used here, thanks! Sorry for the delayed response, been caught up with a lot of projects.

angelolz: Using `hit.artist_names` and `hit.title` for the artist and title respectively presents itself like this:
image
image

PR #1161, opened 2025-09-30
Option to show Feishin icon instead of play icon in discord rpc when "Show rich presence when paused" is disabled

poopoofartnut:
Adds a new setting that lets users show the Feishin app icon instead of the play icon when "Show rich presence when paused" is turned off.


jeffvli:
I don't understand the intent here with your description. 

The `Show rich presence when paused` setting when disabled should *not* show any presence when paused anyways.

PR #1175, opened 2025-10-07
Add PWA to web app

luis-c465:
This PR adds PWA functionality for the web version of feishin.

The PWA functionality allows the app to be downloaded and added to the desktop as a normal desktop app, creating its own window when opened. Though it will use the browser that the user downloaded the PWA from to run the web app. 

image
image

# Why not just use the normal desktop app

In my case there are a couple or reasons:

1. When using the Linux desktop app on my laptop running Fedora 42 audio playback on the desktop app with both available playback methods consumes much more power and CPU than expected. Often pegging a thread on my computer, draining battery life. Chances are this is an issue on my part, though since the web version exists I would rather use that version. 
2. Since the desktop app is using electron to run a Chromium browser for the app to work. Effectively running the app in a browser with additional features. To save on both memory and CPU I would rather run the app in a browser I already have opened. Even if I lose some features because of it.
3. With the PWA sharing resources from the browser it also shares extensions. Notably there is an [extension](https://chromewebstore.google.com/detail/autopause/bcecldolamfbkgokgpnlpmhjcijglhll?hl=en) which can automatically pause the music when another sound in the browser is playing. This is quite convenient, and would otherwise be much more difficult to do with the music being played from a separately installed desktop app.

# The changes

To add the PWA a **development dependency** is added [vite-plugin-pwa](https://github.com/vite-pwa/vite-plugin-pwa).
This plugin handles creating both the manifest file for the PWA, and the service worker which allows the servers assets (HTML, CSS, JS) to be cached on the device.

When the web version is built, the plugin will create both the manifest file and the service worker file and place them in the assets' directory.

As a final change, since the PWA manifest needs both icons and a screenshot preview of the app, some icons and media screenshots are included in the final build.

# Notes

- The PWA will not work nor show up when the web version is being run in development mode. Unfortunately the production web handles showing public assets slightly differently than the development version due to a lack of the public folder.
- Finally, thank you for making this awesome music player, been using daily for a year lol. Easily the best self-hostable one I have found so far.

### DIFF
--- a/ng.conf.template
+++ b/ng.conf.template
@@ -14,7 +14,6 @@ server {
 
   location ${PUBLIC_PATH} {
     alias /usr/share/nginx/html/;
-    try_files $uri $uri/ /index.html =404;
   }
 
   location ${PUBLIC_PATH}settings.js {
--- a/ng.conf.template
+++ b/ng.conf.template
@@ -14,6 +14,7 @@ server {
 
   location ${PUBLIC_PATH} {
     alias /usr/share/nginx/html/;
+    try_files $uri $uri/ /index.html =404;
   }
 
   location ${PUBLIC_PATH}settings.js {
--- a/package.json
+++ b/package.json
@@ -167,7 +167,8 @@
         "vite": "^6.3.6",
         "vite-plugin-conditional-import": "^0.1.7",
         "vite-plugin-dynamic-import": "^1.6.0",
-        "vite-plugin-ejs": "^1.7.0"
+        "vite-plugin-ejs": "^1.7.0",
+        "vite-plugin-pwa": "^1.0.3"
     },
     "pnpm": {
         "onlyBuiltDependencies": [
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -266,7 +266,7 @@ importers:
         version: 8.18.1
       '@vitejs/plugin-react':
         specifier: ^4.3.4
-        version: 4.5.0(vite@6.3.6(@types/node@22.15.32)(sass-embedded@1.89.0)(sugarss@4.0.1(postcss@8.5.3))(terser@5.39.2))
+        version: 4.5.0(vite@6.3.6(@types/node@22.15.32)(sass-embedded@1.89.0)(sugarss@4.0.1(postcss@8.5.6))(terser@5.39.2))
       concurrently:
         specifier: ^7.1.0
         version: 7.6.0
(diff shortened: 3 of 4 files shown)


kgarner7 (changes requested):
1. The service worker is served at `/sw.js`, which is not correct in the nginx/docker configuration (`assets/sw.js`)
2. Please fix your lint errors.

kgarner7 (changes requested):
Thanks for the update. The problem appears to be the way that `index.html` is generated, where it still tries to load from `./sw.js`
```html
<script id="vite-plugin-pwa:inline-sw">if('serviceWorker' in navigator) {window.addEventListener('load', () => {navigator.serviceWorker.register('./sw.js', { scope: '/' })})}</script>
```
It looks like you should change https://vite.dev/config/shared-options.html#base (`./assets`).
To test this, I recommend using the docker build and making sure it works
```bash
docker build -t feishin .
docker run -d --name feishin  -p 9180:9180 feishin
```
(You may need to do this over https, as service worker registration is very picky)

luis-c465:
TLDR, I tried for a long time to get the VitePWA plugin working while the sw.js and manifest are in the assets directory though could not get it to work. Instead, make a one line change to the nginx template config to allow files to be served from the root.

---

I had initially tried to get the PWA working while putting the manifest and the sw in the assets directory though that lead to 2 issues:

1. Then the service worker is under `assets/sw.js` which means at maximum it can only have a scope of `assets` meaning it can no longer cache the `index.html` resulting in the PWA not working if not connected to the internet. (This could only be fixed if the nginx config is changed to return a specific service worker allowed header)
2. The vite PWA plugin seems to be largely built around the assumption that the manifest and sw.js will be placed at the root of the project. If not the configs for both have to be changed, and during tested led to a large amount of issues of the paths being different between the two. Even after finding a configuration that worked the workerbox script seems to be hardcoded to load from the root of the project, resulting in some files not being cached as expected. 

While the 2 issues above could be fixed by finicking with configuration till it works, it may be better to make an exception for both of those files and allow them to live at the root of the site.

To allow that I removed a remove the `try_files $uri $uri/ /index.html =404;` at the `/` location in the nginx template file. Removing it allows the manifest and sw files to be correctly served at the root.

Testing this with docker the app is working as expected and no errors are shown in the console, even when the network is disabled in the developer console. 
image

kgarner7:
Ah, I actually apologize greatly; your prior fix did work (I hadn't pulled it). That has a minimal set of changes, so it's preferable.

luis-c465:
@kgarner7 I reverted the commit to put the sw and manifest at the root.

PR #1179, opened 2025-10-10
show time remaining instead of duration on click

Lyall-A:
this replaces the duration with the time remaining of the current song by clicking on the text, similar to spotify
image

i'm not sure if the boolean should be stored in app.store or somewhere else, let me know if needs changing


src/renderer/features/player/components/center-controls.tsx
```diff
@@ -293,9 +296,10 @@ export const CenterControls = ({ playersRef }: CenterControlsProps) => {
                         fw={600}
                         isMuted
                         isNoSelect
+                        onClick={() => setShowTimeRemaining(!showTimeRemaining)}
```
jeffvli, line 300: There's no indication that the text is clickable.

I would add a button role to the component and change the cursor when it is hovered.

Lyall-A: Added button role and set cursor to pointer when hovered, also disabled user select (is `isNoSelect` supposed to do that?)

PR #1181, opened 2025-10-11
Album page design refactor

ilfarpro:
This PR introduces changes to album page. It aimed to use vertical space better by clever elements placing. Each commit in this PR represents a small change, so we can easily revert any change in case we don't like it. 

What changed:
- Album title is not enormous anymore
- Like and star rating buttons are placed near each other because they represent same feature
- Album metadata rethought and combined with main artist name 

Other small changes visible on screenshots.

**Before**
before

**After**
after

### DIFF
--- a/src/renderer/features/albums/components/album-detail-content.module.css
+++ b/src/renderer/features/albums/components/album-detail-content.module.css
@@ -7,6 +7,6 @@
     display: flex;
     flex-direction: column;
     gap: var(--theme-spacing-lg);
-    padding: 1rem 2rem 5rem;
+    padding-bottom: 5rem;
     overflow: hidden;
 }
--- a/src/renderer/features/albums/components/album-detail-content.tsx
+++ b/src/renderer/features/albums/components/album-detail-content.tsx
@@ -29,7 +29,12 @@ import {
     SONG_CONTEXT_MENU_ITEMS,
 } from '/@/renderer/features/context-menu/context-menu-items';
 import { usePlayQueueAdd } from '/@/renderer/features/player';
-import { PlayButton, useCreateFavorite, useDeleteFavorite } from '/@/renderer/features/shared';
+import {
+    PlayButton,
+    useCreateFavorite,
+    useDeleteFavorite,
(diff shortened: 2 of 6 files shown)


pyxfluff:
I did actually make a layout like this for a fork I work on:

image

I do think that some of the changes here need to happen though. Removing the fade down is a negative in my opinion but removing "ALBUM" and condensing the information is good.

ilfarpro:
@pyxfluff,

> Removing the fade down is a negative in my opinion

Not sure If I follow, do you mean gradient from start to end of album header or gradient that like spills over to next section? 
Or do you mean blurred album background?

When I moved things around tried to be gentle with existing code, so It's not removed, there is option to toggle it, modify blur amount. It was introduced recently, if I am not mistaken in 452803fc722b1f5db0e0773fa3d4ec339d3fe0e9

I see there you have 2 release dates, I assume one of it is date when it was added to library and other - when album was released. I'd love to add official release date in my implementation, but not sure how to query it, so I have only year. Does it have to be baked in song metadata?

pyxfluff:
> do you mean gradient from start to end of album header or gradient that like spills over to next section?

Yes, I meant that little fade part down into the list. I think it looks good. I tried removing the line because I thought it looked bad but got told to add it back for upstream even though it isn't there in my custom layout.

image



> I see there you have 2 release dates, I assume one of it is date when it was added to library and other - when album was released

Not actually, it's just one of them always displays january first. I assume the thing to check it is just broken, I just haven't bothered to fix it. All of the release dates in my library are messed up because Navidrome used to just purge metadata every time my smb share got unmounted (thanks microsoft) so none of my addition dates are right mostly.

jeffvli:
I appreciate the PRs, but I will have to sit on these (and your other PRs) for a while because I'm currently reworking large parts of the app which your changes may clobber.

I'm also a bit conflicted on merging strictly design changes because design is *subjective* and it may not align with what I envision for the pages to be. I will let you know when I have a clearer picture on how the app is changing.

PR #1183, opened 2025-10-11
Artist page design refactor

ilfarpro:
This PR introduces changes to artist page. It aimed to use vertical space better by clever elements placing and grouping. Each commit in this PR represents a small change, so we can easily revert any change in case we don't like it.

What changed:
- SImilar buttons layout like in PR #1181
- Removed duplicated buttons for discography and all songs

**Before**
artist_before

**After**
artist_after


jeffvli:
Pages were redesigned, now outdated

PR #1186, opened 2025-10-13
Add install script for AppImage

mihawk90:
This essentially supersedes #1020 and addresses my concerns raised there. Basically, it's suggested the provided `.desktop` file will just work, but that's not the case.

This PR addresses this:

* adjusts to CPU architecture by reading `uname -m`
* `wayland-native` option for Wayland flags
* `remove` option to delete downloaded files
* generates Desktop entry from template
* downloads various icon sizes to correct locations
  * uses reverse DNS notation so icon can be overridden by icon themes (can technically be done with any filename, but providing a filename from the project side makes it "official" for theme authors)
* add `GenericName`, `Comment`, and `Keywords` to Desktop file
  * The `Comment` is what shows up here:
    image
  * The `GenericName` is an alternate name some DE's application launchers can show when enabled (IDK who uses this, but y'know, it's part of the spec):
    image
  * `Keywords` per the spec:
    > A list of strings which may be used in addition to other metadata to describe this entry. This can be useful e.g. to facilitate searching through entries. The values are not meant for display, and should not be redundant with the values of `Name` or `GenericName`.
    image
* update accompanying README instructions
* remove previous example `.desktop` file

Final `.desktop` file also validates:
```sh
❯ desktop-file-validate org.jeffvli.feishin.desktop
❯ echo $?
0
```

---

One small thing:
The `Comment` string can be localised:
```ini
Comment=A player for your self-hosted music server
Comment[de_DE]=<German TL here>
Comment[fr_FR]=<French TL here>
…
```
But I don't know if/how that can be integrated to Weblate.

---

For testing, you need to use a link from my branch:
```sh
curl 'https://raw.githubusercontent.com/mihawk90/feishin/refs/heads/install-appimage/install-feishin-appimage' | sh
```

### DIFF
--- a/install-feishin-appimage
+++ b/install-feishin-appimage
@@ -1,0 +1,74 @@
+#!/bin/sh
+
+dir=$1
+arg=$2
+arch=$(uname -m)
+
+if [ "$arch" != "x86_64" ] && [ "$arch" != "aarch64" ]; then
+	echo "CPU architecture not recognised (not x86_64 or aarch64). Aborting."
+	exit 1
+fi
+
+# workaround if we're not renaming the artifact
+if [ "$arch" = "aarch64" ]; then
+	arch='arm64'
+fi
+
+if [ ! -d "${dir}" ]; then
+	echo "${dir}} is not a directory or does not exist. Please provide an existing directory."
+	exit 1
+fi
+
+localShare="$HOME/.local/share"
+localShareIcons="${localShare}/icons/hicolor"
+
+if [ "${arg}" = "remove" ]; then
+	rm -v \
+		"${localShareIcons}"/512x512/apps/org.jeffvli.feishin.png \
+		"${localShareIcons}"/256x256/apps/org.jeffvli.feishin.png \
+		"${localShareIcons}"/128x128/apps/org.jeffvli.feishin.png \
+		"${localShareIcons}"/64x64/apps/org.jeffvli.feishin.png \
+		"${localShareIcons}"/32x32/apps/org.jeffvli.feishin.png \
+		"${localShare}"/applications/org.jeffvli.feishin.desktop \
+		"${dir}./Feishin-linux-${arch}.AppImage"
+	exit 0
+fi
+
+curl -L --create-dirs --write-out '%{filename_effective}\n' \
+	-o "${dir}/Feishin-linux-${arch}.AppImage" "https://github.com/jeffvli/feishin/releases/latest/download/Feishin-linux-${arch}.AppImage" \
+	-o "${localShareIcons}"/512x512/apps/org.jeffvli.feishin.png 'https://github.com/jeffvli/feishin/blob/development/assets/icons/512x512.png?raw=true' \
+	-o "${localShareIcons}"/256x256/apps/org.jeffvli.feishin.png 'https://github.com/jeffvli/feishin/blob/development/assets/icons/256x256.png?raw=true' \
+	-o "${localShareIcons}"/128x128/apps/org.jeffvli.feishin.png 'https://github.com/jeffvli/feishin/blob/development/assets/icons/128x128.png?raw=true' \
+	-o "${localShareIcons}"/64x64/apps/org.jeffvli.feishin.png 'https://github.com/jeffvli/feishin/blob/development/assets/icons/64x64.png?raw=true' \
+	-o "${localShareIcons}"/32x32/apps/org.jeffvli.feishin.png 'https://github.com/jeffvli/feishin/blob/development/assets/icons/32x32.png?raw=true'
+chmod -v u+x "${dir}/Feishin-linux-${arch}.AppImage"
+
+waylandFlags=''
+if [ "${arg}" = "wayland-native" ]; then
+	waylandFlags=' --enable-features=UseOzonePlatform,WaylandWindowDecorations --ozone-platform-hint=auto'
+fi
+
+# this is for Debian-based kernels and ALT respectively
+# https://unix.stackexchange.com/a/303214/145722
+sandboxFlag=''
+if [ "$(sysctl kernel.unprivileged_userns_clone 2>/dev/null)" = "0" ] \
+	|| [ "$(sysctl kernel.userns_restrict 2>/dev/null)" = "1" ]; then
+	sandboxFlag=' --no-sandbox'
+fi
+
+mkdir -pv "${localShare}"/applications
+cat > "${localShare}"/applications/org.jeffvli.feishin.desktop << EOL
+[Desktop Entry]
+Name=Feishin
+GenericName=Music player
+Exec=${dir}/Feishin-linux-${arch}.AppImage${sandboxFlag}${waylandFlags}
+TryExec=${dir}/Feishin-linux-${arch}.AppImage
+Terminal=false
+Type=Application
+Icon=org.jeffvli.feishin
+StartupWMClass=feishin
+SingleMainWindow=true
+Categories=AudioVideo;Audio;Player;Music;
+Keywords=Navidrome;Jellyfin;Subsonic;OpenSubsonic
+Comment=A player for your self-hosted music server
+EOL
--- a/README.md
+++ b/README.md
@@ -59,15 +59,27 @@ For media keys to work, you will be prompted to allow Feishin to be a Trusted Ac
 
 #### Linux Notes
 
-If you're using a Linux device, a `.desktop` file is recommended for easy launching of Feishin.
+We provide a small install script to download the latest `.AppImage`, make it executable, and also download the icons required by Desktop Environments. Finally, it generates a `.desktop` file to add Feishin to your Application Launcher.
 
-Download the [latest release (AppImage)](https://github.com/jeffvli/feishin/releases) and [application icon](https://github.com/jeffvli/feishin/blob/development/resources/icon.png?raw=true) to your `~/applications/` folder. This folder may need to be created if it does not already exist.
+Simply run the installer like this:
+```sh
+dir=/your/application/directory
+curl 'https://raw.githubusercontent.com/jeffvli/feishin/refs/heads/development/install-feishin-appimage' | sh -s -- "$dir"
+```
 
-Rename the icon to `Feishin-linux-x86_64.png`.
+The script also has an option to add launch arguments to run Feishin in native Wayland mode. Note that this is experimental in Electron and therefore not officially supported. If you want to use it, run this instead:
+```sh
+dir=/your/application/directory
+curl 'https://raw.githubusercontent.com/jeffvli/feishin/refs/heads/development/install-feishin-appimage' | sh -s -- "$dir" wayland-native
+```
 
-Save the [example desktop file](https://raw.githubusercontent.com/jeffvli/feishin/refs/heads/development/feishin.desktop) as `~/.local/share/applications/feishin.desktop`.
+It also provides a simple uninstall routine, removing the downloaded files:
+```sh
+dir=/your/application/directory
+curl 'https://raw.githubusercontent.com/jeffvli/feishin/refs/heads/development/install-feishin-appimage' | sh -s -- "$dir" remove
+```
 
-You will now see Feishin show up in your menu. The properties in the example desktop file may need to be modified to match your system.
+The entry should show up in your Application Launcher immediately. If it does not, simply log out, wait 10 seconds, and log back in. Your Desktop Environment may alternatively provide a way to reload entries.
 
 ### Web and Docker
 
--- a/feishin.desktop
+++ b/feishin.desktop
@@ -1,9 +1,0 @@
-[Desktop Entry]
-Name=Feishin
-Comment=An Electron-based music streaming app
-Exec=/home/username/.applications/Feishin-linux-x86_64.AppImage
-Icon=/home/username/.applications/Feishin-linux-x86_64.png
-Terminal=false
-Type=Application
-Categories=AudioVideo;Audio;Music;Player;
-StartupNotify=true


install-feishin-appimage
```diff
@@ -0,0 +1,50 @@
+#!/bin/sh
+
+arch=$(uname -m)
+# workaround if we're not renaming the artifact
+if [ "$arch" == "aarch64" ]; then
+	arch='arm64'
+fi
+
```
kgarner7, line 16: Of note, the only releases are `x86_64` and `aarch64` (`arm64`). It might be worthwhile warning if neither of these is found, as this is not supported.

mihawk90: I was thinking about that too, but I wasn't sure what it returned otherwise... maybe just check with 2 string comparisons then?

kgarner7: That's probably sufficient? I don't think it's worth fetching supported artifact types from GitHub, and if for whatever reason new artifacts are added, this could be updated.

mihawk90: I meant what `uname -m` returns :P but anyway, I'll add the abort

mihawk90:
Just pushed some smaller changes:

* moved some blocks around to logically group them together
* used `org.jeffvli.feishin` as the identifier to align with the `appId` in `electron-builder.yml`
  * this functionally doesn't actually do anything on Linux, but it doesn't hurt to have them the same across platforms either. On Windows and Mac they are used for OS internals, on Linux just for icon themes and AppStream identifiers (which we don't currently use anyway because Feishin isn't in repos)
* added `SingleMainWindow=true`
  * This just hints to DEs to not give you "Open New Window" options, which don't work anyway since Electron blocks it. It works on KDE, may or may not work on others.
* lowercased `StartupWMClass`
  * I'm not sure this is actually required (I'm not getting 2 icons on KDE at least), but it doesn't hurt to have and I just lowercased it because it's what the window class actually is and I don't know if it's case-sensitive

Also a note:
The old example desktop file had `StartupNotify`, but after reading the spec on it 5 times I still have no idea what it does, and the spec it links to is a 404 so... anyway, spec also says:
> If absent, a reasonable handling is up to implementations

So I just left it out and it didn't make a difference for me.

install-feishin-appimage
```diff
@@ -0,0 +1,57 @@
...
+arch=$(uname -m)
+
+if [ "$arch" != "x86_64" ] && [ "$arch" != "aarch64" ]; then
+	echo "CPU architecture not recognised (not x86_64 or aarch64). Aborting."
+	exit 1
+fi
+
+# workaround if we're not renaming the artifact
+if [ "$arch" == "aarch64" ]; then
```
kgarner7: `==` is a bash extension, not actually valid for `sh`/plain POSIX

install-feishin-appimage
```diff
@@ -0,0 +1,57 @@
...
+	arch='arm64'
+fi
+
+localShare="$HOME/.local/share"
+localShareIcons="${localShare}/icons/hicolor"
+
+if [ "$1" == "remove" ]; then
```
kgarner7: Likewise here

install-feishin-appimage
```diff
@@ -0,0 +1,57 @@
...
+	rm -v \
+		"${localShareIcons}"/{512x512,256x256,128x128,64x64,32x32}/apps/org.jeffvli.feishin.png \
+		"${localShare}"/applications/org.jeffvli.feishin.desktop \
+		./Feishin-linux-${arch}.AppImage
+	exit 0
+fi
+
+mkdir -pv "${localShareIcons}"/{512x512,256x256,128x128,64x64,32x32}/apps "${localShare}"/applications
```
kgarner7: This is also not a stock posix (at least, not debian `sh` posix).

install-feishin-appimage
```diff
@@ -0,0 +1,57 @@
...
+	-o "${localShareIcons}"/256x256/apps/org.jeffvli.feishin.png 'https://github.com/jeffvli/feishin/blob/development/assets/icons/256x256.png?raw=true' \
+	-o "${localShareIcons}"/128x128/apps/org.jeffvli.feishin.png 'https://github.com/jeffvli/feishin/blob/development/assets/icons/128x128.png?raw=true' \
+	-o "${localShareIcons}"/64x64/apps/org.jeffvli.feishin.png 'https://github.com/jeffvli/feishin/blob/development/assets/icons/64x64.png?raw=true' \
+	-o "${localShareIcons}"/32x32/apps/org.jeffvli.feishin.png 'https://github.com/jeffvli/feishin/blob/development/assets/icons/32x32.png?raw=true'
+chmod -v u+x ./Feishin-linux-${arch}.AppImage
+
+waylandFlags=''
+if [ "$1" == "wayland-native" ]; then
+	waylandFlags='--enable-features=UseOzonePlatform,WaylandWindowDecorations --ozone-platform-hint=auto'
+fi
```
kgarner7, line 57: It feels weird to just have positional argumetns without any help. 
A portable option would be to use `getopts`, although that's only single character options.

install-feishin-appimage
```diff
@@ -0,0 +1,57 @@
...
+
+mkdir -pv "${localShareIcons}"/{512x512,256x256,128x128,64x64,32x32}/apps "${localShare}"/applications
+curl -L --write-out '%{filename_effective}\n' \
+	-O "https://github.com/jeffvli/feishin/releases/latest/download/Feishin-linux-${arch}.AppImage" \
+	-o "${localShareIcons}"/512x512/apps/org.jeffvli.feishin.png 'https://github.com/jeffvli/feishin/blob/development/assets/icons/512x512.png?raw=true' \
+	-o "${localShareIcons}"/256x256/apps/org.jeffvli.feishin.png 'https://github.com/jeffvli/feishin/blob/development/assets/icons/256x256.png?raw=true' \
+	-o "${localShareIcons}"/128x128/apps/org.jeffvli.feishin.png 'https://github.com/jeffvli/feishin/blob/development/assets/icons/128x128.png?raw=true' \
+	-o "${localShareIcons}"/64x64/apps/org.jeffvli.feishin.png 'https://github.com/jeffvli/feishin/blob/development/assets/icons/64x64.png?raw=true' \
+	-o "${localShareIcons}"/32x32/apps/org.jeffvli.feishin.png 'https://github.com/jeffvli/feishin/blob/development/assets/icons/32x32.png?raw=true'
+chmod -v u+x ./Feishin-linux-${arch}.AppImage
```
kgarner7: Feels a bit weird to just download the binary in the current directory? https://specifications.freedesktop.org/basedir-spec/latest/#:~:text=%20user-specific%20executable%20files%20may%20be%20stored%20in%20%24home%2F.local%2Fbin.%20distributions%20should%20ensure%20this%20directory%20shows%20up%20in%20the%20unix%20%24path%20environment%20variable%2C%20at%20an%20appropriate%20place.%20 suggests using `$HOME/.local/bin`, which seems more appropriate.
Alternatively, asking for a custom path?

install-feishin-appimage
```diff
@@ -0,0 +1,57 @@
...
+
+waylandFlags=''
+if [ "$1" == "wayland-native" ]; then
+	waylandFlags='--enable-features=UseOzonePlatform,WaylandWindowDecorations --ozone-platform-hint=auto'
```
kgarner7: Also not POSIX equal.

kgarner7 (changes requested):
Thanks for the additional work. Feedback below.
Most of the POSIX complaints were found testing on debian 12.11 (aarch64). The installer actually didn't succeed as a result of some of the issues noted.

install-feishin-appimage
mihawk90: That would be odd... `bash` is supposed to go into POSIX compatibility mode when called as `sh`:

```sh
$ man sh
…
        If  bash  is  invoked  with  the name sh, it tries to mimic the startup behavior of historical versions of sh as
       closely as possible, while conforming to the POSIX standard as well.
```

And that was working just fine while testing, both with `curl/cat .. | sh`, as well as `sh install-feishin-appimage`, both with and without shebang.

So IDK... if I can't test this because `bash` refuses to work as the man page suggests, that's gonna be annoying.

It also calls into question how useful sticking to POSIX really is, so I could change the shebang `#!/bin/bash` too.

install-feishin-appimage
mihawk90: I could ask for a path using `read`, but I honestly don't see much value in it. When using `read` AFAIK tab completion doesn't work, so `cd` to the correct directory beforehand would be faster. That's also what I put into the README.

install-feishin-appimage
mihawk90: Was thinking about that but I opted to not do it to keep it simple.

It's not a local script so it is only run after reading the README, otherwise users won't even know of its existence. Both arguments are mentioned there.

In addition, when `curl | sh`ing it there is also no way to show the helptext without either a) not running the actual install, which defeats the purpose of the script, or b) running the install *and* show the help, which defeats the purpose of the help.

install-feishin-appimage
mihawk90: Fixed, hope it works :shrug:

install-feishin-appimage
mihawk90: removed the `mkdir` and used `curl`'s `--create-dirs` instead

install-feishin-appimage
mihawk90: Fixed, hope it works :shrug:

install-feishin-appimage
mihawk90: Fixed, hope it works :shrug:

install-feishin-appimage
mihawk90: Coming back to this because I was thinking about it...
> suggests using `$HOME/.local/bin`, which seems more appropriate.

As I read this, this is meant for things you'd want in your `$PATH` so you can call them from a terminal, but I don't think that makes a lot of sense in case of Feishin. I don't know why that would be required.

Besides, I wouldn't want to force a specific directory on users, I certainly don't want my AppImages there as this for me is reserved for binaries I need on the CLI and my AppImages live elsewhere. If someone really wants it in there, they are free to `cd` to it beforehand and put it there.

kgarner7: Putting it in `pwd` is still a very strange choice. I can accept having the path as a _parameter_, but putting it in a more sane location is better.

install-feishin-appimage
```diff
@@ -0,0 +1,60 @@
...
+if [ "$1" = "wayland-native" ]; then
+	waylandFlags='--enable-features=UseOzonePlatform,WaylandWindowDecorations --ozone-platform-hint=auto'
+fi
+
+mkdir -pv "${localShare}"/applications
+cat > "${localShare}"/applications/org.jeffvli.feishin.desktop << EOL
+[Desktop Entry]
+Name=Feishin
+GenericName=Music player
+Exec=$(pwd)/Feishin-linux-${arch}.AppImage --no-sandbox $waylandFlags
```
kgarner7: Why is `--no-sandbox` being passed in? This should not be required by default.

mihawk90: This is what I mentioned [on discord](https://discord.com/channels/922656312888811530/922656312888811536/1429761168620392458) too:

> So I've been looking into something that bothered me for a while... In the AppImage install-script PR you might have noticed I am passing `--no-sandbox` to the `Exec` line:
> <https://github.com/jeffvli/feishin/pull/1186/files#diff-924e48df179523f9487f572de509b6c5d7cded0ccd2bd587f571bdec79f7a3adR49>
> 
> This stems from me deriving that template from the `desktop` file that ships in the AppImage (but to my knowledge isn't actually used by the AppImage):
> ```ini
> [Desktop Entry]
> Name=Feishin
> Exec=AppRun --no-sandbox %U
> ```
> However, I can't find why it has this in the first place. The desktop file is generated by electron-builder, but I don't know why they would enable this by default since I can't find this being enabled in the electron-builder.yml:
> <https://github.com/jeffvli/feishin/blob/848b3f58b3354003c2329c4adbb8e3a10d4042c3/electron-builder.yml#L44-L50>
> Or is this derived from this?
> <https://github.com/jeffvli/feishin/blob/848b3f58b3354003c2329c4adbb8e3a10d4042c3/src/main/index.ts#L289>
> 
> Feishin isn't the only electron app that has this though which leads me to believe this is an electron-builder default because.... reasons?
> 
> The AppImage docs actually have a page on this:
> <https://docs.appimage.org/user-guide/troubleshooting/electron-sandboxing.html>
> This mentions a kernel option that this is supposed to be used for, but after some digging it turns out this only affects Debian-derived kernels since it's a Debian patch?
> <https://serverfault.com/a/939457/558996>
> 
> The Readme also mentions this kernel option
> <https://github.com/jeffvli/feishin?tab=readme-ov-file#i-have-the-issue-the-suid-sandbox-helper-binary-was-found-but-is-not-configured-correctly-on-linux>
> But that whole paragraph makes no sense to me because you can't change permissions in the AppImage so I'm not sure how either of the two lines would work.
> 
> **But long story short, should I drop that option or leave it? I feel like disabling sandboxing just because seems weird**

---

> some digging later...
> <https://github.com/electron-userland/electron-builder/pull/4496>
> 
> Even the PR author acknowledges this isn't actually a good idea

---

> They changed it here though:
> <https://github.com/electron-userland/electron-builder/pull/6429>
> 
> This is only added if you don't add your own `executableArgs` now... 
> But apparently not even Debian requires this anymore
> <https://github.com/electron-userland/electron-builder/issues/5371#issuecomment-1218361862>
> 
---

> But also like... as I said as far as I know the desktop file is never looked at for the actual execution anyway so why they add the argument specifically for the appimage target is beyond me

---

On that note I've also asked in the appimage IRC 3 days ago if/how this is used at all and I've yet to get a reply so... yeah, there's that.

kgarner7: It worked fine in stock debian (arm 12). You can add disabling sandbox as an _option_, but it should **not** be the default.

mihawk90: Agreed, which is why I posed the question on Discord :P

kgarner7: Also of note, `--no-sandbox` is different from the `sandbox` option. The [`--sandbox` option](https://www.electronjs.org/docs/latest/tutorial/sandbox#configuring-the-sandbox) disables the `renderer` sandbox, whereas `--no-sandbox` disables the sandbox for _all_ processes (including other processes).

`--no-sandbox` is necessary primarily for systems where unpriviliged usernamespaces (where `sysctl kernel.unprivileged_userns_clone` is 1), or Ubuntu with whatever apparmor nonsense it has.

I suppose, you could try to detect this variable and then if it's zero, then add no sandbox, per https://docs.appimage.org/user-guide/troubleshooting/electron-sandboxing.html

mihawk90: Problem with detection is that I don't have a system to test this on where it would return true.
When I run the command in the AppImage docs (on Fedora 42), I get this:
```sh
$ sysctl kernel.unprivileged_userns_clone
sysctl: cannot stat /proc/sys/kernel/unprivileged_userns_clone: No such file or directory
$ echo $?
1
```
So I can test when it's *not* required I guess :'D

kgarner7: Ugh, that's really annoying. I also see that Fedora doesn't have this (or any other option as far as I can tell). It looks like Debian-based systems (and Arch-based, for some reason), have this flag, and alt linux (per https://unix.stackexchange.com/questions/303213/how-to-enable-user-namespaces-in-the-kernel-for-unprivileged-unshare) has `/proc/sys/kernel/userns_restrict`.

mihawk90: I guess I can check both `kernel.unprivileged_userns_clone == 0` and `kernel.userns_restrict == 1`.

The question is whether we really want to handle every possible permutation of this option right now, or if we just let it come to us down the line when it's actually needed.

kgarner7: I think a check that's good _enough_ is probably fine. The most foolproof way would of course to just be to try to run the executable, and if it doesn't crash that's probably good enough, but that's not _great_.

mihawk90:
Updated:

* added argument for directory to install to
* removed `--no-sandbox` by default
  * adds a check to add it when required, should mostly™ work
* reference arguments by variable name, not number

install-feishin-appimage
mihawk90: Updated

install-feishin-appimage
mihawk90: Updated with an argument. If users want it in local/bin they can just provide the path to it.

kgarner7 (approved):

PR #1192, opened 2025-10-13
Fix Lyric Translation Not Resetting

ENDlezZenith:
Currently, lyric translation would not reset and would directly pass the translation of the last song to the next.

This PR does the following:

- Clear the lyric translation cache and disable it when changing tracks
- Added a toggle (users who would not like translation to be enabled automatically could still use the translation button that is only temporary for the current track playing) for auto translation, so users no longer need to click translation every time a new track is playing (same behavior as fetch lyric from internet source)
- Trivial lyric translation logic fix and reuse code for fetching lyric translation api responses


ENDlezZenith:
Ops

PR #1195, opened 2025-10-15
Update README photos

pyxfluff:
The photos in the README were from early 2023 and the design is significantly different now so I updated the photos. I tried to stay away from unknown artists and changed up the playing songs in the screenshots.


jeffvli:
Might be nitpicky, but can you use a smaller and more standard resolution for the screenshots? For example, the originals all use 1440 x 900.

pyxfluff:
> Might be nitpicky, but can you use a smaller and more standard resolution for the screenshots? For example, the originals all use 1440 x 900.

corrected them, should be the same size now

jeffvli:
App rework / redesign, now outdated

PR #1212, opened 2025-10-22
Prevent Media Session Handling on MPV

ENDlezZenith:
MPV have its own integration of SMTC, and hence currently behavior of feishin is that the media session option would do nothing additive but disables global hotkeys for MPV. 

In according to:

> This was discussed in the Discord (/matrix bridge). The _correct_ mechanism would be to just use MediaSession API for Electron (when not using MPV player).

This PR does the following:

- Hides the media session option in MPV
- Shows global hotkeys option in MPV
- Redo registering global hotkeys when enabled like before in MPV


src/renderer/features/settings/components/hotkeys/window-hotkey-settings.tsx
```diff
@@ -44,7 +50,9 @@ export const WindowHotkeySettings = () => {
                 context: 'description',
                 postProcess: 'sentenceCase',
             }),
-            isHidden: !isElectron() || (enableWindowsMediaSession && isWindows),
+            isHidden:
+                !isElectron() ||
+                (enableWindowsMediaSession && isWindows && playbackType === PlaybackType.WEB),
```
jeffvli: Personally not in favor of hiding it.

Users who are already using MPV may never know that this option is available if it is hidden. Disabling is enough.

ENDlezZenith: Then should we change the globalMediaHotkeys too, only disable but not hidden for Web users?

mihawk90: Maybe I'm slow but why would MPV users need to see/know about this option when they are already getting the same functionality anyway? They're not missing out on anything not seeing it 🤔
Or am I misunderstanding the PR description?

PR #1219, opened 2025-10-25
Fix: correct app name in Windows notification

ilgabbo:
Hi!
On Windows, when enabled, the path to the executable appeared in the name of the notification application instead of the actual name.

Before:

https://github.com/user-attachments/assets/885877c1-223c-4068-bc45-a72a276dd714

After:

https://github.com/user-attachments/assets/c185d7fb-1692-4c85-b3b7-0657701f20f3

It's a small change, but I think it could be useful for those who are meticulous (or for those who use Windows).
Thanks!

### DIFF
--- a/src/main/index.ts
+++ b/src/main/index.ts
@@ -295,6 +295,10 @@ async function createWindow(first = true): Promise<void> {
         ...(nativeFrame && isWindows() && nativeFrameConfig.windows),
     });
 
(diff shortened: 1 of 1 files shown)


mihawk90:
This is a duplicate of #1197

ilgabbo:
My bad, I'll close the pr

PR #1220, opened 2025-10-25
Add explicit status for Navidrome and OpenSubsonic

Lyall-A:
This allows sorting by explicit status and displays it in album/track info page (either `Explicit`, `Clean` or empty)

image
image
image


src/shared/types/domain-types.ts
```diff
@@ -160,6 +160,7 @@ export type Album = {
     comment: null | string;
     createdAt: string;
     duration: null | number;
+    explicitStatus: null | number;
```
kgarner7: Use an enum instead of just a number.

src/shared/types/domain-types.ts
```diff
@@ -332,6 +333,7 @@ export type Song = {
     discNumber: number;
     discSubtitle: null | string;
     duration: number;
+    explicitStatus: null | number;
```
kgarner7: Same here

src/shared/api/subsonic/subsonic-normalize.ts
```diff
@@ -146,6 +146,7 @@ const normalizeSong = (
         discNumber: item.discNumber || 1,
         discSubtitle: null,
         duration: item.duration ? item.duration * 1000 : 0,
+        explicitStatus: null,
```
kgarner7: OpenSubsonic actually exposes this field. See https://opensubsonic.netlify.app/docs/responses/child/#:~:text=of%20the%20song.-,explicitstatus,-string

src/shared/api/subsonic/subsonic-normalize.ts
```diff
@@ -247,6 +248,7 @@ const normalizeAlbum = (
         comment: null,
         createdAt: item.created,
         duration: item.duration * 1000,
+        explicitStatus: null,
```
kgarner7: Same here, although the behavior is different: https://opensubsonic.netlify.app/docs/responses/albumid3/#:~:text=is%20a%20compilation.-,explicitstatus,-string

src/renderer/features/item-details/components/item-details-modal.tsx
```diff
@@ -128,6 +128,11 @@ const AlbumPropertyMapping: ItemDetailRow<Album>[] = [
     },
     { key: 'releaseYear', label: 'filter.releaseYear' },
     { key: 'songCount', label: 'filter.songCount' },
+    {
+        label: 'filter.explicitStatus',
+        render: (album) =>
```
kgarner7: It would be better to translate this. You can pass in the `TFunction` to `ItemDetailRow.render` instead.

kgarner7 (changes requested):
`explicitStatus` is an OpenSubsonic concept that Navidrome is storing differently. Additional considerations:
1. You can filter by tracks that _are_ explicit in Navidrome (`e`), but for whatever reason you cannot search by tracks that are _not_ explicit
2. Jellyfin has the concept of `OfficialRating`, I couldn't really map that to an explicit/clean (and it's meant more for videos anyway). No clue if anyone actually passes that for _audio_.

Lyall-A:
I switched to enum instead of number for the explicit status and added support for OpenSubsonic (only tested with Navidrome server though). I also added the explicit status to the cards of albums and tracks.

I considered adding support for it on Jellyfin, but as you said it seems to be parental ratings meant for videos and doesn't seem to pull it from metadata

PR #1223, opened 2025-10-26
fix: respect replaygain during crossfade to avoid volume spikes

### LINKED ISSUE #1109: [Bug]: Sound become lower when playback style is set to "Crossfade"

- [x] Yes

### App Version

0.20.0

### Music Server and Version

Navidrome v0.58.0

Desktop Windows, Web Firefox

### What happened?

When I select Crossfade, the sound should get quieter at the end of the track and return to normal pitch shortly after the next track begins.

Here, the sound gradually fades out at the end of one track and rises too slightly at the beginning of the other. When you click on the track's progress bar of this song, the sound suddenly becomes much louder.

### Steps to reproduce

1. Go to "Settings"
2. Click on "Player" tab
3. Set "playback style" to "Crossfade"
4. Start playing an album
5. At the end of the first track, the sound should fade down and then up again on the next track.
6. Click on the player bar and hear the sound suddenly rise.

### Relevant log output

```shell

```

zeroquinc:
- Normalize each player with its own replaygain multiplier during crossfades so the second deck no longer jumps in at the prior track’s gain.

- Reapply those per-track gains immediately and on subsequent frames when we seek in Now Playing, preventing the brief volume spike that used to happen after a scrub.

Fixes: https://github.com/jeffvli/feishin/issues/1109


jeffvli:
Web player was rewritten so this is now outdated

PR #1236, opened 2025-11-02
Refactor add to playlist modal

### LINKED ISSUE #43: Better add to playlist
The current UI is a bit annoying because you need to click the drop down every time, even if you only want to add the songs to one playlist.

The ideal scenario would be to implement something similar to what YouTube has. A list of all the playlists and a checkmark next to them representing if the song is already in a playlist or not.

A good middle ground and a quick fix would be to leave it as is, but to also add a list of all the playlists bellow the current dropdown. That way if you only want to add songs to one playlist, it's one click, and you can still use the dropdown if you want to add them to multiple playlists.

kgarner7:
The goal of this PR is to refactor add to playlist modal to support the following features:
- Ability to create multiple new playlists
- Checkbox select table to select playlists
- Search table to search playlists.

Inspired heavily from the implementation that Navidrome uses.
This also fixes the problem where the select dropdown opens too early.
 
Resolves #43

### DIFF
--- a/src/renderer/features/playlists/components/add-to-playlist-context-modal.tsx
+++ b/src/renderer/features/playlists/components/add-to-playlist-context-modal.tsx
@@ -1,6 +1,7 @@
-import { useForm } from '@mantine/form';
+import { Pill } from '@mantine/core';
+import { useSelection } from '@mantine/hooks';
 import { closeModal, ContextModalProps } from '@mantine/modals';
-import { useMemo, useState } from 'react';
+import { useCallback, useMemo, useState } from 'react';
 import { useTranslation } from 'react-i18next';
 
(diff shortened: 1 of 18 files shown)


src/renderer/features/playlists/components/add-to-playlist-context-modal.tsx
```diff
@@ -1,6 +1,7 @@
-import { useForm } from '@mantine/form';
+import { Pill } from '@mantine/core';
```
jeffvli: Wrap the component instead of importing from @mantine/core directly

jeffvli:
I'm going to hijack this PR and add a couple styling changes to it.

PR #1239, opened 2025-11-02
Disable media session on Linux only

rodriguezdamien:
This PR changes the condition that are checked to disable the MediaSession : now, only Linux user are excluded from using the media session (as there is already something to interact with MPRIS from what I have understood from the previous PR/issues)

I'm doing this PR because mediaSession works well with the [NowPlaying API](https://developer.apple.com/documentation/mediaplayer/becoming-a-now-playable-app) (or MediaRemote?) of Apple devices and I wanted to be able to interact with Feishin on other apps that also uses this API (like [BoringNotch](https://github.com/TheBoredTeam/boring.notch))

This is my first ever PR on an open source project so please tell me if anything is wrong.

PS: I was about to commit the translation (as I think we do not need to precise that Media Session is Windows only now) but I saw that there is Weblate configured and I don't know how it works...

### DIFF
--- a/src/main/features/core/player/media-keys.ts
+++ b/src/main/features/core/player/media-keys.ts
@@ -25,10 +25,10 @@ export const enableMediaKeys = (window: BrowserWindow | null) => {
         }
     }
 
-    const enableWindowsMediaSession = store.get('mediaSession', false) as boolean;
+    const enableMediaSession = store.get('mediaSession', false) as boolean;
     const playbackType = store.get('playbackType', PlaybackType.WEB) as PlaybackType;
 
-    if (!enableWindowsMediaSession || !isWindows() || playbackType !== PlaybackType.WEB) {
+    if (!enableMediaSession || !isWindows() || playbackType !== PlaybackType.WEB) {
         globalShortcut.register('MediaStop', () => {
             window?.webContents.send('renderer-player-stop');
         });
--- a/src/renderer/features/settings/components/hotkeys/window-hotkey-settings.tsx
+++ b/src/renderer/features/settings/components/hotkeys/window-hotkey-settings.tsx
@@ -16,7 +16,7 @@ export const WindowHotkeySettings = () => {
     const { t } = useTranslation();
     const settings = useHotkeySettings();
     const { setSettings } = useSettingsStoreActions();
-    const { mediaSession: enableWindowsMediaSession, type: playbackType } = usePlaybackSettings();
+    const { mediaSession: enableMediaSession, type: playbackType } = usePlaybackSettings();
 
     const options: SettingOption[] = [
         {
@@ -25,9 +25,7 @@ export const WindowHotkeySettings = () => {
                     defaultChecked={settings.globalMediaHotkeys}
                     disabled={
                         !isElectron() ||
-                        (enableWindowsMediaSession &&
-                            isWindows &&
-                            playbackType === PlaybackType.WEB)
+                        (enableMediaSession && isWindows && playbackType === PlaybackType.WEB)
                     }
                     onChange={(e) => {
                         setSettings({
--- a/src/main/index.ts
+++ b/src/main/index.ts
@@ -548,10 +548,10 @@ async function createWindow(first = true): Promise<void> {
     }
 }
 
-const enableWindowsMediaSession = store.get('mediaSession', false) as boolean;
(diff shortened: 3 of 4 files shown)


ENDlezZenith:
> PS: I was about to commit the translation (as I think we do not need to precise that Media Session is Windows only now) but I saw that there is Weblate configured and I don't know how it works...

I believe you can modify `mediaSession_description` in i18n en.json, and remove the “windows media session integration” and “windows only” part, and probably change that to “web player only” kind of thing?

src/main/features/core/player/media-keys.ts
```diff
@@ -25,10 +25,10 @@ export const enableMediaKeys = (window: BrowserWindow | null) => {
         }
     }
 
-    const enableWindowsMediaSession = store.get('mediaSession', false) as boolean;
+    const enableMediaSession = store.get('mediaSession', false) as boolean;
     const playbackType = store.get('playbackType', PlaybackType.WEB) as PlaybackType;
 
-    if (!enableWindowsMediaSession || !isWindows() || playbackType !== PlaybackType.WEB) {
+    if (!enableMediaSession || !isWindows() || playbackType !== PlaybackType.WEB) {
```
kgarner7, line 31:
```suggestion
    if (!enableMediaSession || isLinux() || playbackType !== PlaybackType.WEB) {
```

src/renderer/features/settings/components/hotkeys/window-hotkey-settings.tsx
```diff
@@ -25,9 +25,7 @@ export const WindowHotkeySettings = () => {
                     defaultChecked={settings.globalMediaHotkeys}
                     disabled={
                         !isElectron() ||
```
kgarner7, line 27: This condition should _also_ be updated, although (`!isLinux()`), but I don't quite understand the original point of this.

kgarner7 (changes requested):
A few changes.
As for the translation, you can just update `en.json` (and any other language are familiar with) `settings.mediaSession_description`

src/renderer/features/settings/components/hotkeys/window-hotkey-settings.tsx
ENDlezZenith: `enableMediaSession && isWindows && playbackType === PlaybackType.WEB` here are bound together for detecting the single case that on windows if `mediaSession` is enabled in web player, to prevent state of `globalMediaHotkeys` to change for conflict. Only if `globalMediaHotkeys` is also conflicting with `mediaSession` in macOS, `isWindows` needs to be changed to `!isLinux`, or it shall remain unchanged I believe?

src/main/features/core/player/media-keys.ts
rodriguezdamien: Tried to implement this, it works but it looks like feishin has trouble managing the hotkeys :
- pressing pause does pause the audio, but does not put feishin on a pause state (only pausing the audio the audio tag or whatever the JS API is ?)
- Skipping and Go Back doesn't work at all

I would need to take a deeper look, but I don't know if it fits this PR purpose, do I rename the PR or make another one?

src/renderer/features/settings/components/hotkeys/window-hotkey-settings.tsx
rodriguezdamien: Just tested what @ENDlezZenith said, and it looks like he is right, mediaSession is in conflict with the globalMediaHotkeys on MacOS, and if i understand correctly the window-hotkey-setting is supposed to be only for linux because of MPRIS?

src/main/features/core/player/media-keys.ts
ENDlezZenith:
```suggestion
    if (isWindows() && enableMediaSession && playbackType == PlaybackType.WEB) {
```

rodriguezdamien:
Hey!
Sorry for not doing any progress on this PR, I've been busy with a lot of things IRL and couldn't work on this. I should have free time soon so I will try to finish it as soon as possible.

Thanks again to everyone who checked it, will comeback to you very soon!

jeffvli:
Resolved in https://github.com/jeffvli/feishin/commit/4d626377efedb00b8fb0e493a98ce078ef9ec61c

Issue #1241, opened 2025-11-02
[Feature]: Support for Album comments


zeroquinc:
### I have already checked through the existing feature requests and found no duplicates

- [x] Yes

### Is this a server-specific feature?

Navidrome

### What do you want to be added?

Navidrome WebUI has support for the `%musicbrainz_albumcomment%` tag.
Could this be added to Feishin as well?

<img width="504" height="315" alt="Image" src="https://github.com/user-attachments/assets/d17e19fc-5aba-40cf-bf9c-6dbc33d2ee59" />

<img width="415" height="264" alt="Image" src="https://github.com/user-attachments/assets/2001fe21-65d9-4a09-8e91-2fe3730732cf" />

Same applies to the `%subtitle%` tag.

<img width="577" height="62" alt="Image" src="https://github.com/user-attachments/assets/c4653ba3-8264-4171-bde5-e8208a88cd39" />

<img width="265" height="64" alt="Image" src="https://github.com/user-attachments/assets/f546e183-1d8e-40be-b786-5dedb62d5a8c" />

Much appreciated!!

PR #1242, opened 2025-11-02
Add additional information to album: record label, release type, version

kgarner7:
- Adds multivalued fields `recordLabels`, `releaseTypes` to album (OpenSubsonic/Navidrome)
- Adds singlevalued `version` field to album (OpenSubsonc/Navidrome)
- For Navidrome, removes used fields and `genre` from the `tags` item
- Uses pills om album detail header for fields. Also include release type and album version in the pills where possible
- Show rating on album for Subsonic servers as well
- Show additional fields in album detail modal.

This is a partial implementation of #1241. It doesn't touch the album card to show new details, and it doesn't touch the table columns either.

### DIFF
--- a/src/i18n/locales/en.json
+++ b/src/i18n/locales/en.json
@@ -498,6 +498,29 @@
         "pause": "pause",
         "viewQueue": "view queue"
     },
+    "releaseType": {
+        "primary": {
+            "album": "$t(entity.album_one)",
+            "broadcast": "broadcast",
+            "ep": "ep",
+            "other": "other",
+            "single": "single"
+        },
+        "secondary": {
+            "audiobook": "audiobook",
+            "audio drama": "audio drama",
+            "compilation": "compilation",
+            "dj-mix": "dj mix",
+            "demo": "demo",
+            "field recording": "field recording",
+            "interview": "interview",
+            "live": "live",
+            "mixtape/street": "mixtape",
+            "remix": "remix",
+            "soundtrack": "soundtrack",
+            "spokenword": "spoken word"
+        }
+    },
     "setting": {
         "accentColor_description": "sets the accent color for the application",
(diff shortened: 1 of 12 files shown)


src/renderer/features/albums/components/album-detail-header.tsx
```diff
@@ -78,7 +81,10 @@ export const AlbumDetailHeader = forwardRef(
             }
         }, detailQuery.data !== undefined);
 
-        const metadataItems = [
+        const releaseTypes =
+            detailQuery.data?.releaseTypes.map((type) => ({ id: type, value: type })) || [];
```
jeffvli: Do you foresee any issues with title casing the releaseTypes and version?

kgarner7: Release type, probably not. Version, absolutely (they can be any string, including a sentence).

jeffvli: Let's switch the releasetypes to title case then.

src/i18n/locales/en.json
```diff
@@ -498,6 +498,29 @@
...
+            "dj-mix": "dj mix",
+            "demo": "demo",
+            "field recording": "field recording",
+            "interview": "interview",
+            "live": "live",
+            "mixtape/street": "mixtape",
+            "remix": "remix",
+            "soundtrack": "soundtrack",
+            "spokenword": "spoken word"
+        }
```
jeffvli, line 522: I'm not a fan of using the raw values as keys here. It makes it inconsistent with the rest of the json schema.

Can you rename them to use camelCase and then add a secondary mapping from the i18n key to the releaseType?

kgarner7:
One note, I don't know if the `primary` release type (this _should_ be a single value) should be displayed specially.

PR #1243, opened 2025-11-04
feat: add 'preservePitch' toggle to playback speed "dropdown"

flleeppyy:
This PR adds the 'preservePitch' setting to the "dropdown" (dropup). I've found myself (as well as a few other friends who use Feishin) wanting a toggle for this quickly accessible in the dropup menu, so I thought I would PR it.

This only shows when the web player is selected. It does not show for the Mpv player.

image

I'm not sure the way I'm doing this is up to par with what standards are being used in this codebase. I could abstract the settings part and make it similar to `useSetCurrentSpeed`?

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -42,7 +42,7 @@
         "dist/**/*": true
     },
     "i18n-ally.localesPaths": ["src/i18n", "src/i18n/locales"],
-    "typescript.tsdk": "node_modules\\typescript\\lib",
+    "typescript.tsdk": "node_modules/typescript/lib",
     "typescript.preferences.importModuleSpecifier": "non-relative",
     "stylelint.config": null,
     "stylelint.validate": ["css", "postcss"],
--- a/src/renderer/features/player/components/right-controls.tsx
+++ b/src/renderer/features/player/components/right-controls.tsx
@@ -14,8 +14,10 @@ import {
     useCurrentSong,
     useHotkeySettings,
     useMuted,
+    usePlaybackSettings,
     usePreviousSong,
     useSettingsStore,
+    useSettingsStoreActions,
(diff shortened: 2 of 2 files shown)


Lyall-A:
i'm not a collaborator/reviewer, but i think this option should only be shown if using web player

flleeppyy:
> i'm not a collaborator/reviewer, but i think this option should only be shown if using web player

Preserve pitch words both for the webplayer and the mpv player, assuming you're thinking preserve pitch only works for the web player.

Lyall-A:
> > i'm not a collaborator/reviewer, but i think this option should only be shown if using web player
> 
> 
> 
> Preserve pitch words both for the webplayer and the mpv player, assuming you're thinking preserve pitch only works for the web player.

I'm sure it's possible to add it for MPV, but the option was only implemented for web and is hidden from settings when MPV is chosen (I implemented this feature originally)

flleeppyy:
> > > i'm not a collaborator/reviewer, but i think this option should only be shown if using web player
> > 
> > 
> > Preserve pitch words both for the webplayer and the mpv player, assuming you're thinking preserve pitch only works for the web player.
> 
> I'm sure it's possible to add it for MPV, but the option was only implemented for web and is hidden from settings when MPV is chosen (I implemented this feature originally)

https://github.com/user-attachments/assets/08796846-9d69-4f76-af45-05652d835d4d

buhh?

Lyall-A:
image
tbh idk what that toggle means exactly but it isn't mpv

flleeppyy:
Ohhh im blind, lemme test

flleeppyy:
Okay, I updated it. I had thought that the web audio toggle WAS the switcher for Mpv or the web player, but I was extremely blind lol

Issue #1246, opened 2025-11-07
[Feature]: Way to shuffle *by* albums


eramdam:
### I have already checked through the existing feature requests and found no duplicates

- [x] Yes

### Is this a server-specific feature?

Navidrome

### What do you want to be added?

What the title says, this could either be a "shuffle albums" button at the top of the Albms view like [Supersonic](https://github.com/dweymouth/supersonic) does

<img width="495" height="142" alt="Image" src="https://github.com/user-attachments/assets/0cc306db-395b-4576-9fc0-ed32a3894b2c" />

Or a way to set the "shuffle type" to be Albums instead of Songs. I see that [`PlayerShuffle` has an `ALBUM` member](https://github.com/jeffvli/feishin/blob/8c539aaadccf353747b31cbd867850133054586f/src/shared/types/types.ts#L126-L127) but as far as I can tell there isn't any code that uses it to influence the shuffle logic.

To clarify, given the following albums:

- Album 1
   - Song 1
   - Song 2
   - Song 3
   - Song 4
- Album 2
   - Song A
   - Song B
   - Song C
   - Song D
- Album 3
   - Song I
   - Song II
   - Song III
   - Song IV

A "shuffle by album" would result in the *albums* being shuffled, keeping the songs order inside them intact.

- Album 1
   - Song 1
   - Song 2
   - Song 3
   - Song 4
- Album 3
   - Song I
   - Song II
   - Song III
   - Song IV
- Album 2
   - Song A
   - Song B
   - Song C
   - Song D

I tried hacking together a "shuffle albums" button in the Albums view but I couldn't wrap my head around the code to have the behavior I wanted. I could shuffle `albumIds` in `handlePlay` under `album-list-route` just fine but the resulting queue wouldn't be random at all (would always play the same 4 albums at the start). 

<img width="483" height="135" alt="Image" src="https://github.com/user-attachments/assets/3f8e653d-3b32-40e9-bcfc-2e1c2c24d254" />

eramdam:
Oops, didn't see https://github.com/audioling/audioling/issues/49 existed already.

PR #1247, opened 2025-11-08
Use grid/table data to build queue when playing from Albums view

### LINKED ISSUE #49: [Feature]: Shuffle by album

- [x] Yes

Not server-specific

### Describe the solution you'd like

A feature which was part of iTunes, but which seems to be lacking from modern music players is "[shuffle by album](https://kirkville.com/how-to-shuffle-albums-in-itunes/)". When this mode is chosen, albums, not songs, are played at random while conserving the order of songs on the album. 

This would be a useful option for those of us whose listening is album-oriented, and makes it possible to make, e.g., "music for dinners", "music for writing" convenience playlists. I would love to see it implemented in a player with Subsonic support. It could be implemented as a menu item, like in iTunes, or perhaps hidden behind a right-click of the shuffle button.

Also, thank you for all the work you're doing on feishin (and now audioling) – killer apps!

### Additional context

_No response_

eramdam:
This started as a way to fix https://github.com/audioling/audioling/issues/49 https://github.com/jeffvli/feishin/issues/1246 but ended up fixing more UX issues (IMO) with Feishin's albums view. 

As of right now, pressing "play" at the top of the Albums view is kind of unpredictable because the queue that gets built doesn't match what the user sees. 

This PR changes that by using the grid/table's data to build the list of album IDs passed to `handlePlayQueueAdd` which enables a bunch of cool behaviors:
- Queue random albums by switching the list to the "random" sort and pressing play (fixes https://github.com/audioling/audioling/issues/49 https://github.com/jeffvli/feishin/issues/1246)
- Queue that respects the sorting and filtering used

Here's a demo video showing the behavior

https://github.com/user-attachments/assets/0a11fce2-054c-40a3-a9db-20e32ef8df12

### DIFF
--- a/src/renderer/components/virtual-grid/virtual-infinite-grid.tsx
+++ b/src/renderer/components/virtual-grid/virtual-infinite-grid.tsx
@@ -18,6 +18,7 @@ import { AnyLibraryItem, Genre, LibraryItem } from '/@/shared/types/domain-types
 import { ListDisplayType } from '/@/shared/types/types';
 
 export type VirtualInfiniteGridRef = {
+    getItemData: () => (LibraryItemOrGenre | undefined)[];
     resetLoadMoreItemsCache: () => void;
     scrollTo: (index: number) => void;
     setItemData: (data: LibraryItemOrGenre[]) => void;
@@ -143,6 +144,9 @@ export const VirtualInfiniteGrid = forwardRef(
         const debouncedLoadMoreItems = debounce(loadMoreItems, 500);
 
         useImperativeHandle(ref, () => ({
+            getItemData: () => {
+                return itemData;
+            },
             resetLoadMoreItemsCache: () => {
                 if (loader.current) {
                     loader.current.resetloadMoreItemsCache(false);
--- a/src/renderer/features/albums/routes/album-list-route.tsx
+++ b/src/renderer/features/albums/routes/album-list-route.tsx
@@ -1,12 +1,11 @@
 import type { AgGridReact as AgGridReactType } from '@ag-grid-community/react/lib/agGridReact';
 
 import { useQuery } from '@tanstack/react-query';
+import { compact } from 'lodash';
 import isEmpty from 'lodash/isEmpty';
 import { useCallback, useMemo, useRef } from 'react';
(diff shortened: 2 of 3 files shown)


eramdam:
I can't access the Vercel deploy logs but it looks like it successfully deplyoed in the end

jeffvli:
I appreciate your work on this!

Unfortunately, I'm currently in the process of reworking the entirety of the queue and all the playback handlers in the app so this would be overwritten by my changes in the branch https://github.com/jeffvli/feishin/tree/feat/new-lists.

I will take this behavior into account with my new implementation.

eramdam:
> I appreciate your work on this!
> 
> Unfortunately, I'm currently in the process of reworking the entirety of the queue and all the playback handlers in the app so this would be overwritten by my changes in the branch [`feat`/new-lists](https://github.com/jeffvli/feishin/tree/feat/new-lists?rgh-link-date=2025-11-11T05%3A52%3A57Z).
> 
> I will take this behavior into account with my new implementation.

No worries, makes total sense. I'd be happy to give the new implementation a test drive when available.

eramdam:
@jeffvli do you know in which 1.x beta I can expect this behavior to be backported into? Worst case I can try drafting up a PR against `development` to add that behavior, up to you :)

jeffvli:
This is already implemented in the current beta version *except* the random sorting behavior since I forgot about that.

Let me take a quick look into it.

jeffvli:
It is now properly implemented - but probably not in the way you expect.

image

When using the big header Play button, it will not follow the Random order that is applied in the list, but instead use a new custom random order. In your original implementation, you derived the order from the list data itself, but that would not work in cases where the list has unloaded items. The random sorting mode also does not take into account duplicates that can occur between pages since I don't believe a seed is used to uniquely identify pages in the server.

If you want it to match your existing behavior, where you want to maintain the shuffled order that you can see, you can use the new `CTRL+A` select all functionality on the list page to select all loaded items and then either drag them into the playqueue or use the context menu to add them.

eramdam:
Great, will give it a try in the next beta :)

PR #1263, opened 2025-11-17
fix(rpc): truncate certain fields

kgarner7:
Apparently, the max length for some fields is 128. Truncate text fields of discord rpc to 127, with `'…';`


mihawk90:
Works :+1:

PR #1307, opened 2025-12-02
Updated hungarian langauge

Soderes:
Translating update according to the latest source.
Fix and fine tune, interpretation corrections.


jeffvli:
Did you update these on weblate? If so, you don't need to submit a PR for this as I always merge in the translations before releasing.

Soderes:
Yes, of course. But it's easier for me here. The problem with Weblate is that you can't see how it will look in the app.

PR #1333, opened 2025-12-07
feat(subsonic): support form post

kgarner7:
Automatically use url-encoded form post for OpenSubsonic servers that support it.

Making this a PR instead of a push, for additional testing.


jeffvli:
Looks good. I just renamed the feature enum to be more specific.

PR #1335, opened 2025-12-08
feat: sync play queue for navidrome/subsonic

kgarner7:


src/renderer/api/navidrome/navidrome-controller.ts
```diff
@@ -847,6 +882,30 @@ export const NavidromeController: InternalControllerEndpoint = {
...
+                },
+            });
+
+            if (res.status !== 200) {
+                throw new Error('Failed to save play queue');
+            }
+        }
+
+        return SubsonicController.savePlayQueue(args);
+    },
```
jeffvli, line 909: This logic seems a bit off to me?

It's running both the ND and SS endpoints if both are available.

The comment seems to indicate you probably want an early return for the `hasFeatureWithVersion(apiClientProps.server, ServerFeature.SERVER_PLAY_QUEUE, 1)`, with a fallback to the `hasFeatureWithVersion(apiClientProps.server, ServerFeature.SERVER_PLAY_QUEUE, 2)` and not implemented if both are false.

kgarner7: This is what happens when you forget a `return`...

src/renderer/features/player/components/right-controls.tsx
```diff
@@ -347,6 +353,120 @@ const RatingButton = () => {
...
+            .catch((error) => {
+                toast.error({
+                    message: 'This is most likely because your queue is too large (> 1000 tracks)',
+                    title: 'Failed to save play queue',
+                });
+                console.error(error);
+            });
+    }, [server?.id]);
+
+    return server?.type === ServerType.JELLYFIN ? null : (
```
jeffvli: Use `const hasServerPlayQueue = hasFeature(server, ServerFeature.SERVER_PLAY_QUEUE);` for the condition rather than ServerType.JELLYFIN.

jeffvli: I'm still also unsure if these buttons should be included in the player bar. It might make more sense for them to be moved to the PlayQueueListControls.

kgarner7: I thought about that at one point, but I'm also worried that there isn't much space in the play queue list control. With stock scaling, it looks like the minimum to _fully_ fit everything in the sidebar play queue control (with search open) is 537px.
It wrappers around better now, so it might just work then.

jeffvli: You can remove all the existing buttons on the left side except for the shuffle and clear queue buttons to fit them in.

I don't think they have much utility since the existing functionality is already included in the context menu.

src/renderer/features/now-playing/components/play-queue-list-controls.tsx
```diff
@@ -1,85 +1,121 @@
...
+        player.shuffleAll();
     };
 
-    const handleRemoveSelected = () => {
-        const selectedItems = tableRef?.current?.internalState.getSelected() as
-            | QueueSong[]
-            | undefined;
-        if (!selectedItems || selectedItems.length === 0) return;
+    const handleSaveQueue = useCallback(() => {
+        if (serverId) return;
```
jeffvli: Wrong condition. Should be `!serverId`

src/renderer/features/now-playing/components/play-queue-list-controls.tsx
```diff
@@ -91,46 +127,37 @@ export const PlayQueueListControls = ({
...
+                        />
+                        <ActionIcon
+                            icon="download"
+                            onClick={handleRestoreQueue}
+                            size="sm"
+                            tooltip={{
+                                label: t('player.restoreQueue', { postProcess: 'titleCase' }),
+                                openDelay: 0,
+                            }}
+                            variant="subtle"
```
jeffvli: The tooltip openDelay: 0 is inconsistent with the other buttons. 

Also, move them to the left of the existing controls. The clear button should be on the very right.

kgarner7:
The one area that I can't seem to get working is, restoring the same _position_ when restoring the queue. It does work if the track is _somehow_ already loaded, but not otherwise.

PR #1361, opened 2025-12-10
Add Track Artist to display listing for Compilations

frcooper:
If the displayed album is tagged as a compilation, display the artist as well as the title in the track listing on the album display page.

### DIFF
--- a/src/renderer/features/albums/components/album-detail-content.tsx
+++ b/src/renderer/features/albums/components/album-detail-content.tsx
@@ -52,7 +52,7 @@ import {
     SongListSort,
     SortOrder,
 } from '/@/shared/types/domain-types';
-import { ItemListKey, ListDisplayType, Play } from '/@/shared/types/types';
+import { ItemListKey, ListDisplayType, Play, TableColumn } from '/@/shared/types/types';
 
 interface AlbumMetadataTagsProps {
     album: Album | undefined;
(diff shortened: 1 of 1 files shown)


jeffvli:
I'm not sure I agree with this change.

The columns on the album page are already fully configurable. This would result in a user's custom column configuration being overridden potentially in a way that they would not want it.

frcooper:
> The columns on the album page are already fully configurable. This would result in a user's custom column configuration being overridden potentially in a way that they would not want it.

This modifies the column config for a specific kind of album. The current configuration system applies to all album listings. To make the configuration equivalent would require a rule based system for selecting custom table layouts.  Honestly, while I can probably hack out the backend to that, I have no idea what a useful UI would be.

jeffvli:
Right - but what about the case where a user is viewing a compilation album but does not care about viewing the artists column?

This would force a certain table configuration onto the user, when my intention is to allow it to remain custom based on what the user has configured themselves to see.

Since this is a pretty opinionated UX change (one I don't agree with), I don't feel comfortable merging it in. I appreciate the PR though, and this would definitely be a good fit in a forked build!

jeffvli:
Thinking about this some more - if you want to add a configuration toggle for this behavior I may be willing to merge this in.

Feel free to re-open if so.

PR #1369, opened 2025-12-11
Increase ItemCard image resolution to 300x300 in 'poster' mode

### LINKED ISSUE #1330: [Bug]: Tracks look very low-res after doing any action (search, click on)

- [x] Yes

### App Version

0.22.0

### Music Server and Version

Jellyfin 10.11.3

Desktop Linux

### What happened?

This is how it looks before I start searching:
Image

This is how it looks after I start searching
Image
As you can see, the cover arts are much lower-res here, which is rather inconvenient. I would expect them to display at a decent res. I tried putting "Player album art resolution" to 500px, and 0. Both had this issue.

This same issue also happens when I click on individual tracks:
Image
Image

This is weird because it doesn't affect the cover art on the bottom left, which displays in a good resolution no matter what.

### Steps to reproduce

1. Go to tracks
2. Search for something, anything
3. Everything is low-res
---- or:
1. Go to tracks
2. Click on a track
3. the cover art will be low res on the page itself

### Relevant log output

```shell

```

baahrens:
Increased the resolution slightly for tracks in the search view (when using grid displayType). I used the same pattern as in other places where we need a certain resolution.

### Before
Screenshot 2025-12-11 at 14 23 41

### After
Screenshot 2025-12-11 at 14 20 35

Closes #1330

### DIFF
--- a/src/renderer/components/item-card/item-card.tsx
+++ b/src/renderer/components/item-card/item-card.tsx
@@ -70,6 +70,11 @@ export const ItemCard = ({
     const imageUrl = getImageUrl(data);
     const rows = providedRows || [];
 
+    const upsizedImageUrl = imageUrl
+        ?.replace(/&size=\d+/, '&size=300')
+        .replace(/\?width=\d+/, '?width=300')
+        .replace(/&height=\d+/, '&height=300');
+
     switch (type) {
         case 'compact':
             return (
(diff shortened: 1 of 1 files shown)


jeffvli:
I'm currently reworking the app the generate all image urls at runtime which should resolve this issue.

https://github.com/jeffvli/feishin/compare/development...feat/runtime-images

PR #1379, opened 2025-12-12
Light mode fixes

eramdam:
Fixes a few light mode CSS issues I've noticed when using the current beta:

- Fixes the color of labels inside of text/icon buttons in the album page
CleanShot 2025-12-12 at 10 50 54@2x
- Tweaks colors in the carousel to look better 
CleanShot 2025-12-12 at 10 29 23@2x
- Uses a light color for the scroll shadows in tables


jeffvli:
Looks good, thanks

PR #1383, opened 2025-12-14
Export lyrics

### LINKED ISSUE #1286: [Feature]: Add button for saving lyrics locally to a file

- [x] Yes

Not server-specific

Basically what the title says. 
I want a button in the lyrics screen where I can just download the lyrics being shown and save it as an LRC file or as a plaintext file. Considering you're able to get and display lyrics from remote sources like NetEase and Genius, it would be logical that you would be able to also just save said lyrics locally as a plaintext or LRC file. 
I know that other dedicated lyrics tools exist, but considering that Feishin is already downloading and displaying the lyrics, it would be logical to have the ability to download the lyrics.

kgarner7:
Support exporting lyrics to text (unsynced, synced) or lrc (synced only). The export button could probably go in a better place though...

### DIFF
--- a/src/renderer/features/lyrics/components/lyrics-export-form.tsx
+++ b/src/renderer/features/lyrics/components/lyrics-export-form.tsx
@@ -1,6 +1,6 @@
 import { closeAllModals, openModal } from '@mantine/modals';
 import formatDuration from 'format-duration';
-import { useMemo } from 'react';
+import { useCallback, useMemo } from 'react';
 import { useTranslation } from 'react-i18next';
 
 import i18n from '/@/i18n/i18n';
@@ -39,10 +39,11 @@ export const LyricsExportForm = ({ lyrics, offsetMs, synced }: LyricsExportFormP
             return `[ar:${lyrics.artist}]
 [ti:${lyrics.name}]
 [offset:${form.values.offsetMs + (lyrics.offsetMs ?? 0)}]
-${contents}`;
+${contents}
+`;
         } else {
             if (Array.isArray(lyrics.lyrics)) {
-                return lyrics.lyrics.map((lyric) => lyric[1]).join('\n');
+                return lyrics.lyrics.map((lyric) => lyric[1]).join('\n') + '\n';
             }
             return lyrics.lyrics;
         }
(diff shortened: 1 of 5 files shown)


src/renderer/features/lyrics/lyrics.tsx
jeffvli, line 264: It seems like the `lang` property is missing from some of the lyrics types (notably on the override) which makes it so that the export button is always hidden.

src/renderer/features/lyrics/lyrics-actions.tsx
```diff
@@ -46,92 +48,99 @@ export const LyricsActions = ({
...
+                                clearable={false}
+                                data={languages}
+                                onChange={(value) => setIndex(parseInt(value!, 10))}
+                                style={{ bottom: 30, position: 'absolute' }}
+                                value={index.toString()}
+                            />
+                        )}
+                        <Button onClick={onExportLyrics} uppercase variant="subtle">
+                            Export
+                        </Button>
```
jeffvli, line 66: Needs i18n key.

src/renderer/features/lyrics/components/lyrics-export-form.tsx
```diff
@@ -0,0 +1,128 @@
...
+        const lyricsFileLink = document.createElement('a');
+        const lyricsFileUrl = URL.createObjectURL(lyricFile);
+        lyricsFileLink.href = lyricsFileUrl;
+        lyricsFileLink.download = lyricFile.name;
+        lyricsFileLink.click();
+
+        URL.revokeObjectURL(lyricsFileUrl);
+
+        closeAllModals();
+    }, [displayedLyrics, form.values.synced, lyrics.name]);
```
jeffvli, line 77: I'm assuming it is done this way to allow browser compatibility?

It would be a lot more useful if this is implemented with Electron as well with a separate button "Export File to Song Path" to allow users to save it directly to the song path (if reachable).

kgarner7: Honestly, I just yoinked the same implementation used for settings export.

PR #1388, opened 2025-12-15
Implement Glassy Dark theme

pyxfluff:
Adds a glassy dark theme, formerly from the iiPython fork
image
image
image
image

I opted to not left-justify the fullscreen player text because that is a personal thing


jeffvli:
Lint the files please.

Also side note - it seems like the custom css file doesn't work properly with the web version.

pyxfluff:
@jeffvli 
> Also side note - it seems like the custom css file doesn't work properly with the web version.

This isn't something I could fix probably and would require something upstream. I linted the file however and added ignore rules where i couldn't fix something (like the mantine rules)

PR #1421, opened 2025-12-19
feat: show record label copyright at bottom of track listing

iiPythonx:
This PR adds a very basic copyright listing of each record label associated with a given album, with examples as such:

image <br>
image <br>
image

If no labels are present on the given album, nothing will be displayed and no extra space will be taken up.
Additionally, if a release year is not present on the album, it will be left out while the label name remains.


mihawk90:
Not that it's important in a self-hosted context, but this should likely be [℗](https://en.wikipedia.org/wiki/Sound_recording_copyright_symbol), not ©.

The phonographic copyright is different from the copyright of the underlying work, which is why on commercial streaming services you will often see both.

Issue #1398, opened 2025-12-15
[Feature]: UI option to disable the album rating badges


ilikepeaches:
### I have already checked through the existing feature requests and found no duplicates

- [x] Yes

### Is this a server-specific feature?

Not server-specific

### What do you want to be added?

I do not like the new rating "badges" on album covers in the betas.  They are busy, extraneous, and stressful to me in a "you have five unread messages" kind of way.  I found I was able to remove them with 

```
[class*="rating-badge"] {
  display: none !important;
}
```

but an option in the settings GUI to disable them would be better.

mihawk90:
Duplicate of #1385

Issue #1385, opened 2025-12-14
[Feature]: Disabling rating system


wildland7752:
### I have already checked through the existing feature requests and found no duplicates

- [x] Yes

### Is this a server-specific feature?

Not server-specific

### What do you want to be added?

Hi there,

Is there an option to disable the rating star system, or at least hide them in UI. If there is not, please make one.

mihawk90:
You can hide them via Custom CSS:

```css
.fs-item-card-controls-module-rating {
    display: none;
}
```

PR #1426, opened 2025-12-20
feat: adding setting to show/hide user ratings

salguerooo:
This PR is for adding a setting for showing or hiding the user ratings feature. I was about to open an issue about it but I saw a few others that were already mentioning it so I decided to give it a try.

I think I have covered all the views but please do let me know if I missed something.

These issues are related to this: https://github.com/jeffvli/feishin/issues/1385 https://github.com/jeffvli/feishin/issues/1398


pyxfluff:
I'm confused, is this not what custom CSS is for? Why would you want to remove ratings anyways? 

Just seems like bloat for the app and settings menu to me but excuse me if I'm just misinformed

salguerooo:
I come from the Navidrome where you can hide the ratings feature from the UI with an environment variable. I had it turned on initially as it is the default but me and the other person that use the instance really disliked it so we disabled it and never turned back. Now that we're using Feishin as our client we wanted to do the same. I decided to give a shot at implementing it since I saw other issues showing interest in also disabling the ratings feature here.

If this just feels like bloat and doesn't have enough traction to be a feature that's totally fair! I will just go with the custom css solution, but it would be nice to just be able to open the settings and disable it, especially for users that aren't very tech savvy.

Petersmit27:
Chiming in to the discussion, I believe this is a nice feature to have and that Custom CSS should only be recommended as a temporary band-aid fix. 

Ultimately, I want to recommend Feishin to non-technical people and telling them to do something with Custom CSS is unacceptable and greatly hinders adoption. 
Yes, settings bloat is also a big problem that should be addressed, but I think it's better to add more features and address that than to dismiss user feedback and point them to the Custom CSS setting.

pyxfluff:
I understand the part about custom CSS but I am yet to see a single reason why there is any need to hide ratings in the first place which is why I simply deemed this as bloat and something that doesn't need to exist. If you don't use ratings you can just ignore it, I don't see why we need to be storing and displaying yet another option and spending the CPU time to decide whether or not to render it

src/renderer/features/albums/components/album-list-content.tsx
```diff
@@ -104,6 +110,14 @@ export const AlbumListView = ({
         };
     }, [query, overrideQuery]);
 
+    const columns = useMemo(() => {
+        if (showRatings) {
+            return table.columns;
+        } else {
+            return table.columns.filter((column) => column.id !== 'userRating');
+        }
+    }, [table, showRatings]);
```
jeffvli: I would remove this change from all table column instances, as the columns are already configurable to remove the rating column if not desired.

src/i18n/locales/en.json
```diff
@@ -879,6 +879,8 @@
         "preferLocalLyrics": "prefer local lyrics",
         "showLyricsInSidebar_description": "a panel will be added to the attached play queue that displays the lyrics",
         "showLyricsInSidebar": "show lyrics in player sidebar",
+        "showRatings_description": "controls if the ratings feature shows up in the interface",
+        "showRatings": "show arist/album ratings",
```
jeffvli: Rename this string to `show star ratings`, because it's not only on artist/album, but also songs. (also there is a typo)

salguerooo: renamed the label and also ended up also adding the pt-br translations for it

src/renderer/features/albums/components/album-list-content.tsx
salguerooo: I reverted all the changes on the table columns. I ended up having to change how the cards choose to display the rating stars as well. My initial approach was to remove the onRating completely from all controls but this causes weirdness when the user had the ratings hidden on the settings but choose to display the ratings on any of the tables because the ratings would show up and be interactive on hover but clicking on it would have no effect and I found that interaction to be weird.

src/renderer/components/item-card/item-card.tsx
jeffvli, line 803: Seems like you forgot to pass the showRating to the ItemCardControls for the DefaultItemCard and PosterItemCard

salguerooo: Added it to where it was missing. I ran the typecheck and it did return some errors but nothing related to this PR.

PR #1429, opened 2025-12-21
Fix API parameter for Jellyfin 10.12 and up

ItzELECTR0:
Added a version check for Jellyfin servers 10.11+ to use the new "ApiKey" syntax.
Also added a API_KEY_CAMELCASE feature flag for this case.

Marked as draft because I've only tested it on my machine with my server.

### DIFF
--- a/src/renderer/api/jellyfin/jellyfin-controller.ts
+++ b/src/renderer/api/jellyfin/jellyfin-controller.ts
@@ -38,12 +38,6 @@ const MAX_ITEMS_PER_PLAYLIST_ADD = 50;
 
 const VERSION_INFO: VersionInfo = [
     [
-        '10.12.0',
-        {
-            [ServerFeature.API_KEY_CAMELCASE]: [1],
-        },
-    ],
-    [
         '10.9.0',
         {
             [ServerFeature.LYRICS_SINGLE_STRUCTURED]: [1],
@@ -52,13 +46,6 @@ const VERSION_INFO: VersionInfo = [
     ],
     ['10.0.0', { [ServerFeature.TAGS]: [1] }],
 ];
-
-const getApiKeyParam = (server: any): string => {
-    if (hasFeature(server, ServerFeature.API_KEY_CAMELCASE)) {
-        return 'ApiKey';
-    }
-    return 'api_key';
-};
 
 export const JellyfinController: InternalControllerEndpoint = {
     addToPlaylist: async (args) => {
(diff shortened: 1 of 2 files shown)


jeffvli:
This will definitely need additional consideration, because I'm on the latest stable `10.11.5`, and the `api_key` parameter is still valid.

Renaming `api_key` to `ApiKey` would be a breaking change from Jellyfin's API so I'm not sure why they would do that in a minor version bump.

I'll leave this PR up, but let's wait to see if this change makes it into Jellyfin's stable releases first.

ItzELECTR0:
It is REALLY strange. I'm going to try and test some more things and see if I can find anything.

salguerooo:
So I was taking a look at the jellyfin versions and looks like `api_key` has been deprecated for a while and the preferred way is `ApiKey`. Maybe on 10.12 they will drop it for good?

Some links for reference:
https://github.com/jellyfin/jellyfin/pull/15559
https://github.com/jellyfin/jellyfin/pull/13306
https://gist.github.com/nielsvanvelzen/ea047d9028f676185832e51ffaf12a6f

ItzELECTR0:
I downgraded my server from 10.12 to 10.11.5 and yeah that does still use `api_key` so I have virtually no clue why it didn't work without this patch. Strangely enough, it does work. I also tried copying the Streaming URL and replaced `api_key` with `ApiKey` and it worked, which probably means they're in some sort of transition phase? I have no idea what's going on or why it works this way but 10.11.5 can take both. I then went back to 10.12 to try the reverse (replace `ApiKey` with `api_key`) but that did not work. So 10.12 can only take `ApiKey` which is very bizarre.

ItzELECTR0:
> So I was taking a look at the jellyfin versions and looks like `api_key` has been deprecated for a while and the preferred way is `ApiKey`. Maybe on 10.12 they will drop it for good?
> 
> Some links for reference: [jellyfin/jellyfin#15559](https://github.com/jellyfin/jellyfin/pull/15559) [jellyfin/jellyfin#13306](https://github.com/jellyfin/jellyfin/pull/13306) [gist.github.com/nielsvanvelzen/ea047d9028f676185832e51ffaf12a6f](https://gist.github.com/nielsvanvelzen/ea047d9028f676185832e51ffaf12a6f)

This popped up just as I finished typing. Wow.

jeffvli:
Good find.

For this PR, you should pin the version for the feature to be `10.12.0` instead.

ItzELECTR0:
Fair enough.

jeffvli:
I decided to simplify the approach here to just force the new `ApiKey` parameter, as the legacy one has been deprecated for some time now.

PR #1437, opened 2025-12-22
feat: add artist radio and track radio (in context menu)

farfromrefug:
This PR adds:

* a new Artist Radio button in Artist page header
* a new context menu on tracks to start track radio

The artist radio use appropriate API for all server kind.
I am just not sure about the artist radio button style. Could not get it to use the same style as play button. But do we want the same style?

Possible improvements:
* add artist radio entry in the context menu for artists
* add artist radio button to artist list item "overlay" (with play button and others)


src/renderer/features/artists/components/album-artist-detail-header.tsx
```diff
@@ -63,8 +65,31 @@ export const AlbumArtistDetailHeader = forwardRef((_props, ref: Ref<HTMLDivEleme
...
+        try {
+            const artistRadioSongs = await queryClient.fetchQuery(
+                songsQueries.artistRadio({
+                    query: {
+                        artistId: routeId,
+                        count: artistRadioCount,
+                    },
+                    serverId: server.id,
+                }),
+            );
```
jeffvli: I would like to see the `queryKeys.player.fetch` query key utilized on these fetches, as well as attaching loading state to the button using the `useIsPlayerFetching` hook, otherwise the user can press the button multiple times without any indication that a fetch is happening.

src/renderer/features/context-menu/actions/play-track-radio-action.tsx
```diff
@@ -0,0 +1,51 @@
...
+            const similarSongs = await queryClient.fetchQuery(
+                songsQueries.similar({
+                    query: {
+                        albumArtistIds: song.albumArtists.map((art) => art.id),
+                        count: artistRadioCount,
+                        songId: song.id,
+                    },
+                    serverId,
+                }),
+            );
```
jeffvli: Same as the other comment - attach the `queryKeys.player.fetch` query key to this fetch.

src/renderer/features/context-menu/menus/song-context-menu.tsx
```diff
@@ -28,6 +29,7 @@ export const SongContextMenu = ({ items, type }: SongContextMenuProps) => {
             bottomStickyContent={<ContextMenuPreview items={items} itemType={type} />}
         >
             <PlayAction ids={ids} itemType={LibraryItem.SONG} songs={items} />
+            <PlayTrackRadioAction song={items[0]} />
```
jeffvli: It would also make sense to add this to the QueueContextMenu and PlaylistSongContext menus as those are both derivatives of the `Song` type.

src/renderer/features/shared/components/library-header.tsx
```diff
@@ -234,6 +239,21 @@ export const LibraryHeaderMenu = ({
...
+                            root: styles.wideTextButton,
+                        }}
+                        leftSection={<Icon icon="radio" />}
+                        onClick={onArtistRadio}
+                        size="md"
+                        variant="subtle"
+                    >
+                        {t('page.albumArtistDetail.artistRadio', { postProcess: 'sentenceCase' })}
+                    </Button>
+                )}
```
jeffvli, line 260: I'm not sold on the design here.

I would:
1. remove the className and classNames as they are specifically only applied for the special text play button styles.
2. Use variant `transparent` instead

src/renderer/features/context-menu/menus/song-context-menu.tsx
farfromrefug: @jeffvli sure but QeueContextMenu and PlaylistSongContext are for multiple ids? How would i choose the one to pass to PlayTrackRadioAction?

jeffvli: The same way it's being handled in the SongContextMenu? Either pick the first item, or disable the item if more than one is selected. Preferably disable now that I think about it.

farfromrefug: @jeffvli done!

farfromrefug:
@jeffvli i added a new menu entry for artist context menu to start radio.

src/renderer/features/context-menu/menus/song-context-menu.tsx
jeffvli: This also needs to be disabled to match the others.

src/renderer/features/context-menu/menus/artist-context-menu.tsx
```diff
@@ -30,6 +31,7 @@ export const ArtistContextMenu = ({ items, type }: ArtistContextMenuProps) => {
             <PlayAction ids={ids} itemType={LibraryItem.ARTIST} />
             <ContextMenu.Divider />
             <AddToPlaylistAction items={ids} itemType={LibraryItem.ARTIST} />
+            <PlayArtistRadioAction artist={items[0]} disabled={items.length > 1} />
```
jeffvli: Also add this to the album-artist-context-menu. Most of the app uses album artist as opposed to artist.

jeffvli:
Another aspect of this feature that I feel may be desirable is to be able to add the additional play options (play / next / last) when starting the radio, in case the user does not want to overwrite their entire queue.

If you don't feel comfortable implementing this, let me know. It's not a requirement, but definitely a nice to have.

src/renderer/features/artists/components/album-artist-detail-header.tsx
farfromrefug: i added the missing loading indicator

farfromrefug:
@jeffvli i addressed all comments. As per additional play options (play / next / last) to be honest i am not sure what you are thinking about. So i would prefer to leave it to you if that s ok.
One thing i was thinking about is an option to open fullscreen player when started. Right now you might be wondering what s happening

PR #1442, opened 2025-12-24
fix:music filtering subsonic

farfromrefug:
this PR fixes subsonic `musicFolderId` handling for most home sections.
Ai helped me on the getGenres. But i am not sure we want that solution. It seems pretty heavy and requires a lot of remote requests


jeffvli:
Yeah, definitely not on the getGenreList. The other requests should take care of filtering songs out by the musicFolderId.

farfromrefug:
@jeffvli done i removed the custom getGenre logic

PR #1476, opened 2026-01-01
feat: playlist grid view

Fluttrr:
I added an option to view playlists in grid view mode. 

image

I left the playlist edit feature as a list because I'm not sure displaying that as a grid makes much sense. I apologize if anything I did is bad I'm super new to open source contributions.. please let me know if there is anything I did wrong or could improve!!


jeffvli:
> I left the playlist edit feature as a list because I'm not sure displaying that as a grid makes much sense.

Yep, that was the correct move.

PR looks good to me.

Issue #1467, opened 2025-12-31
[Bug]: Chrome crashes when used with the Subsonic protocol.


chenqimiao:
### I have already checked through the existing bug reports and found no duplicates

- [x] Yes

### App Version

1.1.0

### Music Server and Version

Qm-music v1.9.9

### What local environments are you seeing the problem on?

Web Chrome

### What happened?

First of all, I sincerely thank feishin for the recent feature updates. I'm currently encountering a very confusing problem, and I haven't seen any error logs, neither on the server nor the client side. After upgrading feishin from version 0.22.0 to the latest version 1.1.0, I found that the feishin client crashes as shown in the image. This might be due to a protocol incompatibility issue on the server side. I don't know how to troubleshoot this, and I hope you can provide some guidance. Thank you.

<img width="1521" height="618" alt="Image" src="https://github.com/user-attachments/assets/7fcc8769-26ff-4666-993d-c6ccb68d8a6b" />

<img width="1534" height="508" alt="Image" src="https://github.com/user-attachments/assets/11eb7566-c23f-4fda-82ad-0338e290fdad" />

<img width="1550" height="507" alt="Image" src="https://github.com/user-attachments/assets/6ba0273c-52c1-48ec-8dd2-717301f50c7f" />

### Steps to reproduce

1. Deploy qm-music
2. Deploy the latest version of Feishin（1.1.0） using Docker
3. Open the Feishin address in your browser and configure the OpenSubsonic address and account credentials.
4. Access the homepage.
5. The browser crashes.

### Relevant log output

```shell

```

jeffvli:
I just set up qm-music and it is not crashing on my browser nor do I see any indications that it should crash.

You can try:

1. Clearing your browser cache
2. Delete all localstorage / indexed db items
3. Try a different browser

Unrelated, but one thing I noticed is that you are returning the `track` value on song elements as a string instead of an number/integer which is causing issues with the trackNumber used in Feishin.

[OpenSubsonic](https://opensubsonic.netlify.app/docs/endpoints/getsong/) expects track to be a number.

chenqimiao:
Thank you for your reply and your time.
I debugged the issue in Chrome and found that the root cause was that `artist.name` was an empty string, which led to an infinite loop in the code, as shown in the image below.

<img width="1153" height="468" alt="Image" src="https://github.com/user-attachments/assets/0d47cfd9-8b33-4fb5-86c8-2d016c3fddae" />

[...]

PR #1483, opened 2026-01-02
Avoid an infinite loop when artist.name is an empty string.

chenqimiao:


chenqimiao:
Closes #1467

jeffvli:
Good find. I made some minor adjustments to remove the trim since it is mostly unneeded.

PR #1513, opened 2026-01-05
fix: Sort top songs by play count instead of community rating

### LINKED ISSUE #1512: [Bug]: Top Songs are not ordered by play count in album artist view

- [x] Yes

### App Version

1.2.0

### Music Server and Version

Jellyfin 10.11.5

Desktop Linux

### What happened?

In the album artist view, the top songs for each artist doesn't sort by play count.

Image

After opening the top songs section after clicking view all, the same issue with the order persists.

Image

I validated that there are songs with higher play count values.

Image

### Steps to reproduce

1. Open Feishin 1.2.0
2. Select an album artist
3. Look at Top Songs section

### Relevant log output

```shell

```

jeremylongshore:
## Summary
- Changes the "Top Songs" sort from `CommunityRating` to `PlayCount` for Jellyfin servers

## Problem
The Top Songs section in the album artist view was sorting songs by `CommunityRating,SortName` instead of by play count. This meant the songs displayed as "Top Songs" weren't actually the most played songs, which is the expected behavior for this feature.

## Solution
Changed the `SortBy` parameter in `getTopSongs` for the Jellyfin controller from `CommunityRating,SortName` to `PlayCount,SortName`. This ensures songs are ordered by their actual play count (descending), showing the most played songs first.

## Test plan
- [x] Verify Top Songs section shows songs ordered by play count
- [x] Verify sorting works correctly with Jellyfin server

Fixes #1512

### DIFF
--- a/src/renderer/api/jellyfin/jellyfin-controller.ts
+++ b/src/renderer/api/jellyfin/jellyfin-controller.ts
@@ -1257,7 +1257,7 @@ export const JellyfinController: InternalControllerEndpoint = {
                 IncludeItemTypes: 'Audio',
                 Limit: query.limit,
                 Recursive: true,
-                SortBy: 'CommunityRating,SortName',
+                SortBy: 'PlayCount,SortName',
                 SortOrder: 'Descending',
                 UserId: apiClientProps.server?.userId,
             },


jeffvli:
The existing behavior is actually the *intended* behavior, because for Navidrome and other OpenSubsonic servers, the top songs are fetched based on last.fm popularity orders. From what I can tell, Jellyfin also supports this in some capacity but I'm not sure where the data is fetched from.

The actual way to resolve this would be to add a new parameter to the getTopSongs controller to swap between a `playCount` order and `community` order, which would need to be implemented for all 3 server controllers.

jeremylongshore:
Thank you for explaining, @jeffvli. That makes sense - I didn't realize the top songs were intentionally using last.fm popularity ordering.

I'd be glad to implement the approach you described (adding a parameter to swap between playCount and community order across all three server controllers) if that would be helpful. Just let me know if you'd like me to take that on, or if you'd prefer to close this PR.

Thanks again for taking the time to explain the design.

jeffvli:
Implemented in https://github.com/jeffvli/feishin/commit/f56a836ffd8a2a2117495a864ede9db1a8d5784d

PR #1531, opened 2026-01-08
fix: Chinese locale missing *_one key

haha4ni:
## Problem
When using the Chinese interface I was confused why many important words were still shown in English. I found that the entity.*_one keys were missing.

## Solution
In Chinese, the concept of pluralization is relatively weak, and the current Chinese translations actually all use singular forms, so I added "*_one" set as-is to fill in the missing entries.


kgarner7:
Thanks!

PR #1540, opened 2026-01-09
Improve file container display

pyxfluff:
This PR changes the file type display on the fullscreen player to remove the AUDIO/ prefix (present sometimes) and adds a rounded bitrate to it. If the server does not provide bitrate, it is simply not displayed (falls back to just the file type).

Before | after
image

### DIFF
--- a/src/renderer/features/player/components/full-screen-player-image.tsx
+++ b/src/renderer/features/player/components/full-screen-player-image.tsx
@@ -211,17 +211,12 @@ export const FullScreenPlayerImage = () => {
                     ))}
                 </Text>
                 <Group justify="center" mt="sm">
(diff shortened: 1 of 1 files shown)


jeffvli:
I'm not sure I like this as much as it is currently. Do most people care about the FLAC bitrate?

pyxfluff:
> I'm not sure I like this as much as it is currently. Do most people care about the FLAC bitrate?

I know I only speak for myself but I used to enjoy seeing it and still do, don't know how other people feel but it's been quite nice for me considering the area isn't too cluttered visually (especially once we remove audio/)

I could make it an option but that seems redundant to me and a waste of player settings space

PR #1542, opened 2026-01-10
Enable Playlist in sidebar to be sorted

Der-Penz:
This PR adds the ability to be able to drag and drop the playlists in the sidebar to change there order. Playlist order is saved in localstorage to be able to persist the order of the playlists.


jeffvli:
I like the idea, but I think this would also need a setting toggle to enable/disable the feature in case the user wants to revert back to the original server-provided name ascending order.

Der-Penz:
Yeah good idea, i will try to implement this

src/renderer/features/sidebar/components/sidebar-playlist-list.tsx
```diff
@@ -39,199 +43,261 @@ import {
 import { DragOperation, DragTarget } from '/@/shared/types/drag-and-drop';
 import { Play } from '/@/shared/types/types';
 
+const getPlaylistOrderKey = (serverId: string | undefined, scope: 'owned' | 'shared') => {
+    const sid = serverId || 'local';
+    return `playlistOrder:${sid}:${scope}`;
```
jeffvli: Change the localstorage key to snake case to match other localstorage items (some of my existing keys are inconsistent using kebab-case, but I don't want to add a new camel case version).

src/renderer/features/sidebar/components/sidebar-playlist-list.tsx
```diff
@@ -362,6 +433,79 @@ export const SidebarPlaylistList = () => {
...
+        const remaining = items.filter((it) => !saved.includes(it.id));
+        const new_order = [...ordered, ...remaining];
+
+        setDisplayItems(new_order);
+        saveOrderForServer(
+            server?.id,
+            new_order.map((playlist) => playlist.id),
+            'owned',
+        );
+    }, [memoizedItemData.items, server?.id, sidebarPlaylistSorting]);
```
jeffvli: You do not need a new state item and a useEffect here.

The playlist order is derived directly from the playlist query and the memoizedItemData, so you can move the additional sorting behavior in the memoizedItemData function itself.

As long as the useMemo dependency array includes the custom sort order, it will properly rerender on any changes. For example, you use the `loadSavedOrderForServer` function currently which is synchronous. You might want to use the `useLocalStorage` hook to tie the state to the render cycle.

src/renderer/features/sidebar/components/sidebar-playlist-list.tsx
```diff
@@ -481,6 +631,79 @@ export const SidebarSharedPlaylistList = () => {
...
+        const remaining = items.filter((it) => !saved.includes(it.id));
+        const new_order = [...ordered, ...remaining];
+
+        setDisplayItems(new_order);
+        saveOrderForServer(
+            server?.id,
+            new_order.map((playlist) => playlist.id),
+            'shared',
+        );
+    }, [memoizedItemData.items, server?.id, sidebarPlaylistSorting]);
```
jeffvli: Same as above

PR #1544, opened 2026-01-11
fix: Use .ico file on Windows

michaelevy:
Windows prefers .ico files to .pngs, so when using a .png as the icon, it gets compressed badly during the resize. This PR modifies the icon path to use the existing .ico file on Windows.

Note that there's an argument that macOS should be changed to use the .icns file for similar reasons, but I can't test the macOS so I haven't added that change.

### DIFF
--- a/electron-builder.yml
+++ b/electron-builder.yml
@@ -15,7 +15,7 @@ win:
     target:
         - zip
         - nsis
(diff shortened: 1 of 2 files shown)


jeffvli:
You'll need to replace the icon path in the electron-builder.yml file as well.

PR #1547, opened 2026-01-11
fix: Add missing _other strings in Polish translation

### LINKED ISSUE #1514: [Bug]: Some strings are still in English despite language being set to a diffrent one

- [x] Yes

### App Version

1.2.0

### Music Server and Version

Navidrome 0.59.0

Desktop Linux

### What happened?

Strings in the sidebar are still in English despite the app being set to Polish. Translation in weblate is at 99% so I don't think these are missing.
Image

### Steps to reproduce

1. Set language to (in this case) to Polish
2. Some strings mostly in the sidebar are still in English

### Relevant log output

```shell

```

skajmer:
Adds strings that were previously missing. These were clearly missing when looking at the sidebar.
Fixes: #1514 
Before:
before
After:
after


kgarner7:
Thanks!

PR #1563, opened 2026-01-16
Add album cover stacks to Artists and Album Artists views

alchemyyy:
Changes:

* Added `getCoverArtValidator` endpoint to Navidrome and Jellyfin controllers (Subsonic stub returns true for all)
* New `useCoverArtValidator` hook creates a closure that captures server-specific placeholder detection state
* New `useArtistAlbumStack` hook fetches artist albums and filters out placeholder covers

Settings:

* toggle to enable/disable artist album cover stack
* toggle to prefer artist cover (shows artist image if available, falls back to stack)
* stack size setting (1-10 albums)
* style selector (spun or staggered)
* fitment selector (underfit, fit, overfit) with overfit size option
* style-specific settings
	* spun: rotation angle (1-45 degrees)
	* staggered: horizontal and vertical offset values

* all sub-settings hidden when main stack toggle is disabled
* style-specific settings are shown only when that style is selected

Placeholder detection approach:

* Jellyfin: Simple HEAD request returning 200 (real image) or 404 (no image).
* Navidrome: I found a horrible way to do it. By fetching the cover art endpoint with no ID we are guaranteed to get the placeholder image. For whatever reason the placeholder images have their LastModified values set to the server startup time, which gives us something to compare against.
* Subsonic: unimplemented.

Why the closure pattern:

The validator fetches server-specific state (the placeholder timestamp) once when the page mounts, captures it in a closure, and that closure lives for the duration of the page. When the user navigates away the component unmounts and the closure dies with it.


alchemyyy:
I expect this to be pretty "raw" and in need of some work.

The styling options stuff was more or less me messing around with possibilities. 

I didn't look too deeply into the Subsonic API for a way to run placeholder detections. Running image hashing "dynamically" by creating a single request for the size of placeholder the page would be requesting was one way I thought of doing it, but I'm not sure if retrieving the placeholder like I am with Navidrome is even guaranteed.

jeffvli:
From my testing, the artist grid works well, but the artist grid carousel is blank.

camemb3rt:
Came here to suggest exactly this, looking forward to it!

PR #1566, opened 2026-01-17
Initial work: support showing studios for jellyfin, allow pill to be clickable

kgarner7:
Initial work for jellyfin record label ("studio") support.
Note that these are not picked up from tags, I had to add it manually to an album.


src/renderer/features/albums/components/tag-filter.tsx
```diff
@@ -0,0 +1,150 @@
...
+            limit={100}
+            onChange={handleChange}
+            searchable
+        />
+    );
+};
+
+TagFilterItem.displayName = 'TagFilterItem';
+
+export const TagFilters = () => {
```
jeffvli: Is this component supposed to be reused for the song filters components as well?

kgarner7: I didn't think of it at the time because I'm not sure if the same filters for Jellyfin would work for song. Maybe?

kgarner7: This reworked component is now shared for both album and track. And `boolTags` is removed entirely, with Jellyfin providing the multivalued tag.

PR #1589, opened 2026-01-21
feat: add regex filter setting for sidebar playlists

Exceen:
I have a lot of playlists but only few which I use regularly. For easier access I thought a sidebar regex filter option would be cool.

If you don't like it, just close this PR


jeffvli:
There are a couple of merge conflicts as a result of #1542.

If you don't mind fixing those, I'd be happy to merge this in.

Exceen:
nice! I solved the merge conflicts and tested the manual sorting in combination with the regex filter - works for me 👍🏻

PR #1608, opened 2026-01-23
Cover art support for folder view (Subsonic servers only)

marank:
As a follow-up to #1607 I poked around and created an implementation for displaying cover art in folder view if the Subsonic server sends coverArt in getIndexes or getMusicDirectory.


jeffvli:
Implemention looks fine, although your custom server might be the only one that actually supports folder images.

PR #1627, opened 2026-01-30
feat: Add support for player controls in macOS dock menu

steffenz:
Hi! 
I've added a feature I've been missing on macOS - namely support for player controls in the macOS dock (play/pause/next/previous/stop) 🤓  This is a feature Spotify and Tidal have (and likely many others too), and is something I think many Mac users (myself included) is used to or even expecting to see (I know, we are honestly the worst - I'm seriously considering switching to Linux!)

**Screenshots:**

Screenshot 2026-01-30 at 15 23 49
Screenshot 2026-01-30 at 15 24 32

**Technical considerations**
I exposed the ``use-mpris.ts`` hook to macOS (previously Linux only). From my limited knowledge and understanding of the codebase it could appear that the MPRIS name is a bit misleading in this hook - but I just "slapped" on a simple ``isMacOs()`` check that hopefully won't create problems on other platforms. The TypeScript compiler also complained about a missing ``sortName`` property which I fixed by adding ``sortName: title`` as a separate commit that can be omitted. 

**UX considerations for adding this**
Although the "tray" feature offers similar functionality this is in my opinion oftentimes a poor replacement for the more native experience that is having a menu on the Dock-icon (at least for more seasoned users). Lately a lot of apps adds unnecessary icons to the menu bar which I assume is because a lot of them is Electron-based and this is the default cross-platform behavior. Having both options however is trivial, as the tray functionality can be turned off and the dock menu isn't really intrusive. However it might lead to _slightly_ more code to maintain - but the code could eventually be shared with the tray variant to minimize this.

**Closing words**
Forgive me if my PR structure is a bit rough in the edges here. I couldn't really find a contribution guide, so I just did my best with previous knowledge. Besides that I just wanted to say I really love this application and the great work that's been put into it so far. After boycotting Spotify and switching to Tidal the user experience have been rough. Feishin however is punching way above it's class here (and I'd love to contribute to make it even better). 

In the future, adding something like Repeat and Shuffle might be something to consider - but for me personally that are actions I use less frequently and don't mind open the app to adjust.

### DIFF
--- a/src/main/features/darwin/dock-menu.ts
+++ b/src/main/features/darwin/dock-menu.ts
@@ -1,0 +1,57 @@
+import { app, ipcMain, Menu } from 'electron';
+
+import { getMainWindow } from '/@/main/index';
+import { PlayerStatus } from '/@/shared/types/types';
+
+let currentStatus: PlayerStatus = PlayerStatus.PAUSED;
+
+const updateDockMenu = () => {
+    if (!app.dock) return;
+
+    const isPlaying = currentStatus === PlayerStatus.PLAYING;
+
+    const dockMenu = Menu.buildFromTemplate([
+        {
+            click: () => {
+                getMainWindow()?.webContents.send('renderer-player-play-pause');
+            },
+            label: isPlaying ? 'Pause' : 'Play',
+        },
+        {
+            type: 'separator',
+        },
+        {
+            click: () => {
+                getMainWindow()?.webContents.send('renderer-player-next');
+            },
+            label: 'Next',
+        },
+        {
+            click: () => {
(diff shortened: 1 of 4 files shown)


jeffvli:
> From my limited knowledge and understanding of the codebase it could appear that the MPRIS name is a bit misleading in this hook

It is indeed. Originally, the main process only used these callback handlers for Linux MPRIS functionality as it wasn't needed elsewhere.

> Forgive me if my PR structure is a bit rough in the edges here. I couldn't really find a contribution guide, so I just did my best with previous knowledge.

Maybe someday I will add a contribution guide, but generally as long as the code is logically sound and follows relatively closely to the current app structure it's fine.

Looks good btw, thanks!

PR #1637, opened 2026-02-02
Prevent double fetching when force refreshing paginated views

eramdam:
This fixes a bug I recently noticed when refreshing the Albums view with the random sort, it seems the `refresh` method in `item-list-paginated-loader.ts` triggers two network requests instead of one. This is usually fine[^1] in views where two requests with the same arguments will return the same result but this is visible on lists with a random sort, as shown in this recording I took earlier today.

https://github.com/user-attachments/assets/4a1b0255-9c50-4bbf-a0ee-e02d55afef43

This PR fixes that by not calling `invalidateQueries()` when passing `force` to the `refresh` method, which results in only one network request as expected. My react-query chops are a bit rusty so perhaps this isn't how you're supposed to fix that issue, happy to adjust if necessary.

[^1]: Albeit not optimal because the first request still happens and isn't being cancelled like it probably should.


eramdam:
Seems like that behavior might have been there as long as this code, it wasn't as noticeable in 1.2.0 for example (flashes more quickly) but recent updates made the render slower which makes the issue more noticeable 

https://github.com/user-attachments/assets/32c734d0-9f5f-449d-b427-0036db4a6a45

eramdam:
I opened a PR to fix the performance regression that made that bug more apparent in [this PR](https://github.com/jeffvli/feishin/pull/1638)

jeffvli:
I believe the two fetches you're noticing is actually:
1. The listCount query
2. The actual page query

It's more apparent if you're not on page 1 of the list, as the listCount query is always run on the first page.

eramdam:
> I believe the two fetches you're noticing is actually:
> 
>     1. The listCount query
> 
>     2. The actual page query
> 
> 
> It's more apparent if you're not on page 1 of the list, as the listCount query is always run on the first page.

Ah yes indeed. Still, it feels like the listCount query shouldn't refresh the UI in the case of a `sortBy: random` query?

jeffvli:
Yep, I needed to add a condition to exclude this behavior specifically for random https://github.com/jeffvli/feishin/pull/1637/changes/0af5ff6ecdcc56e58eeca591b95aeeb7f276222a. The function was setting page data based on the list count fetch to improve the initial load time of the list which usually requires sequential fetches for the count and then the actual data.

PR #1638, opened 2026-02-02
Use a re-usable Intl.Collator instance for locale compare when possible

eramdam:
After opening https://github.com/jeffvli/feishin/pull/1637, I tried to run a `git bisect` to see when and the UI seemed slower since ~1.3.0. Bisect lead me to [this commit that added natural sorting](https://github.com/jeffvli/feishin/commit/45df497ee6d9efa3419ff09808c6250f09053981). Unfortunately, using `localeCompare` on a lot of items in an array isn't free because the JS engine re-creates a new instance of `Intl.Collator` for every item.

This PR fixes that by setting up a re-usable instance of `Intl.Collator()` which yields a [~84% speed improvement](https://jsperf.app/fekera/1/preview) (when sorting 1,000 items).

### DIFF
--- a/src/renderer/api/jellyfin/jellyfin-controller.ts
+++ b/src/renderer/api/jellyfin/jellyfin-controller.ts
@@ -39,7 +39,8 @@ const formatCommaDelimitedString = (value: string[]) => {
 const MAX_ITEMS_PER_PLAYLIST_ADD = 50;
 
 // Defining a re-usable Collator instance for performance reasons.
-const collator = new Intl.Collator(undefined, { numeric: true });
+const numericSortCollator = new Intl.Collator(undefined, { numeric: true });
+const collator = new Intl.Collator();
 
 const VERSION_INFO: VersionInfo = [
     [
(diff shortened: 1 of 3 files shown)


src/renderer/api/jellyfin/jellyfin-controller.ts
```diff
@@ -38,6 +38,10 @@ const formatCommaDelimitedString = (value: string[]) => {
 // not the POST body
 const MAX_ITEMS_PER_PLAYLIST_ADD = 50;
 
+// Defining a re-usable Collator instance for performance reasons.
+const numericSortCollator = new Intl.Collator(undefined, { numeric: true });
+const collator = new Intl.Collator();
```
eramdam, line 43: We could put these in a shared file but I wasn't sure of where to put them.

jeffvli: That location is fine!

jeffvli:
Good find on this! I don't think my library was large enough to notice a meaningful difference when this change was added, but the performance benchmark speaks for itself.

PR #1641, opened 2026-02-03
fix(Settings): mpv path selector displays wrong default

rushiiMachine:
The MPV path setting has a broken default value even though a separate `useEffect` hook is supposed to properly load the setting value later.

image


jeffvli:
Did this change resolve the error for you?

I did notice an incorrect path being set as well, which resulted in mpv erroring out in some cases.
Other than that, I'm not noticing playback breaking on *specific* albums.

rushiiMachine:
No, unfortunately it did not. This bug only affects viewing the MPV path from the UI.

Issue #1425, opened 2025-12-20
[Bug]: AudioMuse-AI Instant Mix not working


dxspotxfinsxnity:
### I have already checked through the existing bug reports and found no duplicates

- [x] Yes

### App Version

0.21.0, 0.22.0

### Music Server and Version

Jellyfin 10.11.5

### What local environments are you seeing the problem on?

Desktop Windows

### What happened?

Feishin seems to use its default Similar songs behavior instead of the one provided by the AudioMuse-AI Plug-In

It might be either Feishin, AudioMuse-AI, or Jellyfin bug. I'm leaving it there because Jellyfin and AudioMuse work but it's unknown whether the API is broken from this or their side.

### Steps to reproduce

1. Setup AudioMuse-AI using Docker Compose and run it through a set music library (Analyze and Cluster)
2. Observe Instant mixes working in the Jellyfin interface correctly with corresponding logs in the AudioMuse-AI container. Playlist generated has similar songs by acoustic content
3. Try Similar songs in the Feishin interface on the same song and have no logs in the AudioMuse-AI while also having unsimilar tracks in the created playlist. It relies on Artist/Genre instead

### Relevant log output

```shell

```

dxspotxfinsxnity:
I should've searched more about this bug because I see it mentioned in the #1214

kgarner7:
In your server settings (server > context menu > manage servers > your jellyfin server), make sure "Prefer instant mix" is selected. If the instant mix endpoint doesn't provide any data, that is an upstream problem, although it appears to be returning values for me.

A-wels:
Sorry for commenting on a closed issue, but this does not seem to work for me. I readded my Jellyfin server and made sure that "prefer instant mix" is selected. Starting an instant mix in Jellyfin also queries audiomuse.

Which functionality in feishin makes use of the "prefer instant mix" feature? I would guess the track radio and auto dj. Neither result in audiomuse queries for me. I feel like I am missing something here.

PR #1642, opened 2026-02-03
Update instant mix to use the new items endpoint instead of songs

A-wels:
The `items` endpoint can generate an instant playlist from any media item, including albums, artists, genres, playlists, or individual tracks, when provided with an ID.
(See: https://api.jellyfin.org/#tag/InstantMix/operation/GetInstantMixFromItem )

Using this endpoint offers two advantages:

1. It enables broader “Radio” functionality. While it would need to be implemented in the client, the endpoint already supports instant mixes for albums, genres, and playlists in addition to songs and artists.

2. It ensures compatibility with audiomuse-ai. For users without the audiomuse plugin, nothing changes, since Jellyfin internally translates the `items` endpoint to the existing endpoints such as `/song`. When audiomuse is installed, the overridden endpoint is used, which resolves:
   https://github.com/jeffvli/feishin/issues/1425
   https://github.com/NeptuneHub/audiomuse-ai-plugin/issues/14

### DIFF
--- a/src/renderer/api/jellyfin/jellyfin-api.ts
+++ b/src/renderer/api/jellyfin/jellyfin-api.ts
@@ -137,7 +137,7 @@ export const contract = c.router({
     },
     getInstantMix: {
         method: 'GET',
-        path: 'songs/:itemId/InstantMix',
+        path: 'items/:itemId/InstantMix',
         query: jfType._parameters.similarSongs,
         responses: {
             200: jfType._response.songList,


jeffvli:
Is this a new endpoint?

If so, it would need to be able to fallback to the original endpoint for older Jellyfin versions.

A-wels:
> Is this a new endpoint?
> 
> If so, it would need to be able to fallback to the original endpoint for older Jellyfin versions.

What is the oldest supported version?

The endpoint has been in Jellyfin for at least five years:
https://github.com/jellyfin/jellyfin/blob/9ac9543ee2b69f423f0102f80671fb09893534ae/Jellyfin.Api/Controllers/InstantMixController.cs

If necessary, I'll do a deeper dive to find when it was added.

jeffvli:
> The endpoint has been in Jellyfin for at least five years:

No need if this is the case. I doubt any users are on versions older than that.

PR #1655, opened 2026-02-06
Open settings with shortcut

martgnz:
This change adds a shortcut to open to settings menu by pressing <kbd>Command+,</kbd> on macOS and <kbd>Ctrl+,</kbd> on Windows or Linux. This is a common [convention](https://support.apple.com/en-us/102650#copy:~:text=Command%2DComma%20%28%2C%29%3A%20Open%20settings%20%28preferences%29%20for%20the%20front%20app).

It also adds the option to the global application menubar:

imagen

### DIFF
--- a/src/preload/mpv-player.ts
+++ b/src/preload/mpv-player.ts
@@ -174,6 +174,10 @@ const rendererPlayerFallback = (cb: (event: IpcRendererEvent, data: boolean) => 
     ipcRenderer.on('renderer-player-fallback', cb);
 };
 
+const rendererOpenSettings = (cb: (event: IpcRendererEvent) => void) => {
+    ipcRenderer.on('renderer-open-settings', cb);
+};
+
 export const mpvPlayer = {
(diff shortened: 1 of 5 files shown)


src/preload/mpv-player.ts
```diff
@@ -174,6 +174,10 @@ const rendererPlayerFallback = (cb: (event: IpcRendererEvent, data: boolean) =>
     ipcRenderer.on('renderer-player-fallback', cb);
 };
 
+const rendererOpenSettings = (cb: (event: IpcRendererEvent) => void) => {
+    ipcRenderer.on('renderer-open-settings', cb);
+};
```
jeffvli: It doesn't make sense to add this to the mpvPlayer. Add it to the utils preload instead

martgnz: Thanks for taking a look. I've moved it to utils preload and also changed it so it gets called in app.tsx.

PR #1664, opened 2026-02-07
Move settings panel to native menu bar

martgnz:
Works off https://github.com/jeffvli/feishin/pull/1655 to move all settings to the native menu bar.

- Moved settings, manage servers and private mode to main app menu:

imagen

<p>

- It also integrates checkbox state, so for example, if you toggle private mode the menu bar gets checked
imagen

<p>

- Moved command palette icon and collapse sidebar to "View" menu option:
imagen

<p>

- The same thing happens when collapse sidebar is activated:

imagen

<p>

- Moved version to "Help" menu option:
imagen

<p>

- This means that we can hide the settings icon from sidebar, both on normal and collapsed states:
imagen

imagen

### DIFF
--- a/src/renderer/app.tsx
+++ b/src/renderer/app.tsx
@@ -1,5 +1,6 @@
 /* eslint-disable perfectionist/sort-imports */
 import { MantineProvider } from '@mantine/core';
+import { openModal } from '@mantine/modals';
 import { Notifications } from '@mantine/notifications';
 import 'overlayscrollbars/overlayscrollbars.css';
 import '/styles/overlayscrollbars.css';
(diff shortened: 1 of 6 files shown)


src/renderer/app.tsx
```diff
@@ -79,6 +96,109 @@ export const App = () => {
...
+            });
+
+            return () => {
+                ipc?.removeAllListeners('renderer-open-release-notes');
+            };
+        }
+
+        return () => {};
+    }, [t]);
+
```
jeffvli, line 201: Can you refactor this to move this into a hook? I don't want this custom functionality muddying up the App.tsx entrypoint.

src/renderer/features/sidebar/components/collapsed-sidebar.tsx
```diff
@@ -99,23 +100,25 @@ export const CollapsedSidebar = () => {
...
+                                    cursor: 'pointer',
+                                    padding: 'var(--theme-spacing-md) 0',
+                                }}
+                            />
+                        </DropdownMenu.Target>
+                        <DropdownMenu.Dropdown>
+                            <AppMenu />
+                        </DropdownMenu.Dropdown>
+                    </DropdownMenu>
+                )}
```
jeffvli, line 121: This should not be hidden. The native settings panel is only visible on macOS, so with this change you're hiding it from all Windows and Linux users as well.

src/renderer/features/sidebar/components/action-bar.tsx
```diff
@@ -39,16 +40,18 @@ export const ActionBar = () => {
...
+                                <DropdownMenu.Target>
+                                    <Button p="0">
+                                        <Icon icon="menu" size="lg" />
+                                    </Button>
+                                </DropdownMenu.Target>
+                                <DropdownMenu.Dropdown>
+                                    <AppMenu />
+                                </DropdownMenu.Dropdown>
+                            </DropdownMenu>
+                        )}
```
jeffvli, line 54: Same thing here. Should not be hidden.

PR #1668, opened 2026-02-07
MusicBrainz integration

jeffvli:
- Fetches artist and releases based on MusicBrainz Id and displays them as external sources in the Artist page 
- Includes youtube search API to attempt to play external sources as preview
- Adds new `Integrations` settings tab to handle MusicBrainz integration
image
image


russssl:
Maybe Lidarr integration would be beneficial ? It would allow to expand the library easier

PR #1669, opened 2026-02-08
fix: remove duplicate CommandPalette in mobile layout

### LINKED ISSUE #1666: [Bug]: Mobile view: pressing CTRL+K opens two command menus and types into the background one

- [x] Yes

### App Version

1.4.2

### Music Server and Version

subsonci (gonic)

Desktop Windows

### What happened?

When using Feishin in mobile view (small window), pressing CTRL + K causes a UI issue:

Two command menus open at the same time

Typing is entered into the background command menu rather than the one in the foreground

This results in overlapping UI elements and makes the command palette difficult to use.

Expected behavior:
Only one command palette should open, and input should be focused on the visible one.

Screenshot:

Image

### Steps to reproduce

1. Resize Feishin window to mobile layout
2. Press CTRL + K
3. Observe that two overlapping menus appear
4. Start typing, text goes into the background menu

### Relevant log output

```shell

```

kaigritun:
## Summary

Fixes #1666

The `CommandPalette` component was being rendered twice when in mobile view:
1. In `ResponsiveLayout` via `LayoutHotkeys` component (which provides CommandPalette for all layouts)
2. In `MobileLayout` directly

This caused two overlapping command menus to open when pressing Ctrl+K in mobile view, with keyboard input going to the background menu instead of the visible foreground one.

## Solution

Remove the duplicate `CommandPalette` from `MobileLayout` since `LayoutHotkeys` (in `ResponsiveLayout`) already provides it for all layouts - both desktop and mobile.

Note: `DefaultLayout` never had its own `CommandPalette` and relied on `LayoutHotkeys`, so this brings `MobileLayout` in line with that pattern.

## Testing

1. Resize Feishin window to mobile layout
2. Press Ctrl+K
3. Verify only one command palette opens
4. Verify typing goes into the visible menu

### DIFF
--- a/src/renderer/layouts/mobile-layout/mobile-layout.tsx
+++ b/src/renderer/layouts/mobile-layout/mobile-layout.tsx
@@ -8,11 +8,10 @@ import styles from './mobile-layout.module.css';
 import { ContextMenuController } from '/@/renderer/features/context-menu/context-menu-controller';
 import { FullScreenVisualizer } from '/@/renderer/features/player/components/full-screen-visualizer';
 import { MobileFullscreenPlayer } from '/@/renderer/features/player/components/mobile-fullscreen-player';
-import { CommandPalette } from '/@/renderer/features/search/components/command-palette';
 import { MobileSidebar } from '/@/renderer/features/sidebar/components/mobile-sidebar';
 import { PlayerBar } from '/@/renderer/layouts/default-layout/player-bar';
 import { useFullScreenPlayerStore } from '/@/renderer/store';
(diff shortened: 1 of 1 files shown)


MichaelDeBoey:
I'm not making any claims about the quality of their work, but I wanted to let you know that @kaigritun is a fully-autonomous non-human actor
https://socket.dev/blog/ai-agent-lands-prs-in-major-oss-projects-targets-maintainers-via-cold-outreach

jeffvli:
@MichaelDeBoey Interesting! Thanks for letting me know.

PR #1681, opened 2026-02-10
Add album detail list view

### LINKED ISSUE #1661: [Feature]: Tracks/playlists view grouping tracks from the same album while showing album art only once

- [x] Yes

Not server-specific

Feishin is amazing but there's one thing I am sorely missing from iTunes/Apple Music: the ability to organize my tracks/playlist view in a way that **groups tracks from the same album, without showing the _same_ album art** over and over.

Here's an example from my collection:

Image

jeffvli:
image
image


jeffvli:
Related to #1661

Currently only available on the Album list. Needs additional refactoring to support album view on playlist view.

PR #1688, opened 2026-02-10
Regenerate macOS icon (.icns) to fix glitched small icons

### LINKED ISSUE #1152: [Bug]: Icon glitched on macOS

- [x] Yes

### App Version

0.20.1

### Music Server and Version

N/A

Desktop macOS

### What happened?

The icon appears glitched out. Only happens for smaller versions of the icon, doesn't happen in the dock or in the detailed view of the Finder.

Image
Image

### Steps to reproduce

1. Download either ` Feishin-0.20.1-mac-arm64.dmg ` or `
Feishin-0.20.1-mac-arm64.zip `
2. Mount or extract the contents
3. Look at the Feishin icon 🤓

### Relevant log output

```shell

```

york9675:
Regenerated icon.icns using macOS native iconutil with properly structured iconset. The original icns used legacy icon types (icp4, icp5, icp6) which render incorrectly at small sizes on modern macOS.

Also normalized DPI metadata on source PNGs from 96 to 72 DPI (macOS standard). Pixel data is unchanged.

Fixes #1152


jeffvli:
Thanks!

PR #1692, opened 2026-02-10
fix: use media session by default on mac and hide setting to disable

fogrye:
Following up the discussion in discord, there is no reason for mac to have media session (Now playing) disabled. This change enables it by default and hides setting to disable.

### DIFF
--- a/src/main/features/core/player/media-keys.ts
+++ b/src/main/features/core/player/media-keys.ts
@@ -1,30 +1,11 @@
-import { BrowserWindow, globalShortcut, systemPreferences } from 'electron';
+import { BrowserWindow, globalShortcut } from 'electron';
 
-import { isLinux, isMacOS } from '../../../utils';
+import { isLinux } from '../../../utils';
 import { store } from '../settings';
 
 import { PlayerType } from '/@/shared/types/types';
 
 export const enableMediaKeys = (window: BrowserWindow | null) => {
-    if (isMacOS()) {
-        const shouldPrompt = store.get('should_prompt_accessibility', true) as boolean;
-        const shownWarning = store.get('shown_accessibility_warning', false) as boolean;
-        const trusted = systemPreferences.isTrustedAccessibilityClient(shouldPrompt);
-
-        if (shouldPrompt) {
-            store.set('should_prompt_accessibility', false);
-        }
-
-        if (!trusted && !shownWarning) {
-            window?.webContents.send('toast-from-main', {
-                message:
-                    'Feishin is not a trusted accessibility client. Media keys will not work until this setting is changed',
-                type: 'warning',
-            });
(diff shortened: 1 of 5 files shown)


src/main/features/core/player/media-keys.ts
```diff
@@ -1,30 +1,11 @@
...
-        if (!trusted && !shownWarning) {
-            window?.webContents.send('toast-from-main', {
-                message:
-                    'Feishin is not a trusted accessibility client. Media keys will not work until this setting is changed',
-                type: 'warning',
-            });
-            store.set('shown_accessibility_warning', true);
-        }
-    }
-
```
jeffvli, line 27: After thinking about it some more, macOS should probably remain with mediasession disabled by default because the user *explicitly* needs to accept the prompt for trusted accessibility before it can even work.

If the setting is enabled by default, then it would be confusing because the toggle would show that it's enabled, when in reality it's being blocked by the OS.

fogrye: let me explain, app should not bind to keys and in case of macos with media session enabled it won't bind based on [condition on line 12](https://github.com/fogrye/feishin/blob/f1a97a2bbd74fbc83b6e829d787695861ec9b4bd/src/main/features/core/player/media-keys.ts#L12) same as no other mac app do. When you enable mediasession you give mac native "now playing" functionality control over playback so media keys would work same as headset keys do. Accessibility feature will not be called once and app will function just fine.

jeffvli: > Accessibility feature will not be called once and app will function just fine.

I don't understand this? From my understanding, macOS blocks *all* media keys from the app unless explicitly allowed by the user under accessibility.

The reason to keep it disabled by default behind the user toggle is so that the user knows and understands why their media hotkeys are not working.

Exceen: > I don't understand this? From my understanding, macOS blocks all media keys from the app unless explicitly allowed by the user under accessibility.

This is correct. There should be a way to prompt the user for this though. A lot of apps prompt the user when the app is launched, but Feishin does not trigger this prompt.

I think having this option set to enabled by default is good, because this is what I'd expect from a Music player on macOS, BUT someone needs to figure out how to prompt the user.

Also in the past I had issues with the accessibility permissions after switching builds of Feishin and macOS has difficulty detecting this sometimes. But maybe that was just me and it actually works fine anyway.

fogrye: you don't have to prompt user if you don't bind media playback keys. you don't have to bind keys since OS will handle them on it's own same to other music players. If we agree that enabled by default is good on macos I see no point to have option to disable it since it makes things worse: you have to bind to media keys and give app accessibility permissions which it doesn't need and lose convenience of headset media control

Issue #1494, opened 2026-01-04
[Feature]: Recently played on the homepage also displays recently played song (Jellyfin)


SpongeRoss:
### I have already checked through the existing feature requests and found no duplicates

- [x] Yes

### Is this a server-specific feature?

Jellyfin

### What do you want to be added?

Related to issue #926 It would be nice to have the _recently played_ section on the Homepage display the recently played songs for Jellyfin users, similar to Sonixd.

PR #1697, opened 2026-02-11
fix: actually show Jellyfin recently played songs carousel

yoshw:
Follow-up to #1494

Support for showing a Jellyfin recently-played songs carousel was added in a8604dd150c421ba32a7237015b1b1d0af744c33 (thanks!), but the carousel wouldn't show up when enabled, due to this filter.

### Testing

Ran the app locally, and confirmed I could see the "Recently played" carousel as expected.


jeffvli:
... I thought I removed that condition. Thanks for the fix!

PR #1700, opened 2026-02-12
Add album view for playlists

### LINKED ISSUE #1574: [Feature]: (Smart) Playlist album view

- [x] Yes

Not server-specific

Album view for (smart) playlists.

It is already possible to see all tracks in a grid view as repeated album images but it would be great to have them combined. This way you can create a smart playlist view containing the newest albumst from a genre or artist. The Marvis app on iOS has the same option and makes it very flexible.

jeffvli:
- Adds a toggle to switch between track / album view in playlist song list
- Moves smart playlist query editor to the bottom
- Update client-side song filtering to include album/discnumber/tracknumber for most filters
  - properly uses the actual sort order for discnumber/tracknumber so that it follows the user-selected order instead of asc/asc
- Fixes the Album detail view to decouple the list configuration between different list keys

closes #1574


pjsurk:
Where can I find this toggle to switch between track / album view in playlist song list?

PR #1701, opened 2026-02-12
add explicit blurring to left expanded image and full screen

kgarner7:
Adds blurs to expanded left image and full screen player. This is not added to the lower detail popup on the album page, not sure if that should also be included.

### DIFF
--- a/src/renderer/features/player/components/full-screen-player-image.module.css
+++ b/src/renderer/features/player/components/full-screen-player-image.module.css
@@ -6,6 +6,10 @@
     filter: drop-shadow(0 0 5px rgb(0 0 0 / 40%)) drop-shadow(0 0 5px rgb(0 0 0 / 40%));
 }
 
+.censored.image {
+    filter: blur(30px);
+}
+
 .image-container {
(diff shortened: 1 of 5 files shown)


src/renderer/features/player/components/full-screen-player-image.module.css
```diff
@@ -6,6 +6,10 @@
     filter: drop-shadow(0 0 5px rgb(0 0 0 / 40%)) drop-shadow(0 0 5px rgb(0 0 0 / 40%));
 }
 
+.censored.image {
+    filter: blur(30px);
+}
```
jeffvli, line 11: The blur for the Image component (image.module.css) uses a blur of 10px. Is that not enough for the fullscreen player image? Just wondering for styling consistency purposes, since you set the SidebarImage to use 20px as well so now there are 3 separate blur configurations.

kgarner7: Not in my opinion anyway. Since it's a large image, the 10px blur isn't as strong (same with left-controls).

PR #1703, opened 2026-02-12
fix: restore original macOS squircle icon and regenerate icns properly

york9675:
The previous fix (#1688) incorrectly regenerated the icns from circular PNG icons, replacing the macOS-specific squircle design. 

- Restores the original squircle icon design from the pre-#1688 icns
- Regenerates icns using macOS native iconutil with proper modern icon types (ic04, ic05 instead of legacy icp4, icp5, icp6)
- Fixes `@2x` retina DPI metadata (144 DPI instead of 72)
- Reverts unnecessary DPI changes to source PNGs (Windows/Linux icons)

I really apologize for this oversight and the need for a follow-up fix

### App launcher
image 

### Finder list/column view
image
image
image

### Spotlight/Raycast
image
image

### DMG itself
image


jeffvli:
No problem, I appreciate you testing on macOS and implementing changes since I don't have one

PR #1705, opened 2026-02-13
Regenerate small icon entries in icns to fix glitch on non-retina displays

york9675:
Follow-up to #1703.

# Problem
The fix in #1703 restored the correct squircle icon design and regenerated the .icns using iconutil, but the small icon entries (16x16, 32x32) extracted from the original .icns were already corrupted — they contained garbled pixel data. This wasn't caught because in #1703 I only tested on a MacBook's built-in Retina display, where macOS uses the larger `@2x` entries which were fine. On a 1080p (non-Retina) external display, macOS uses the 1x small entries, which still showed the glitch.

# Fix
Regenerated all icon sizes by downscaling from the known-good 1024x1024 squircle source using Lanczos resampling, instead of reusing the corrupted small entries from the original .icns.

# Testing
Verified on:
- External 1080p non-Retina display
- MacBook built-in Retina display
- Finder list/column view, Spotlight, app launcher, DMG

Again, I sincerely apologize for multiple rounds of fixes on this — should have tested on a non-Retina display from the start... 🙏


jeffvli:
There's a merge conflict with your icns file for some reason.

york9675:
I'm away from my desk right now, but I’ll take a look and try to resolve the conflict as soon as I get home. Appreciate the patience!

york9675:
> There's a merge conflict with your icns file for some reason.

conflict resolved and ready to merge!

Sorry again for the multiple rounds on this

jeffvli:
I think you accidentally removed the resized .png icons when you force pushed?

york9675:
> I think you accidentally removed the resized .png icons when you force pushed?

The .icns is the only file that had corrupted small icon entries causing the glitch on non-Retina displays. The PNGs (used for Windows/Linux) are unaffected, so only icon.icns shows in the diff, which is correct.

PR #1706, opened 2026-02-13
Fix sleep timer end-of-song mode not stopping web player audio when audio fade is disabled

york9675:
## Summary

Fixes sleep timer end-of-song mode not properly stopping audio playback when using the web player when "Audio fade on status change" is disabled. Follow-up to #1671.

## Problem

When "Audio fade on status change" is disabled in playback settings, the sleep timer's end-of-song mode had an issue with the web player: the UI correctly showed a paused state after the current song ended, but audio continued playing the next song in the background.

This did not occur with the MPV player. I didn't catch this while developing the sleep timer in #1671 because my dev environment uses the default settings (audio fade enabled), while my production build had this option disabled — so the bug only showed up there.

## Fix

Adds a `pauseOnNextSongEnd` flag to the player store. When sleep timer end-of-song mode is active, this flag makes `mediaAutoNext` return `PAUSED` status, and the web player's gapless/crossfade handlers skip pre-starting the next audio element.

### DIFF
--- a/src/renderer/features/player/audio-player/web-player.tsx
+++ b/src/renderer/features/player/audio-player/web-player.tsx
@@ -20,6 +20,7 @@ import {
     usePlayerData,
     usePlayerMuted,
     usePlayerProperties,
+    usePlayerStoreBase,
     usePlayerVolume,
 } from '/@/renderer/store';
 import { toast } from '/@/shared/components/toast/toast';
@@ -180,7 +181,15 @@ export function WebPlayer() {
 
         promise.then(() => {
             playerRef.current?.player1()?.ref?.getInternalPlayer().pause();
-            playerRef.current?.setVolume(volume);
+
+            // If mediaAutoNext resulted in a paused state (e.g. end of queue,
+            // or pauseOnNextSongEnd flag), stop all audio instead of restoring volume.
+            const currentStatus = usePlayerStoreBase.getState().player.status;
+            if (currentStatus === PlayerStatus.PAUSED) {
+                playerRef.current?.pause();
+            } else {
+                playerRef.current?.setVolume(volume);
+            }
             setIsTransitioning(false);
         });
     }, [mediaAutoNext, volume]);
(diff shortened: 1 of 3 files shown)


jeffvli:
> When running a production build (pnpm run start or the packaged executable), the sleep timer's end-of-song mode had an issue with the web player: the UI correctly showed a paused state after the current song ended, but audio continued playing the next song in the background.

I'm not seeing this issue in production or dev.

york9675:
Thanks for checking! I found the specific repro condition — the issue only occurs when **"Audio fade on status change" is disabled** in playback settings (it's enabled by default).

### Steps to reproduce
1. Go to Settings -> Playback -> disable "Audio fade on status change"
2. Set sleep timer to "End of current song"
3. Wait for the song to end

### What happens
The UI shows paused, but audio from the next song continues playing in the background.

### Why
When fade is enabled, `fadeAndSetStatus()` properly fades volume to 0 before pausing; when disabled, the `onPlayerStatus` handler sets volume to 0, but the `handleOnEnded` `.then()` callback immediately restores it via `setVolume(volume)`. The next player's audio element was already started by the gapless/crossfade pre-start (direct `.play()` DOM call), and since ReactPlayer's `playing` prop doesn't change for it, `.pause()` is never called.


By the way, this fix also prevents the next song from briefly playing audible audio before stopping (the gapless handler pre-starts the next player ~0.1s before the song ends, which is audible without this fix).

src/renderer/features/player/audio-player/web-player.tsx
```diff
@@ -249,6 +289,13 @@ export function WebPlayer() {
                     }
                 }
 
+                // Explicitly pause all internal audio elements when status becomes PAUSED
+                // This handles cases where gapless/crossfade started the next player
+                // via direct .play() calls, bypassing ReactPlayer's playing prop
+                if (status === PlayerStatus.PAUSED) {
+                    playerRef.current?.pause();
+                }
```
jeffvli: This breaks the audio fade out when audioFadeOnStatusChange is enabled. This would need to be moved inside the `else` block with if(status === PlayerStatus.Paused)

jeffvli:
I'm not a fan of putting business logic (the sleep timer) into the web-player or web-player-engine directly, as it quickly becomes unmaintainable if more things are added.

I'll try to think about this can be better managed.

york9675:
Hey @jeffvli, I've reworked the fix. Instead of handling the sleep timer inside the sleep timer component, this adds a `pauseOnNextSongEnd` flag to the player store. When sleep timer end-of-song mode is active, this flag makes `mediaAutoNext` return `PAUSED` status, and the web player's gapless/crossfade handlers skip pre-starting the next audio element.

PR #1719, opened 2026-02-15
Add AppAtream xml file to publishing

WhyKickAmooCow:
Add electron builder hook to update the AppStream metainfo file and add it to the list of files to be published for the linux build.

This will allow the file to be kept in sync at publish time, rather than having having to use npm scripts manually.


WhyKickAmooCow:
Example release https://github.com/WhyKickAmooCow/feishin/releases

PR #1731, opened 2026-02-17
Redo Icons in SVG

mihawk90:
Recreated as an SVG in Inkscape using the existing assets as a reference.

Also includes exports from SVG for all relevant files:
- `assets/icons/` in 16-1024px size PNGs
- `assets/icons/icon.ico` using same sizes as the old one
- `assets/icons/` IconTemplate
- `media/` black and white logo-only variants
- `resource/icon.png`

---

Couple notes:
- wasn't sure which of these are even still used, so I just did them all
- deleted the PDN files, let me know if you want to keep them and I'll revert it
- added the original Inkscape SVG which includes Inkscape's project tags like guides and reference images and stuff, but also a plain SVG export
- The `icon.ico` somehow became huge and I have no idea why. The individual PNGs are actually smaller than the ones extracted from the old `icon.ico`, but somehow it ended up at 4x the size? I used [`icoutils`](https://www.nongnu.org/icoutils/) to create it, maybe whatever the old one was created with does some kind of compression?
- I did *not* create a new `favicon.ico` for 2 reasons:
  - the old one seems to be corrupted in some way so I was unable to check what resolutions it includes
  - It might be worth using a [PNG](https://caniuse.com/link-icon-png) and/or [SVG](https://caniuse.com/link-icon-svg) favicon instead, see also [this article](https://dev.to/masakudamatsu/favicon-nightmare-how-to-maintain-sanity-3al7)
- I thought about deduplicating `assets/icons/icon.png` and `resources/icon.png` by using symlinks instead but since I don't know how they are used, I used regular files instead
- I didn't do a new Mac ICNS file because I don't have the tools or knowledge. It could be easily adapted from the SVG though if need be.


jeffvli:
> The icon.ico somehow became huge and I have no idea why. The individual PNGs are actually smaller than the ones extracted from the old icon.ico, but somehow it ended up at 4x the size? I used [icoutils](https://www.nongnu.org/icoutils/) to create it, maybe whatever the old one was created with does some kind of compression?

The original ico probably only contained a single resolution image, as I probably used an online converter for it.

> I thought about deduplicating assets/icons/icon.png and resources/icon.png by using symlinks instead but since I don't know how they are used, I used regular files instead

That's fine.

> I did not create a new favicon.ico for 2 reasons:
the old one seems to be corrupted in some way so I was unable to check what resolutions it includes
It might be worth using a [PNG](https://caniuse.com/link-icon-png) and/or [SVG](https://caniuse.com/link-icon-svg) favicon instead, see also [this article](https://dev.to/masakudamatsu/favicon-nightmare-how-to-maintain-sanity-3al7)

Can be addressed later. Favicon still works as-is for now, but is a low priority optimization if we want to change it.

> I didn't do a new Mac ICNS file because I don't have the tools or knowledge. It could be easily adapted from the SVG though if need be.

Not touching ICNS is fine, as it was already resolved by a macOS user a little while ago.

PR #1760, opened 2026-02-25
feat: add animated Apple Music cover artworks

### LINKED ISSUE #1156: [Feature]: Animated Cover Arts in Player & Album View

- [x] Yes

Not server-specific

I think Apple Music’s animated cover arts are extremely cool. They add motion and a modern aesthetic to the music experience.

For example:
👉 [Cypress Hill – IV][https://music.apple.com/de/album/dr-greenthumb/263570525?i=263570933]
There you can see the animation both in the album track list view and when clicking on the small cover art inside the player.

**Current Behavior in Feishin**
Feishin supports gifs 💃 
In the main player (left side) and the album track list view, however, only static artwork is shown.

**Proposal**
Main Player (left side): If animated artwork is available → display the animation instead of the static image.
Album Track List View: Also use the animated version, just like Apple Music does.
Fallback: If no animated artwork is available, fall back to the static cover art.

**Benefits**
Brings more of the Apple Music feel into Feishin 🎨
Strongly enhances the visual experience
Allows artists/labels to showcase their work in a more engaging way
Creates a more modern and immersive listening experience for users

**Source for Test Material**
The [Apple Music Artwork Finder](https://bendodson.com/projects/apple-music-artwork-finder/?utm_source=chatgpt.com) by Ben Dodson makes it possible to download Apple’s animated cover arts directly and use them for testing in Feishin.

**Nice-to-Have**
A toggle in Settings: Enable/disable animated cover art (useful for users who want to save performance or bandwidth).
Proper caching, so animations don’t need to be reloaded every time.
**Open Questions**

Would it be feasible to reuse the GIF display logic that already exists and apply it to the main player and album view?
Or would it be better to implement native video loop support (e.g. MP4/H.264) for smoother playback?
Should this feature be enabled by default, or opt-in via Settings?

This would bring Apple Music–style animated cover art directly into Feishin!

m8tec:
Introduces support for fetching and displaying animated album covers during playback and navigation. The artworks are dynamically fetched using the album and artist name via a public instance of https://github.com/m8tec/apple-music-animated-artworks at https://artwork.m8tec.top/

Core functionalities & Settings:
- Global toggle to enable/disable animated covers
- Granular toggles to enable/disable the feature per screen
- Configurable custom API base URL in settings (for self-hosting)
- Implemented a runtime IPC cache to prevent redundant API calls and improve performance
- Clear m3u8-cache button

Technical Implementation (HLS):
- Integrates `hls.js` to parse and stream Apple Music's `.m3u8` video manifests.
- Robust stream management including an initialization timeout, error detection, and instance cleanup on unmount.
- Fallback: Automatically renders the static artwork if the HLS video stream fails.

Supported Screens:
- Full screen player
- Mobile full screen player
- Mini player
- Album detail page
- Sidebar image

Closes #1156 

## Preview

https://github.com/user-attachments/assets/ce369bee-7ceb-4da3-8a12-66e45959c6b5

https://github.com/user-attachments/assets/bbe4399e-f83c-4d33-98e3-4bdd16beb7a0

https://github.com/user-attachments/assets/38326945-df7f-4c48-b12d-c6a6874fe09c


src/renderer/components/animated-video-cover/animated-video-cover.tsx
```diff
@@ -0,0 +1,177 @@
...
+                autoPlay={isPlaying}
+                className={className}
+                crossOrigin="anonymous"
+                loop
+                muted
+                onError={handleError}
+                playsInline
+                ref={ref}
+                {...props}
+            />
```
jeffvli: Is there a reason you're using the motion component here instead of the native video element?

src/renderer/features/settings/components/general/general-tab.tsx
```diff
@@ -26,6 +27,7 @@ export const GeneralTab = memo(() => {
             { component: SidebarSettings, key: 'sidebar' },
             { component: ScrobbleSettings, key: 'scrobble' },
             { component: LyricSettings, key: 'lyrics' },
+            { component: AnimatedCoverSettings, key: 'animatedCovers' },
```
jeffvli, line 30: The AnimatedCoverSettings should also be conditionally added only if `isElectrion()` is true, since this feature relies on fetching in the main process.

kgarner7:
I'll be honest, I'm not too keen integrating some _random_ service into this application just to fetch cover art.
But also, it feels extremely strange to move said process in the main thread. If you have proper CORS headers, this should be done from the renderer side.

jeffvli:
> I'll be honest, I'm not too keen integrating some random service into this application just to fetch cover art.

I had the same thoughts initially, but given that the fetching implementation is somewhat interfaced and and easily adapatible (if also using hls) to a new service (if needed) or in the future I don't see a significant problem to merging it.

jeffvli:
> But also, it feels extremely strange to move said process in the main thread. If you have proper CORS headers, this should be done from the renderer side.

This part is true. Can be moved to the renderer since the service is hosted by @m8tec, with support for changing the api URL for locally hosted instance as well.

kgarner7:
> > I'll be honest, I'm not too keen integrating some random service into this application just to fetch cover art.
> 
> I had the same thoughts initially, but given that the fetching implementation is somewhat interfaced and and easily adapatible (if also using hls) to a new service (if needed) or in the future I don't see a significant problem to merging it.

I understand that apple specifically provides HLS, but how many other services actually do that (as opposed to say, an animated webp/gif, or even just a video element)? This is _explicitly_ tied to how Apple does things, which I don't think would be as extensible to other services (it also assumes that Apple doesn't change its CORS headers in the future, although that might break other applications).

Anyway, that's all from me, do with this as you will.

src/renderer/components/animated-video-cover/animated-video-cover.tsx
m8tec: Good catch! There are no specific animations applied directly to the video element itself, so I reverted this to a native <video> tag.

src/renderer/features/settings/components/general/general-tab.tsx
m8tec: Since we moved the fetching logic entirely to the renderer (as discussed below), the feature now also works in the web version! Therefore, I kept the settings visible for all users instead of wrapping them in `isElectron()`.

m8tec:
> This part is true. Can be moved to the renderer since the service is hosted by @m8tec, with support for changing the api URL for locally hosted instance as well.

Agreed! I just pushed a commit that completely rips out the IPC boilerplate and moves the fetching directly into the renderer. To make the browser happy, I also updated the artwork backend to send the correct CORS headers.

The feature now fully works in the regular web version too. Because of that, we also don't have to hide the settings behind isElectron(). Thanks for the suggestion!

m8tec:
> I understand that apple specifically provides HLS, but how many other services actually do that (as opposed to say, an animated webp/gif, or even just a video element)? This is explicitly tied to how Apple does things, which I don't think would be as extensible to other services (it also assumes that Apple doesn't change its CORS headers in the future, although that might break other applications).

That’s a valid concern!

Regarding the Apple dependency: The good thing is that the UI won't actually break if Apple decides to lock down their CORS headers in the future. If the stream fails to load, the user won't even notice except for no animated artwork displayed.

About other formats: I actually had a chat with some of the Navidrome devs. They mentioned that they are open minded about the idea of supporting .mp4 video covers as part of a future UI update (or potentially via their new plugin system).

The current `<AnimatedVideoCover>` component is already somewhat prepared for this. It currently checks for .m3u8 to initialize hls.js, but it can easily be tweaked to just pass standard .mp4 or .webm URLs directly to the native <video> tag once servers like Navidrome start providing them natively. So hopefully, this lays some groundwork for the future!

m8tec:
We should probably also talk about what the default state should be:

Should animated covers be enabled or disabled out of the box? Since the feature makes external requests (even though they're heavily cached), I could see why we might want to make it opt-in. On the flip side, having it enabled by default means people actually get to see the new feature right away.

I don't really have a strong preference either way. Just let me know what makes the most sense for the project, and I'll flip the boolean if needed.

jeffvli:
Sorry for the delay on this, but I'm still unsure of whether or not this is appropriate to merge into main.

I've been thinking about it some more and I'm concerned about the maintenance aspect of this feature as it relies on your external service (as well as Apple upstream). It might make more sense to wait until I can implement some form of a plugin system so that it can implemented externally instead of internally.

m8tec:
No worries at all, that's understandable.
I had originally hoped that making the backend API easy to self-host and keeping the feature disabled by default would help mitigate that exact maintenance risk.

A plugin system would be a good way to solve this. Feel free to close this PR for now, the code isn't going anywhere, so it can easily serve as a reference or a starting point once the plugin system is ready.

Let me know whenever you start playing around with the plugin architecture, I'd love to help test it out or port this over :)

PR #1769, opened 2026-02-26
feat: customizable item layout on fullscreen player

iiPythonx:
This PR changes the fullscreen player display to allow customizing what items are displayed underneath the artist display, and in what order. Right now it supports toggling between release year, codec (container), and release type.

Additionally, I've readded the default badge styling to improve contrast against the background, since I personally think it looks extremely weird without some sort of a border at the minimum.

Screenshots:

image'
<br>
image
<br>
image

<br>
The implementation remains the same, if there's no data for a given item then the badge will be omitted.

### DIFF
--- a/src/renderer/store/settings.store.ts
+++ b/src/renderer/store/settings.store.ts
@@ -74,6 +74,8 @@ const HomeItemSchema = z.enum([
     'recentlyReleased',
 ]);
 
+const PlayerItemSchema = z.enum(['codec', 'year', 'release_type']);
+
 const ArtistItemSchema = z.enum([
     'biography',
     'compilations',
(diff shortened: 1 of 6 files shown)


src/renderer/store/settings.store.ts
```diff
@@ -862,6 +892,53 @@ export type TranscodingConfig = z.infer<typeof TranscodingConfigSchema>;
...
+    },
+    {
+        disabled: true,
+        id: PlayerItem.SAMPLE_RATE,
+    },
+    {
+        disabled: true,
+        id: PlayerItem.TRACK_NUMBER,
+    },
+];
```
jeffvli, line 940: Can you adjust the defaults to just Codec and Release Year, similar to how it was previously?

iiPythonx: Pushed

Issue #1676, opened 2026-02-09
[Bug]: Waveform generation takes ages


CoutHi:
### I have already checked through the existing bug reports and found no duplicates

- [x] Yes

### App Version

1.4.2

### Music Server and Version

navidrome 0.59

### What local environments are you seeing the problem on?

Desktop Linux

### What happened?

I'm not sure if this counts as a bug, but waveform generation for the playback slider works very slowly and it halts playback during generation.
Disabling it makes switching songs instant so I think that it's the cause of the problem. With it on songs won't start playing for 5-6 seconds sometimes.
Tweaking it's parameters doesn't seem to have an affect either.

### Steps to reproduce

1. Change the playback slider to waveform
2. Click play on a song

### Relevant log output

```shell

```

darius-it:
I've tried fixing the UI to load the normal seek bar first and only then dropping in the waveform once it's finished loading, see following commit on my fork: https://github.com/darius-it/feishin/commit/140dfcb9712e985c0926f7935cc3449b1ea4b560

Unfortunately this only works sometimes, and in some cases the song starts playing smoothly but there is a small pause of ~1s once the waveform gets loaded. Seems like there's still an underlying issue with how wavesurfer loads the waveform (seems to interfere with the player, I assume cause they're both running on the main thread?).

If anyone wants to use my code as a starting point, feel free to. Sadly I don't have enough knowledge about the player and I would most likely need lots of digging to figure it out.

darius-it:
Tried to further optimize by implementing manual peaks decoding and then passing that to wavesurfer's `.load()`.
Seems to work a bit better (no more stutters), but for some songs it still seems to have a small pause sometimes.

For full disclosure, this code is AI-generated using Claude Opus 4.6 (I gave it the input to use manual decoding + the load method, the code is generated): https://github.com/darius-it/feishin/commit/ba5bb436bd0450205d31b51d984edbb02d5c66d8

Not sure if this the right approach, maybe more aggressive optimizations are possible by analyzing peaks in a more manual way (maybe working with raw bytes?) and offloading that to something like a web worker (from my research `OfflineAudioContext` runs in a new thread but something still seems to hang).

Issue #1626, opened 2026-01-30
[Bug]: Waveform stuck eternally loading for formats that can't be processed


mihawk90:
### I have already checked through the existing bug reports and found no duplicates

- [x] Yes

### App Version

1.3.0

### Music Server and Version

Jellyfin 10.11.5

### What local environments are you seeing the problem on?

Desktop Linux

### What happened?

When playing a track in a format that can't be processed by Web Audio, the waveform seekbar is eternally stuck in the loading state. This means there is no way to seek in the track because it doesn't fallback to a regular seekbar.

On a related note if for whatever reason the processing takes long, it presents the same issue, if only temporary.

I think instead of implementing a fallback in case of failure it would be prudent to implement it the other way round, i.e. always render the default seekbar and replace it once wavesurfer finishes processing.

Neither terminal nor Dev Tools show any output.

### Steps to reproduce

1. Add a track to the library that Web Audio can't process
    - in my case this is an MKA/EAC3 file, ALAC presents the same issue, and things like APE would as well
2. Enable waveform seekbar
3. Play back the track

### Relevant log output

```shell

```

dxspotxfinsxnity:
Can confirm for Navidrome.

darius-it:
I've tried implementing a fix for #1676 which might also help in this case, but I'm unable to iron out some issues with that implementation. If anyone is interested in working on this, feel free to use my code as a reference/starting point.

mihawk90:
This appears to be fixed. The track I was testing with just shows the progress bar now.

PR #1781, opened 2026-03-01
Waveform playerbar improvements

darius-it:
This PR implements a few improvements to how the waveform playerbar is handled. (If you believe these changes are not useful, feel free to just close the PR.)

**Main changes:**
- Replace loading spinner with default playerbar (as a fallback, means we can seek while waveform is still loading, should help with #1676, potentially #1626)
- As discussed in the Discord, I've also added a 2 second delay before waveform generation kicks in, to make sure we can start playback ASAP and don't have any network request conflicts. In order to do this, I had to remove some of the old checks and use the manual `wavesurfer.load()`. Seems to work fine, but if any extra checks are needed please let me know.

**Showcase video**:

https://github.com/user-attachments/assets/4e077351-5345-47ad-8dab-acc0c93faf87

I've been using this implementation on and off for a while and didn't find any issues with it, the "stutter" that we talked about in the Discord seems to be an unrelated issue (I've recently encountered it with the waveform playerbar disabled).


pyxfluff:
Are you able to make the delay optional? My server runs locally so network conflicts are not really a big issue which kinda makes the 2s delay unnecessary

darius-it:
> Are you able to make the delay optional? My server runs locally so network conflicts are not really a big issue which kinda makes the 2s delay unnecessary

That should be possible, I'd just have to see where to put the option in the settings and if @jeffvli thinks that's a useful addition.

darius-it:
Done, not sure if I did everything right, but the delay is now configurable. From my end, this PR is ready to be reviewed now.

PR #1800, opened 2026-03-06
fix(mobile): safe-area insets, carousel swipe, skip-error suppression, header badge overflow

Dennist03:
## Summary

This PR fixes several mobile-specific bugs that affect iOS and Android browsers.

### Safe-area insets (iOS notch / Android nav bar)
- The mobile playerbar and layout were clipped by the browser's bottom chrome (home indicator, navigation bar)
- Fixed by applying `env(safe-area-inset-bottom)` to both the layout grid row height and the playerbar padding

### Carousel touch swipe
- Horizontal swipe on the grid carousels did nothing on mobile
- Root cause 1: React synthetic touch listeners are registered as passive by default, so `e.preventDefault()` to block page scroll was silently ignored
- Root cause 2: `touch-action: pan-x` on the navigation container surrendered horizontal control to the browser
- Fix: native DOM `addEventListener('touchmove', handler, { passive: false })` via `useEffect`, and `touch-action: pan-y` on the container so the app controls horizontal direction

### Mobile skip / crossfade error ("Playback paused due to error")
- On Safari/mobile, skipping a track fires `MEDIA_ERR_NETWORK` on the interrupted audio request
- This was incorrectly treated as a real error, pausing playback
- Fix: track `isTransitioning` in a ref (so it stays current inside native callbacks) and return early when a network error arrives during a transition

### Library header badge overflow on small screens
- The total-songs and total-duration badge pills pushed the page title off-screen on narrow viewports
- Fix: hide `.header-badge` via a `@media (max-width: 768px)` rule and clamp the heading font size

## Test plan
- [ ] iOS Safari: playerbar is fully visible above the home indicator
- [ ] iOS/Android: swipe left/right on any home page carousel
- [ ] iOS Safari: skip tracks rapidly — no "Playback paused due to error" toast
- [ ] Mobile (~375 px wide): open an album or playlist detail page — title is readable, badges are hidden


README.md
```diff
@@ -1,213 +1,207 @@
...
-- `pnpm run i18next` - Generate i18n files
-
-## Translation
-
-This project uses [Weblate](https://hosted.weblate.org/projects/feishin/) for translations. If you would like to contribute, please visit the link and submit a translation.
-
-## License
-
-[GNU General Public License v3.0 ©](https://github.com/jeffvli/feishin/blob/dev/LICENSE)
+[GNU General Public License v3.0](https://github.com/jeffvli/feishin/blob/dev/LICENSE)
```
kgarner7, line 207: If this PR is meant for Feishin, then you need to revert this. If it's meant for your fork, then you probably just opened it on the wrong repo.

kgarner7 (changes requested):
Did you mean to make this PR on a fork?

Dennist03:
Yes on a fork, sorry, I have no clue what im doing on GitHub. You can remove it if its in the wrong place

kgarner7:
image

Here's a tip: when making a PR, check which repository you are merging into.
If it's jeffvli/feishin, that's this repository (you _are_ welcome to submit PRs to Feishin for the record).
If you only see one select, then you are merging into your own fork.

PR #1804, opened 2026-03-08
chore: make useCurretnServer nullable, prever useCurrentServerId where possible

kgarner7:
`useCurrentServer()` was incorrectly typed as always returning a `ServerListItem`.
This PR changes it such that it returns null when not set, and then uses `useCurrentServerId()` where posible.

Of note, most of the core logic isn't changed, so many of the places will use an empty string if the server isn't present.

### DIFF
--- a/src/renderer/features/albums/components/album-detail-content.tsx
+++ b/src/renderer/features/albums/components/album-detail-content.tsx
@@ -32,7 +32,7 @@ import {
 import { ListSortOrderToggleButtonControlled } from '/@/renderer/features/shared/components/list-sort-order-toggle-button';
 import { FILTER_KEYS, searchLibraryItems } from '/@/renderer/features/shared/utils';
 import { AppRoute } from '/@/renderer/router/routes';
-import { useCurrentServer, usePlayerSong } from '/@/renderer/store';
+import { useCurrentServerId, usePlayerSong } from '/@/renderer/store';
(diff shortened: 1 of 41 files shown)


mihawk90:
Not really relevant but there's a typo in `useCurretnServer` in the PR title and commit message

jeffvli:
The reason useCurrentServer was changed to always return ServerListItem was because normally you can't reach the components that consume the hook without a server selected due to the `server required` validation. It made it easier for typing purposes as then you don't always have to add a type guard.

I don't mind the refactoring all the hooks to just use useCurrentServerId though.

kgarner7:
> The reason useCurrentServer was changed to always return ServerListItem was because normally you can't reach the components that consume the hook without a server selected due to the `server required` validation. It made it easier for typing purposes as then you don't always have to add a type guard.

Except (for me anyway), it's not obvious and causes the risk of using it in places where said guard does _not_ exist.
It's poor typing.

kgarner7:
Would it be possible to use a context for the current server (instead of `useCurrentServer`)?

jeffvli:
You could, but it's basically the same thing as handling it in the hook, since the zustand store already acts as a (global) context. The proper way is probably to throw an error (and redirect to server selection) if no server is available.

That way it enforces that the hook can't be used outside of where it's made available.

PR #1810, opened 2026-03-09
Remote URL hardcoding compatibility

polyjitter:
This PR should add functionality to hardcode the public/remote URL, as there is some external functionality which may need it (for example, custom sharing URLs in a hosted Navidrome setup). 

Creating this as a draft since I don't have access to proper testing tools. I would appreciate any testing and feedback that can be provided, I use a prettier custom ShareURL for navidrome shares and I really want to get that functionality up and running on Feishin on my homeserver <3


jeffvli:
Fixed one small typo, but besides that it's working well from my testing.
Thanks!

PR #1828, opened 2026-03-11
Fix server queue saving/restoring on Navidrome and OpenSubsonic

Lyall-A:
This fixes a bug causing the queue to be saved in an incorrect order and an error when attempting to restore an empty queue

### DIFF
--- a/src/renderer/api/navidrome/navidrome-controller.ts
+++ b/src/renderer/api/navidrome/navidrome-controller.ts
@@ -636,7 +636,7 @@ export const NavidromeController: InternalControllerEndpoint = {
                 throw new Error('Failed to get play queue');
             }
 
-            const { changedBy, current, items, position, updatedAt } = res.body.data;
+            const { changedBy, current, items = [], position, updatedAt } = res.body.data; // if there is no queue saved, items is undefined
 
             const entries = items.map((song) =>
                 ndNormalize.song(
--- a/src/renderer/api/subsonic/subsonic-controller.ts
+++ b/src/renderer/api/subsonic/subsonic-controller.ts
@@ -1141,7 +1141,7 @@ export const SubsonicController: InternalControllerEndpoint = {
             }
 
             const { changed, changedBy, currentIndex, entry, position, username } =
-                res.body.playQueueByIndex;
+                res.body.playQueueByIndex || {}; // if there is no queue saved, playQueueByIndex is undefined
 
             return {
                 changed,
--- a/src/renderer/api/subsonic/subsonic-controller.ts
+++ b/src/renderer/api/subsonic/subsonic-controller.ts
@@ -1144,8 +1144,8 @@ export const SubsonicController: InternalControllerEndpoint = {
                 res.body.playQueueByIndex || {}; // if there is no queue saved, playQueueByIndex may be undefined from a bug in Navidrome
 
             return {
-                changed,
-                changedBy,
+                changed: changed ?? '',
+                changedBy: changedBy ?? '',
                 currentIndex: currentIndex ?? 0,
                 entry:
                     entry?.map((song) =>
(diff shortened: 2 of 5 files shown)


src/renderer/api/navidrome/navidrome-controller.ts
```diff
@@ -636,7 +636,7 @@ export const NavidromeController: InternalControllerEndpoint = {
                 throw new Error('Failed to get play queue');
             }
 
-            const { changedBy, current, items, position, updatedAt } = res.body.data;
+            const { changedBy, current, items = [], position, updatedAt } = res.body.data; // if there is no queue saved, items is undefined
```
kgarner7, line 639: If items can be undefined, please also update the type definitions.

src/renderer/api/subsonic/subsonic-controller.ts
```diff
@@ -1141,7 +1141,7 @@ export const SubsonicController: InternalControllerEndpoint = {
             }
 
             const { changed, changedBy, currentIndex, entry, position, username } =
-                res.body.playQueueByIndex;
+                res.body.playQueueByIndex || {}; // if there is no queue saved, playQueueByIndex is undefined
```
kgarner7: Subsonic should not be undefined. Some fields are required, and the server not providing it is a bug. I see Navidrome isn't which is something I plan to PR to fix.

src/renderer/features/player/hooks/use-queue-restore.ts
```diff
@@ -46,45 +46,29 @@ export const useSaveQueue = () => {
...
             }
 
             try {
                 await api.controller.savePlayQueue({
                     apiClientProps: { serverId },
                     query: {
-                        currentIndex: queue.default.length > 0 ? player.index : undefined,
+                        currentIndex: queue.items.length > 0 ? state.player.index : undefined,
                         positionMs: useTimestampStoreBase.getState().timestamp * 1000,
-                        songs,
```
kgarner7, line 87: I'm not sure if I agree with these changes.

kgarner7 (changes requested):

src/renderer/api/subsonic/subsonic-controller.ts
Lyall-A: i clarified that this is due to a bug and is not intended, i left it in tho since it is still actively a issue and would cause errors

src/renderer/features/player/hooks/use-queue-restore.ts
Lyall-A: not sure what to do about that then, this fixes the issue and the queue now gets saved in the correct order on both navidrome and opensubsonic. not exactly sure if there was a reason `getState().queue` was used instead of `getState().getQueue()`, but if there was then let me know

kgarner7: My main point is that you do lose shuffle order, but maybe that's correct? There isn't a way to separately save the shuffle state on the server.

src/renderer/api/subsonic/subsonic-controller.ts
```diff
@@ -1141,7 +1141,7 @@ export const SubsonicController: InternalControllerEndpoint = {
             }
 
             const { changed, changedBy, currentIndex, entry, position, username } =
-                res.body.playQueueByIndex;
+                res.body.playQueueByIndex || {}; // if there is no queue saved, playQueueByIndex may be undefined from a bug in Navidrome
```
kgarner7, line 1144: If this is possible, probably better to reflect the subsonic types as well?

src/renderer/features/player/hooks/use-queue-restore.ts
Lyall-A: oh true, i tested that by playing a playlist shuffled but forgot that shuffles the entire queue rather than picking a random position

kgarner7: Not going to lie, I think I'm fine with the shuffle being lost. It isn't reflected visually, so preserving the original queue order is probably more correct. This is fine (and if someone complains, a better idea, possibly with server involvement can come).

PR #1431, opened 2025-12-22
feat: add dlna support

manuel-rw:
As discussed in #195, this is a proof of concept for streaming audio to network connected devices, such as SONOS speakers.
It is not complete; audio streaming is not implemented at all.
It uses DNS discovery to discover the devices and them lists them in a cast menu in the media player.

#### Functionality
- [x] Permanent Sessions ``"Make session persistent" in "src/renderer/features/player/components/right-controls.tsx" I'm using settings for storing current device. It should probably be stored somewhere else, since I would expect it to reset after restart (right now it persists). Where is the best place to put it in Feishin @jeffvli?``
- [x]  The auto-next functionality works right now, but kind of sucks. The DLNA notification about status change can take anywhere between 1 to like 20 seconds to arrive, so the current song is incorrectly displayed in Feishin for that entire time. I could poll for `AVTransportURI` to speed it up, just like with time tracking, which I guess would be the best solution.
- [ ]  To implement rendering I used `upnp-mediarenderer-client` package. It is kind of ancient and doesn't support a lot of options, but it's the best I could find. I had to manually extend it with `enqueue`, `set/getMute` and `play` (to implement speed selection). For `enqueue` I had to copy `buildMetadata` from the original source, because it was private. I don't like this whole solution, but don't have any better ideas apart from scrapping that lib all together and doing it myself. That would be quite a bit of work though.
- [ ]  I copied `mpv-player` and `mpv-player-engine` as a starting point for `dlna`. I don't understand most stuff that's happening in there, but it all seems to work. Not everything is probably necessary though, so I could try to strip it down.
- [ ] Respond to client-side (aka. remote) events `` I'm not sure if this is possible, but I would like to handle device-side control as well (when someone clicks pause or skip or whatever on the speaker). I did not test it yet at all.``
#### Code Clean-up & Refactoring
- [ ]  Factor out common parts: `mpv-player` vs `dlna-player` and `mpv-player-engine` vs `dlna-player-engine`
- [x] Split ``src/shared/types/types.ts`` into multiple files? It is cluttered with all kind of types
- [x] Submit missing types into the DefinetlyTyped package -> @manuel-rw 
- [x]  I'm not sure if `src/shared/types/types.ts` is the best file for keeping DLNA types for IPC. Is there a better place @jeffvli?
- [x]  Poll for `AVTransportURI` change, instead of listening for DLNA notifications.

#### Design
- [x] Improve when loading DLNA devices. It is ugly and spacing is off. -> @manuel-rw 

[...]

PR #1837, opened 2026-03-14
feat: Add DLNA/UPnP casting support

ymidk:
## Summary
- Adds DLNA/UPnP casting to media renderers (WiiM, Sonos, smart TVs, etc.)
- Gapless playback via `SetNextAVTransportURI` with position-based transition detection
- Full transport control: play, pause, stop, seek, volume, mute
- Device discovery via SSDP multicast with cast button UI in player controls
- DIDL-Lite metadata with album art, artist, album name, and duration
- Volume sync from device on connect
- macOS network entitlements and Electron `net.fetch` for sandbox compatibility

## Implementation Details
- **Discovery**: SSDP runs in a child Node process to bypass Electron/macOS sandbox restrictions on UDP multicast
- **SOAP control**: Uses Electron's `net.fetch` (Chromium networking stack) instead of Node's `http` module for macOS LAN access
- **Gapless**: Pre-loads next track on the renderer; detects auto-transitions via position jump-back monitoring
- **Safety**: Error boundary around DLNA player component; startup fallback from DLNA to web player type

## Test plan
- [ ] Build and run on macOS with a DLNA renderer on the same network
- [ ] Click cast button → discover devices → select device
- [ ] Verify playback, pause, seek, volume, and mute
- [ ] Verify gapless track transitions
- [ ] Verify album art and metadata display on the renderer
- [ ] Verify disconnect returns to web player
- [ ] Test on Linux and Windows (SSDP child process should work without sandbox issues)

Closes #1431


jeffvli:
1. Is this based on #1431?
2. Did you test this?

ymidk:
No, it’s not (I didn’t look on here before I did it). Tested with one dlna device and it’s been working all weekend.

Villhume:
I attempted to test with two Sonos speakers (uPnP). However, I get 'No DLNA devices found' in the devices section. The only logs I'm getting are: 

[DLNA] Discovering devices...
[DLNA] Found 0 device(s): []

I've confirmed that these devices are discoverable on my network with other applications, like Spotify's desktop app.
I'm on Windows 10 using the packaged desktop app. I'm going to test on Mac and with the web version later and possibly draft a fix if I have time.

Villhume:
> I attempted to test with two Sonos speakers (uPnP). However, I get 'No DLNA devices found' in the devices section. The only logs I'm getting are:
> 
> [DLNA] Discovering devices... [DLNA] Found 0 device(s): []
> 
> I've confirmed that these devices are discoverable on my network with other applications, like Spotify's desktop app. I'm on Windows 10 using the packaged desktop app. I'm going to test on Mac and with the web version later and possibly draft a fix if I have time.

Testing with the MacOS version does detect my speakers. When playing audio, playback doesn't actually start, though the command is sent to the speakers, and they do receive the relevant metadata. I've noticed similar issues in the past when trying to play local files over UPnP, so I'll do a test with a remote library. The web version doesn't have the feature implemented at all.

Villhume:
@ymidk I've fixed all of the issues I found:

- It now seems to work on MacOS and Windows
- It responds to renderer-side DLNA commands: 
        1.       Changing volume
        2.       Going to the next song
        3.       Going to the previous song (I had some issues implementing this, and the solution is imperfect)
        4.       Playing/pausing
- Your previous fork didn't allow for playing anything but files with an audio/mpeg MIME type, so I have supported most other types, such as FLAC, OGG, etc
- DLNA renderer names should be a bit prettier in the connection menu, especially for Sonos speakers, which need to have their own logic for name conversion because their actual device names are just codes
- A more rigorous metadata transmission

In terms of the tests that you listed, I was able to verify:

- [x] Build and run on macOS with a DLNA renderer on the same network
- [x] Click cast button → discover devices → select device (in my fork I actually changed the behaviour of the popup box so that it behaves more like traditional connection popups, i.e. clicking the wireless symbol doesn't necessarily disconnect the current device automatically, feel free to revert if necessary)
- [x] Verify playback, pause, seek, volume, and mute (now not only controller -> renderer, but renderer -> controller)
- [x] Verify gapless track transitions
- [x] Verify album art and metadata display on the renderer (seemed to work for all metadata types, including art URIs)
- [x] Verify disconnect returns to web player (as I was testing this I noticed that it always goes back to the Web player and not the MPV player were it previously selected, so I'll implement this), but the logic works for Web)
- [ ] Test on Linux and Windows (the child process did not work without issues, because you hardcoded it to /usr/bin/node, so it failed on Windows. I did extensive testing on Windows and lighter testing on Mac, and everything seems fine. Linux remains untested.)

To patch the SSDP discovery, I had to rewrite some of it, and as I was unable to reproduce your HTTP and UDP restrictions on Mac, I didn't keep it as a Node script since that would mean manual installation for users. If this is an issue, I'd be happy to re-include it, or to create a setting to change the behavior. Aside from that, let me know if you have any issues testing with your system, and if not, I'll create a PR here.

https://github.com/Villhume/feishin

Villhume:
I've now also fixed issues with the previous implementation of the DLNA player and the overwriting of volume data. Previously, you would connect to a DLNA renderer, and the volume of the player would be overwritten. Now, the previous volume is reverted upon disconnection, and even if the player would have their client close or crash while casting, the previous volume can be restored from the store on startup. Similar functionality is implemented with the player type. Since ymidk's implementation adds a new DLNA player alongside the Web and MPV players, I made sure that we weren't just blindly switching back to Web after disconnecting from DLNA when MPV was selected before. Like with the volume, this is stored and retrieved should the player crash or close while casting. Also, I noticed that the user could still change the player type in settings when casting, which caused issues. Now, they must disconnect before changing it this way. 

There was also a very minor issue where the Web player would play the first fraction of a second of the first song in the queue upon finishing the queue, this should now be fixed, along with many synchronisation issues between the DLNA renderer and the player/queue system.

@jeffvli I haven't seen activity on ymidk or manuel-rw's forks for some time despite both having issues, so if necessary I can create a new PR for my changes to ymidk's fork, just let me know.

https://github.com/Villhume/feishin/tree/feature/dlna-casting

jeffvli:
@Villhume Feel free to open another PR.

manuel-rw:
> @jeffvli I haven't seen activity on ymidk or manuel-rw's forks for some time despite both having issues, so if necessary I can create a new PR for my changes to ymidk's fork, just let me know.

I have stopped working on mine since this was opened.

In my opinion we need some coordination on this. There are multiple PRs, each with different states and issues. @jeffvli would it be possible that you coordinate?

I suggest we close #1431 and this one (#1837). Then, lets keep #1887 and collaborate together.

manuel-rw:
**Update**: I have closed #1431 and posted test results on #1887

PR #1869, opened 2026-03-23
feat: macOS Now Playing / Control Center integration

baumloveggod:
## Summary

- Adds macOS Now Playing / Control Center integration using the HTML5 MediaSession API
- The main process fetches album artwork and sends track metadata to the renderer via IPC; Electron automatically mirrors this into the OS-level Now Playing info-center
- Wires up MediaSession transport action handlers (`play`, `pause`, `stop`, `nexttrack`, `previoustrack`) so Control Center buttons and media keys control playback

## How it works

1. `now-playing.ts` — triggered on every `player-update-metadata` IPC event; fetches artwork as a data URL and sends metadata to the renderer via `update-media-session`
2. `app.tsx` — listens for `update-media-session` and updates `navigator.mediaSession` metadata and playback state
3. `use-main-player-listener.tsx` — registers/unregisters MediaSession action handlers alongside existing media key handlers

Requires "Media session" to be enabled in Settings → Hotkeys.

## Test plan

- [ ] Enable "Media session" in Settings → Hotkeys
- [ ] Play a track — verify title, artist, album and artwork appear in Control Center
- [ ] Use Control Center transport buttons to skip/pause — verify playback responds
- [ ] Pause playback — verify Control Center shows paused state
- [ ] Stop playback — verify Now Playing disappears from Control Center

### DIFF
--- a/src/main/features/core/player/index.ts
+++ b/src/main/features/core/player/index.ts
@@ -9,6 +9,7 @@ import process from 'process';
 import { getMainWindow, sendToastToRenderer } from '../../../index';
 import { createLog, isWindows } from '../../../utils';
 import { store } from '../settings';
+import { updateNowPlaying } from './now-playing';
 
 import { PlayerData } from '/@/shared/types/domain-types';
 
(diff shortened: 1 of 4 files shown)


kgarner7 (changes requested):
You are missing the fact that media session is **already** available if in the web browser, or using the web player and media session is enabled. If that isn't working for you, try debugging that first.

PR #1872, opened 2026-03-25
feat(albums): show grouping tags on album detail page

romainvigneres:
The grouping field is already exposed as a filter on the Album list page for Navidrome but isn't displayed on the Album detail page unlike other filterable fields.

Added Grouping tags on the Album detail page matching pattern used for Mood, Record labels, Genre, Tags.

The pills are clickable on Navidrome servers and link to the Album list page filtered by that grouping value (matching Record labels behaviour).


romainvigneres:
Added several translations for the grouping tag, based on the documentation of kid3

PR #1892, opened 2026-03-30
feat: show unknown artist release types as dynamic categories

korpseluv:
changes:
- normalize release-type values used by artist discography grouping
- surface unknown release types as their own categories instead of falling back to Album behavior
- preserve existing known category behavior

why:
- tags like remix should group consistently
- unknown tags eg. slowed should be visible to user as a section

scope:
- navidrome/subsonic release-type metadata paths
- no jellyfin changes made
- no settings migration or enum expansion

https://github.com/user-attachments/assets/f4359f8c-6b32-496f-ad7b-f16793903497


jeffvli:
Are you getting the non-standard release types from musicbrainz directly or are you manually tagging them?

korpseluv:
In most cases theyre musicbrainz tags. Remixes, Lives, Demos, Compilations, Live preformances.

In the specific case I provided, I have the albums tagged manually.

PR #1898, opened 2026-03-31
fix: prevent media session handlers from breaking/vanishing during playback

Auzlex:
Hello! This contains a small fix for the use media session disappearing during rapid or quick use of media controls through the electron app or on android in web browser.

Issues I have encountered is usage of the media controls in quick succession or in rare cases will loose media controls until playback is stopped and resumed again. I use media controls a lot and noticed some issues using them.

I believe the cause was handlers were being registered or destroyed on change state or render of the page (i think). 

There is also debounce to handle rapid track skipping.

I only looked at the project for a short time so im open to suggestions if this is the correct or wrong place to implement/fix this.

tested on windows and android and got a friend to test it on linux. media controls now stick around.


jeffvli:
Can you clean up your branch?

PR #1904, opened 2026-04-04
Fix: Queue sidebar obscured by bottom controls

### LINKED ISSUE #1895: [Bug]: Queue sidebar obscured by bottom controls

- [x] Yes

### App Version

1.9.0

### Music Server and Version

Navidrome

Desktop Linux

### What happened?

Image

### Steps to reproduce

1. Pin the right queue sidebar
2. It is obscured at the bottom by player controls

### Relevant log output

```shell

```

aryanjsingh:
## Summary

The pinned queue sidebar was being obscured by the bottom player controls because the main content area did not account for the player's height. This has been fixed by wrapping the main application router in a container with bottom padding, ensuring there is enough space for content to be scrolled into view above the controls.

## Changes

- **src/renderer/app.tsx**: Import the Box component from Mantine to use as a layout container.
- **src/renderer/app.tsx**: Wrap the AppRouter in a Box component with bottom padding to prevent the player controls from overlapping the content. The Box is also given full height to ensure child layouts behave as expected.

## Related Issue

Closes #1895

### DIFF
--- a/src/renderer/app.tsx
+++ b/src/renderer/app.tsx
@@ -1,5 +1,5 @@
 /* eslint-disable perfectionist/sort-imports */
-import { MantineProvider } from '@mantine/core';
+import { Box, MantineProvider } from '@mantine/core';
 import { Notifications } from '@mantine/notifications';
 import 'overlayscrollbars/overlayscrollbars.css';
 import '/styles/overlayscrollbars.css';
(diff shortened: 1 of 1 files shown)


jeffvli:
That does not fix the issue. The issue stems from the glassy dark theme itself, which is overriding the app layout.

PR #1906, opened 2026-04-04
chore: upgrade dependencies

kgarner7:
Run `pnpm upgrade`. Also add forced resolutions for `xml2js` (`mpris-service` has not been updated) and `react-router` (for whatever reason, another dependency is resulting it it using an older version). This also resulted in new lint changes.

I have not upgraded vite to 8.0, or tried to upgrade the transitive dependencies of `vite-plugin-pwa`
Breaking change from upgrading mantine (minor version) is that `modalKey` -> `modal`


jeffvli:
`"fast-average-color": "^9.5.1",` seems to be broken for the header colors.

kgarner7:
Looks like https://github.com/fast-average-color/fast-average-color/commit/666cf2f3a15eede4a4fa876a4be374b6b04b6a1c is causing some issues. I'm going to just hold it at 9.5.0.

kgarner7:
Why 9.5.1 appears to be causing this problem: https://github.com/fast-average-color/fast-average-color/pull/120 (I tested this manually by changing the version to `github:kgarner7/fast-average-color#507ca1c0d5e1a68d49d05ae8b5281ef194819fc8`)

Issue #1917, opened 2026-04-06
[Bug]: Glassy Dark theme issue with Lyrics on Sidebar


Nicolach5:
### I have already checked through the existing bug reports and found no duplicates

- [x] Yes

### App Version

1.10.0

### Music Server and Version

Navidrome v0.61.0

### What local environments are you seeing the problem on?

Desktop Linux, Desktop Windows

### What happened?

The lyrics box in the sidebar is displaying incorrectly when using the Glassy Dark theme

Default Dark theme:

<img width="2560" height="1404" alt="Image" src="https://github.com/user-attachments/assets/725673bd-6e05-416c-b75f-151b3c866be9" />
<img width="2560" height="1404" alt="Image" src="https://github.com/user-attachments/assets/61af79a6-8bc8-48dc-840b-7df6ab671187" />

Glassy Dark theme:

<img width="2560" height="1404" alt="Image" src="https://github.com/user-attachments/assets/26a0768a-6f8a-4012-aa2b-784c7d6a454c" />
<img width="2560" height="1404" alt="Image" src="https://github.com/user-attachments/assets/9d7a52dc-4604-4f9f-99c7-e0acba9afc3f" />

The issue was also present in 1.9.0
I couldn't find any way to fix it

### Steps to reproduce

1. Set theme to Glassy Dark
2. Play a song with lyrics on sidebar

### Relevant log output

```shell

```

pyxfluff:
duplicate of #1895

PR #1887, opened 2026-03-30
Implementation of DLNA casting with additional fixes and features

Villhume:
This PR fixes many issues in ymidk's fork to introduce DLNA/UPnP support to Feishin. I have tested both that fork and manuel-rw's similar implementation, and found issues with both. ymidk's only worked on MacOS (and possibly Linux), failing to discover DLNA devices on Windows and was only able to play files with a MIME type of audio/mpeg (MP3). Testing on Mac, I was unable to reproduce any HTTP and UDP restrictions during discovery, and as the previous solution to run discovery within a Node script (which would require the user to install Node to use DLNA functionality), I rewrote SSDP discovery without it. If that ends up being an issue, I'd be happy to reimplement it, possibly as an additional setting. I also enabled the identification and sending of MIME types other than MP3, such as FLAC, WAV, OGG, etc, so those files can now be played. More minor fixes to the original fork include: 
- Improved metadata transmission to the renderer
- Devices in the connection menu are now displayed just with their device name, and Sonos devices, whose device names are codes, are displayed with their 'room' names (like 'Bedroom Speaker' as opposed to 'RINCON_<MAC Address>', or something like that). I only had two DLNA devices to test on, a Sonos speaker and a Smart TV, so other speakers might need similar treatment.)
- Clicking the connection menu button now does not automatically disconnect a speaker. Like Spotify's, it now requires the pressing of the 'disconnect' button, and then allows the user to select another speaker if desired.
- When disconnecting from a DLNA device, default player (Web player) and volume setting are no longer blindly applied; the previous local player and previous local volume are used as well. This will work even if the app crashes or is closed while casting by storing these values for the next startup.
- Minor and major synchronisation issues between the DLNA player, Feishin, and the queue system should be fixed.
- There was a very minor issue with the Web player (in Feishin, not ymidk's fork) in which the Web player, when finishing playing the queue, would play the first fraction of a second of the first song before stopping, which was inaudible except for tracks in which audio starts immediately. This has been resolved.
- The user could previously change the player setting to MPV or Web during DLNA playback or connection, which could cause playback issues. Now, DLNA (while connected) will come up as an actual player in this menu, and the drop-down will be locked and greyed out while connected, preventing those changes.

I also implemented a new feature, which is the ability for certain renderers to send commands to Feishin (or rather for Feishin to interpret the button presses based on the status of playback), which then change the local playback on Feishin:
[...]

PR #1924, opened 2026-04-07
fix: glassy dark theme

Nicolach5:
This PR fixes both the issue with the lyrics module (#1917) and the queue (#1895)

After fix:
Screenshot_20260407_130130
Screenshot_20260407_130146

### DIFF
--- a/.github/workflows/publish-pr.yml
+++ b/.github/workflows/publish-pr.yml
@@ -27,6 +27,8 @@ jobs:
         needs: wait-for-lint
         if: always() && (needs.wait-for-lint.result == 'success' || needs.wait-for-lint.result == 'skipped')
         runs-on: ${{ matrix.os }}
+        env:
+            NODE_OPTIONS: --max-old-space-size=4096
 
         strategy:
             matrix:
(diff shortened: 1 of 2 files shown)


Nicolach5:
The first fix removed the glassy effect for the player bar, so i readded that and instead calculated the height of the right sidebar separately

jeffvli:
@pyxfluff

pyxfluff (changes requested):
I don't think this is ready, it seems to be pushing up for me, at least on web.

Image

Additionally, the removal of the glassy playerbar isn't something you did and can be fixed with this:
```css
.fs-native-scroll-area-module-scroll-area {
  height: 100%
}
```

Nicolach5:
I'll try to fix it tomorrow as I really like the theme.
About the glassy player bar, in my first edit I removed
```
.fs-main-content-module-main-content-body {
    height: 100vh;
} 
```

which is how the content goes underneath the playerbar, but it also affects the right sidebar, so I was fixing my mistake there

Nicolach5:
As mentioned in #1887, it seems there were other issues causing the publish task to fail because of an OOM error in the macos build. I tested in my fork and this happened even without my changes, inconsistently failing to build.

PR #1926, opened 2026-04-07
fix(linux): remove unnecessary screen capture from audio loopback handler

noctuum:
## Problem

Screenshot_20260408_032622

On Wayland, Feishin shows an xdg-desktop-portal "Share screen" dialog on every launch. This happens because `setDisplayMediaRequestHandler` calls `desktopCapturer.getSources({ types: ['screen'] })`, which creates a new ScreenCast session each time. Electron does not persist PipeWire restore tokens across `desktopCapturer` sessions, so the dialog cannot be suppressed.

The video source provided by the handler is unused — the renderer requests `video: false` (since ef129e46) and only consumes audio tracks for the visualizer.

## Changes

- Removed `desktopCapturer.getSources()` call from the display media handler
- Simplified handler to `callback({ audio: 'loopback' })`
- Removed unused `desktopCapturer` import

## Alternatives considered

- **Pass video + strip tracks in renderer**: Does not solve the problem — `desktopCapturer.getSources()` still triggers the portal dialog
- **Platform-conditional logic** (skip video on Wayland only): Adds complexity for no benefit since video is unused on all platforms
- **PipeWire restore tokens**: Electron's `desktopCapturer.getSources()` API does not support storing or re-presenting restore tokens — this is an Electron limitation, not a portal bug

## Testing

- Arch Linux, KDE Plasma 6, Wayland, Electron 39
- Verified: no portal dialog on launch
- Verified: visualizer works with system audio loopback
- Verified: reverting to unpatched build reproduces the dialog

### DIFF
--- a/src/main/index.ts
+++ b/src/main/index.ts
@@ -5,7 +5,6 @@ import {
     app,
     BrowserWindow,
     BrowserWindowConstructorOptions,
-    desktopCapturer,
     globalShortcut,
     ipcMain,
     Menu,
@@ -734,19 +733,7 @@ async function createWindow(first = true): Promise<void> {
     });
 
     mainWindow.webContents.session.setDisplayMediaRequestHandler((_request, callback) => {
-        desktopCapturer
-            .getSources({ types: ['screen'] })
-            .then((sources) => {
-                if (sources.length > 0) {
-                    callback({ audio: 'loopback', video: sources[0] });
-                } else {
-                    callback({});
-                }
-            })
-            .catch((err) => {
-                log.warn('desktopCapturer.getSources failed', err);
-                callback({});
-            });
+        callback({ audio: 'loopback' });
     });
 
     if (!disableAutoUpdates() && store.get('disable_auto_updates') !== true) {


kgarner7 (changes requested):
This capture is necessary to work on macOS and Linux, so this PR is incorrect.
There is a prompt about whether to enable audio capture in Feishin first, although it looks like that can't be toggled from settings.
Linux audio capture doesn't seem to be working, so a better option might be to just disable it.

noctuum:
@kgarner7 Thanks for the review!

System audio loopback is a brand-new feature added in v1.11.0 (commits 94886a2d, ef129e46, 3f300c40 — all from April 5). It does work on Linux — I've tested it on Arch Linux with KDE Plasma 6, Wayland, PipeWire, and Electron 39. The visualizer captures system audio correctly with just `{ audio: 'loopback' }`.

The key observation is that the maintainer already set `video: false` in the renderer (ef129e46), so the video source provided by the handler is explicitly unused. On Wayland, the `desktopCapturer.getSources()` call triggers an xdg-desktop-portal ScreenCast dialog on every launch because Electron doesn't persist PipeWire restore tokens — this PR simply removes that unnecessary call to match the renderer's intent.

I don't have access to a Mac, so I can't verify macOS behavior. If macOS requires the video source for Chromium's internal audio capture pipeline, I'm happy to make this conditional — skipping desktopCapturer.getSources() only on Linux where PipeWire/PulseAudio loopback works independently of screen capture.

kgarner7:
I have no idea how you got it to work on Linux, because I can confirm on my own setup it's just a flatline. Nevertheless, I do see that with just audio loopback macos (probably; at least the visualizer is doing a silly) and windows are working, so I guess this is good enough.

PR #1930, opened 2026-04-08
Fix MPV visualizer not working on macOS and handle exclusive mode UX

york9675:
# Summary
This PR fixes MPV visualizer capture on macOS by restoring a macOS-compatible system audio capture flow and adding the required `NSAudioCaptureUsageDescription` permission key for mac builds.

It also improves UX around MPV Audio Exclusive Mode:
- Adds a error when opening the visualizer while Exclusive Mode is enabled
- Prevents visualizer system-audio capture while Exclusive Mode is on
- Updates the Audio Exclusive Mode setting description to clearly state that visualizer capture will not work in this mode

## Cause
Switch visualizer capture to `getDisplayMedia({ audio: true, video: false })` works for Linux/Windows loopback paths, but on macOS this can result in no usable system-audio track for the visualizer flow.
At the same time, packaged macOS builds also need the audio capture usage description key for proper permission flow.

## Fixes
- Restored a macOS-specific display media handler path that provides a real screen source together with loopback audio.
- Updated renderer capture constraints to request video on macOS (while keeping audio-only behavior on non-macOS).
- Added `NSAudioCaptureUsageDescription` to mac build configs.


kgarner7:
I'm going to have to rely on your testing for whether macOS works. 

As for exclusive mode, at least on Windows this is a better flow.
I do wonder whether just disabling the visualizer tab (at least in the full screen player) might be a better option, but I don't think there's sufficient space to communicate _why_ it's disabled (or hidden).
Thoughts @jeffvli ?

york9675:
> I'm going to have to rely on your testing for whether macOS works.

Here’s a short demo showing the implementation on macOS. I’ve also included a look at the new exclusive mode error prompt in action.

https://github.com/user-attachments/assets/56d892f4-ea12-4692-aeca-cfa0279237fc


> I do wonder whether just disabling the visualizer tab (at least in the full screen player) might be a better option, but I don't think there's sufficient space to communicate why it's disabled (or hidden).

Regarding the UI, I think a disabled state with a hover tooltip might be a good middle ground. It keeps the layout consistent while clearly communicating why the feature is locked (e.g., "Unavailable during Exclusive Mode") without needing extra UI real estate

kgarner7:
Thanks, I do think that is probably a good option for the full screen player.
Not as sure about the sidebar mixed lyrics/visualizer, although I'm sure there's a good mechanism.
Also, thanks for the video on macOS.

PR #1941, opened 2026-04-10
macOS 26-friendly icon

eramdam:
What the title says.

clear-dock
dark-dock
light-dock


eramdam:
@jeffvli the build issue was caused by the GitHub workflow using `macos-latest` instead of `masos-26`, it builds fine now.

PR #1948, opened 2026-04-12
Improve case for romance translations

speedy-axolotl:
Right now the sidebar shows the text on title case on languages such as Spanish, where it is not correct. This tweak makes the language more natural all across the GUI in Spanish and other romance languages that follow the same rule.

### DIFF
--- a/src/i18n/i18n.ts
+++ b/src/i18n/i18n.ts
@@ -193,9 +193,24 @@ const upperCasePostProcessor: PostProcessorModule = {
     type: 'postProcessor',
 };
 
+// Languages that follow sentence-style capitalization in titles
+// Only the first word (and proper nouns) are capitalized, unlike English-style title case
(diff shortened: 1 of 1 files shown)


jeffvli:
This is a good idea, although it's kind of a workaround fix to the core issue. When I originally planned the locale files, I wanted to make all the strings lowercase to improve reusability, but it ended up as a mistake due to the issue you're pointing out here. 

I'm going to think about this for a bit, because I've been meaning to eventually do a full conversion to proper i18n casing which would then remove the need (in most cases) for the text postprocessors.

speedy-axolotl:
I wondered about that. It makes sense to leave casing to the translators, specially if the strings are not really being reused at the moment.

jeffvli:
Track https://github.com/jeffvli/feishin/pull/1998 for updates

PR #1962, opened 2026-04-16
Stretch the wavesurfer waveform to the full height

mitchray:
Currently with ```normalize: false``` the Wavesurfer waveform 'accurately' displays the loudness of the raw file, but in cases where that mastering is low volume the end result is very flat and not that useful

Changing to ```normalize: true``` will [stretch the waveform](https://wavesurfer.xyz/docs/types/wavesurfer.WaveSurferOptions) between min and max height (purely visual, has no effect on audio) to better convey the dynamic range of the track.

Before
firefox_w7ppQ72rnS

After
firefox_xTpxmdBFnI

### DIFF
--- a/src/i18n/locales/en.json
+++ b/src/i18n/locales/en.json
@@ -983,6 +983,8 @@
         "playerbarWaveformBarWidth": "waveform bar width",
         "playerbarWaveformGap": "waveform gap",
         "playerbarWaveformRadius": "waveform radius",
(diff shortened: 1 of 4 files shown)


jeffvli:
Would you mind adding this as a toggle in the settings?

mitchray:
I'll give it a shot

mitchray:
Setting added, let me know if anything else is needed

PR #1970, opened 2026-04-20
feat: use urlsearchparams instead of qs

kgarner7:
`qs` is _extremely_ slow when parsing large query strings.
Since Subsonic query strings can be large (for creating large playlists, or saving queues), this introduces a _lot_ of blocking latency.
Prefer using `URLSearchParams` and generating the same mapping instead. 
For 35k items, this goes from ~10s to less than 1.


jeffvli:
Looks good. I did a couple benchmarks myself and was seeing a ~6.7x speed increase.

This fortunately doesn't have much risk for not using qs.parse since we don't have any complex query strings to parse.

PR #1972, opened 2026-04-20
Added Graphic EQ and Compressor

### LINKED ISSUE #1013: [Feature]: Equalizer

- [x] Yes

Not server-specific

Having a graphic equalizer similar to the one in Supersonic would be great.
ffmpeg supports EQ through [filters](https://ffmpeg.org/ffmpeg-filters.html#equalizer)

screenshot from supersonic

keebsxd:
Uses MPV playback. 12 Band EQ and a simple Compressor with some basic presets.


githubssh.pub
```diff
@@ -0,0 +1 @@
+ssh-ed25519 AAAAC3NzaC1lZDI1NTE5AAAAIBSg3B3tLvjv4twwxcrTwJE4kH+lMA4okwEk+nvq0MrS kevinadamaral@gmail.com
```
kgarner7: Delete this

.github/workflows/feishintest.yml
```diff
@@ -0,0 +1,22 @@
+name: Test
```
kgarner7: Why is this adding a new workflow?

package.json
```diff
@@ -188,7 +188,9 @@
     },
     "pnpm": {
         "onlyBuiltDependencies": [
+            "abstract-socket",
```
kgarner7: Why are you changing the build dependencies?

kgarner7 (changes requested):
See required changes. Note that I will not approve this unless it is also supported for web player (the most common player).

.github/workflows/feishintest.yml
keebsxd: removed. Was using it to test locally and wanted to demo the --fix flag.

package.json
keebsxd: Fixed. They were transient dependencies that were auto added during my pnpm install. Now removed.

keebsxd:
Working on adding Web player functionality. I will request another review when appropriate

keebsxd:
Added Web player EQ

kgarner7:
Thanks! The web audio implementation actually works really well. It looks like for MPV, the ranges have too high of a collision, tested by just turning up all the knobs to +12. In web audio, this just makes the whole thing loud, but with MPV the audio is very distorted.

keebsxd:
I'll do some investigating and see what I can do about the clipping but I suspect there won't be much as clipping is expected and the same thing happens with VLC and other players that use the FFMPEG backend

keebsxd:
> Thanks! The web audio implementation actually works really well. It looks like for MPV, the ranges have too high of a collision, tested by just turning up all the knobs to +12. In web audio, this just makes the whole thing loud, but with MPV the audio is very distorted.

I pushed some updates to fix the EQ on MPV to prevent distortion. Its consistent with how VLC handles EQ and Compression.

Feature should now be complete unless bugs are found.

kgarner7:
Thanks. Last two points (although you'll have to ask @jeffvli to help review those):
1. When switching between mpv to web audio, it looks like the equalizer is only applied _after_ you make a change.
2. The behavior of the equalizer feels rather different between web audio and MPV. In web audio, if I increase one of the frequencies, that region gets louder and the rest is (mostly) not impacted. In MPV, while I don't hear distortion anymore, the effect appears to be make _everything else_ quieter instead when increasing. Decreasing however appears to have the proper effect.

keebsxd:
I've changed the behaviour of the preamp

> Thanks. Last two points (although you'll have to ask @jeffvli to help review those):
> 
>     1. When switching between mpv to web audio, it looks like the equalizer is only applied _after_ you make a change.
> 
>     2. The behavior of the equalizer feels rather different between web audio and MPV. In web audio, if I increase one of the frequencies, that region gets louder and the rest is (mostly) not impacted. In MPV, while I don't hear distortion anymore, the effect appears to be make _everything else_ quieter instead when increasing. Decreasing however appears to have the proper effect.

This should be fixed now. I cant seem to replicate your issue with the EQ not being auto applied? Check the new version and let me know

jeffvli:
Could you refactor the EQ settings component and anywhere else that you're using custom inputs to use the existing components instead?

This includes the VerticalSlider, HSlider, and Toggle components that already have existing wrapped components available. You will need to rebase development onto your branch as the newly commited Mantine v9 upgrade now includes vertical sliders natively.

Using your custom components means that they do not follow the existing app theme.

keebsxd:
@jeffvli

Hey so, I made the changes to remove the custom components. Let me know if its suitable or not and ill make any changes as needed.

kgarner7:
https://github.com/user-attachments/assets/0ff56fb9-1cab-4ad8-ba6d-7bf0dc52fe9f


Apologies for the delay. I gave this a look, and at least on Linux, the sliders are _really_ hard to use (video attached; for whatever reason the first video attached was borked by GitHub).

keebsxd:
> equalizer-inconsistent-scrolling-2.mp4
> 
> Apologies for the delay. I gave this a look, and at least on Linux, the sliders are _really_ hard to use (video attached; for whatever reason the first video attached was borked by GitHub).

Hey so the issue was the rotated sliders I used from mantine. The sliders were changing value based on horizontal movement of the cursor rather than vertical. I did some more digging and am now using useMove for vertical sliders. I also added in the ability to manually type in the value for the EQ for those that prefer it as such.

jeffvli:
The preamp setting slider seems to be broken. The component duplicates itself when dragged.

image

keebsxd:
Woops that issue happened due to the change of how the sliders operate. I've gone ahead and bug tested it all too now and I cant find any more issues either. I was kinda being lazy before and not thinking the problems through but the fix I have now should be final. Fix has been implemented locally just gotta do lint will get to it in the next 48hours.

PR #1975, opened 2026-04-20
[UI] Made sidebar image just use flex

veryprofessionaldodo:
Noticed an issue when dragging the sidebar pane, where the image would not update until the end of the movement. This makes it overlap with other elements. Showing the before and after:

Before: 

https://github.com/user-attachments/assets/3bb94d96-8aed-4b34-aa4e-950d1ca3545f

After:

https://github.com/user-attachments/assets/4284f828-7b16-4832-8e2e-7cdbecbccaa5


jeffvli:
You may need to keep a min-height to the .image-container. The container does not expand if no image is available.

image

veryprofessionaldodo:
@jeffvli Apologies for the delay with the follow-up, but it is now fixed! Noticed explicitly setting heights wasn't working, but turns out that wasn't the right approach either. If we set the height, when we resize the sidebar it doesn't collapse, it just stays the same height with decreasing width. 

With the new change, if nothing is playing back, we now just use the aspect ratio of 1 to maintain a square. This makes it look quite close with the album art, but it doesn't override the cover art aspect ratio. So if for some reason the song metadata has a non-square ratio, it should retain the existing behavior as before :) 


https://github.com/user-attachments/assets/6e3bc4e9-f565-4de6-9755-55d3edcbac6f

src/renderer/features/sidebar/components/sidebar.tsx
```diff
@@ -216,11 +215,9 @@ const SidebarImage = () => {
             onClick={expandFullScreenPlayer}
             onContextMenu={handleToggleContextMenu}
             role="button"
-            style={
-                {
-                    '--sidebar-image-height': leftWidth,
-                } as CSSProperties
-            }
+            style={{
+                aspectRatio: (isRadioActive && radioImageUrl) || imageUrl ? 'auto' : 1,
```
jeffvli: Using `aspectRatio: auto` makes it so that the image can have variable height which isn't intended for the sidebar image. It should always be a 1 / 1 square.

veryprofessionaldodo: Fixed and synced with latest development

PR #1981, opened 2026-04-26
Fix syntax errors in publish-alpha.yml

Kekechann:


jeffvli:
Where are you seeing syntax errors?

The publish-alpha.yml file is in valid yml format.

PR #1998, opened 2026-05-02
Use proper casing for i18n locales

jeffvli:
- Converts en.json and all existing locales to follow proper casing as opposed to all lowercase
- Removes all i18n postprocessing in the renderer


speedy-axolotl:
Looks great for Spanish, thanks!

PR #2007, opened 2026-05-04
also use external image url for discord-rpc with navidrome

photovoltex:
This just simply allows `ServerType.NAVIDROME` to also use `imageUrlRef.current` when available.

This seems to just work as it is. I tried to find the reason why this was separated in 72d0fca28b32a7c56fd98879032c2e8e10f48218, but I couldn't find an answer for it. So was there a reason behind the separation, or maybe just missing testing for the individual `ServerType`'s? 

Thanks in advance :)

### DIFF
--- a/src/renderer/features/discord-rpc/use-discord-rpc.ts
+++ b/src/renderer/features/discord-rpc/use-discord-rpc.ts
@@ -276,12 +276,12 @@ export const useDiscordRpc = () => {
 
                 if (discordSettings.showServerImage && song) {
                     if (song._uniqueId === currentSong?._uniqueId && imageUrlRef.current) {
-                        if (song._serverType === ServerType.JELLYFIN) {
-                            activity.largeImageKey = imageUrlRef.current;
-                        } else if (
-                            song._serverType === ServerType.NAVIDROME ||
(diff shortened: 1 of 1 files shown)


jeffvli:
The reason it is disabled for Navidrome / Subsonic is because authentication is included in the image URL with the [getCoverArt](https://opensubsonic.netlify.app/docs/endpoints/getcoverart/) endpoint.

Example:
`http://your-server/rest/getCoverArt.view?id=123&u=demo&p=demo&v=1.13.0&c=AwesomeClientName&f=json
`

If you send this to RPC, you expose the following data to all discord users:
1. Your public server URL
2. Your username
3. Either your plaintext password  or [tokens that can be decoded to plaintext](https://opensubsonic.netlify.app/docs/api-reference/#parameters)

If you tested this yourself on dev, I would recommend rotating your password.

photovoltex:
Thanks for the inside and that missing part of information. And yeah makes total sense with that knowledge.

I had a different solution previously where I simply replaced the hostname from the retrieved `info.imageUrl` with the hostname from `imageUrlRef.current` when they didn't matched. I guess something into that direction would be a preferred solution then?

PR #2011, opened 2026-05-06
Fixed bad smart playlist fields

### LINKED ISSUE #1645: [Bug]: Inaccurate sorting in smart playlist

- [x] Yes

### App Version

1.4.1

### Music Server and Version

Navidrome latest

Desktop Windows

### What happened?

I'm trying to order a smart playlist using the following parameters

Image

Note that :
- Playcount is reversed (played songs are invisible at the bottom)
- Release Date is reversed as well (i have more recent albums)
- Track number seems reversed but switching ascending/descending has no effect, so this parameter seems ignored.

Actual track order in that album :

Image

On navidrome, i see the files orderd as most played from the top

### Steps to reproduce

Create a smart playlist with similar parameters i showed above, see that it's not the same in navidrome

### Relevant log output

```shell

```

nova77:
Smart playlist sorting was not working as expected for many of the fields.
There seems to be significant discrepancies between `NDSongQueryFields` values and [Navidrome's listed fields](https://www.navidrome.org/docs/usage/features/smart-playlists/#fields).

This PR fixes #1645, as well as all the other entries that were clearly wrong.

**BEFORE**:

before

**AFTER**:

after

**IMPORTANT**: This will not affect the JSON value of existing playlists! If you have `track` in your `sort` list, you will have to manually update it. Alternatively, you can use the UI to delete and re-add the correct value.

Note: there seems to be other potentially wrong entries because they're not found in the documentation, such as `recordingdate`, `albumartistssort`, `artistssort`, `composersort`, and `lyricistsort`. We also have several `musicbrainz_*` entries, but the documentation only mentions `mbz_*`. I haven't touched those, but might be worth considering in this PR or a follow up.


jeffvli:
Thanks!

PR #2012, opened 2026-05-06
feat: reading custom css from external file

hexahigh:
This PR adds a new feature that makes the custom CSS feature read and write to an external file. This allows for integration with programs like Matugen to automatically generate a color theme based on the wallpaper.
image


jeffvli:
I think it would make more sense to add this as another setting (component, not in the actual settings store) instead of modifying the existing description.

Instead of an input for the setting, you can add a button to open the directory that the file should be placed.

hexahigh:
Do you suggest having this new option be alongside the existing custom CSS option or having it replace the existing option?

jeffvli:
Underneath the existing one.

hexahigh:
Is this what you meant?
image

jeffvli:
That works!

hexahigh:
Should be ready for merging now. I would recommend squashing it since I made quite a few commits.

PR #2017, opened 2026-05-08
feat(sidebar): group playlists with "/" into collapsible folders

shawnsquire:
## Summary

Adds an optional, front-end-only, cosmetic grouping for sidebar playlists. Playlists whose name contains a forward slash (e.g., `Style/Phonk House`, `Atmospheric/Relaxation`) are grouped into collapsible folder rows by the slash-delimited prefix. The folders can be collapsed to hide playlists of some folders, and expanded to show the playlists in that folder (which display without the prefix). Folders without a `/` will render at the root as they do today.

I believe this aligns with roadmap item "Client-side playlist folders". I wanted this feature because I have very different use cases for different playlists, and usually want to only focus on a small subset of them when I'm looking for music. Since there is no support for playlists in Subsonic API, I decided a front-end modification was the simplest path to get this.

Enable or disable via **Settings -> General -> Sidebar -> Enable folders**.

## Scope
- No API or store changes beyond the new boolean setting
- Only the first-level grouping name is considered (multi-level is out of scope)
- The dedicated `/playlists` page is unchanged; sidebar only effect

## Screenshots

**Before**

image

**After**

image

## Details

- Added a new setting to enable / disable this feature
- No changes occur if the user does not feature `/` in their playlists
- Updating playlists names to have slash-delimiters will display the folders
- Added a toggle to "expand all" and "collapse all" if there are folders present
- Playlists will continue to appear in alphabetical order (e.g., root-level playlists may display between folders)
- Modified `sidebar-playlist-list` to export `PlaylistRowButton` so it can render a `<PlaylistFolderTree>`
- English language added


jeffvli:
This is actually a good idea! I have a couple thoughts on it:

1. Is there any limitation to making this multi-level or why is multi-level out of scope here?
2. It would be a good idea to make the SEPARATOR also configurable from the settings, because in certain cases users may already include `/` in their playlist names but not want that to be the separator

shawnsquire:
Thanks for the feedback @jeffvli ! Glad you like it.

I considered making it multi-level! I know there could be a use for it. My holdback was finding the sweet spot for how to indicate folder level. Too little left-side margin and the difference between different folder depths is hard to separate. Too much, and nested folders can be too wide for the sidebar. The left-side indentation indicators could add to visual clutter. I first wanted to see if there was community interest in the idea before committing to any style identity.

I like your idea about the configurable separator. It solves a problem I was worried users would have if they had playlists with "/" in them. Want me to roll it into this PR, or keep it focused and follow up with a separate PR?

I'd be very happy to have multi-level if we can establish a style decision on how to make sure higher orders of folder indentation are clearly distinguished. I can mock up a few examples soon to compare.

jeffvli:
Some options for multi-level trees off the top of my head:
1. Similar to how it's implemented in the [Mantine tree](https://mantine.dev/core/tree/#connecting-lines) component with connecting lines for each node. That way the tree structure is easily discernable, although you still run into the issue of multi-level nesting running out of space in the sidebar
2. Instead of using a tree view, you can use a navigation view. When clicking on a folder, instead of opening a tree, it would only show playlists that are part of that folder with a `..` button to go back to the parent. This approach is probably more scalable although I'm not sure about the UX tradeoff for having to navigate back and forth.

Best case scenario would probably to implement both approaches with an additional toggle although that's extra work.

shawnsquire:
@jeffvli : I like the thought process if we're okay with some settings. I think there is a lot of flexibility in letting the user choose how to make it work best for them. I made some changes to the code, so let me know thoughts.

image
New settings options.

"Single folder" is an option that works like it does now, where they are only one level deep. Allows for circumstances where the user doesn't need deep nesting and the cleanest interface.

**Navigation view**

image 
image
The lines above are breadcrumbs to track where the user is in the hierarchy. Folders appear above playlists as clickable items, putting only items in that folder in the sidebar view.

**Tree View**

image
Customizable line color and indentation, but can show unlimited folders.

Confirmed that manual sorting of playlists still works, although you currently can not sort folders. They will always display before playlists and in alphabetical order. You also can not drag playlists into other folders when in tree view. I figure that would be a good "phase 2" if we like this direction. Let me know your thoughts.

jeffvli:
It's looking really good!

Some additional thoughts:

- I'm not sure how I feel about the design here for the `navigation` view in the cases of multi-level nesting. I think you could actually get away with replacing the `Playlists` accordion string with the name of the current-level folder as well as adding a back button to the left of it. That way you can get rid of the navigation tree as well as the folder name to make it more clean. Same thing for the shared playlists section.

image

shawnsquire:
Got it. I removed the breadcrumb chain in favor of a cleaner header, but wanted to confirm before pushing. Here are some example screenshots.

image

image

image


I did like how you could jump an arbitrary number of levels up with the breadcrumb navigation, but I understand how the simplicity is better. Maybe if there's interest in the feature we could look at features like a long-press to jump up to an arbitrary level (like how long pressing Back on most browsers will show a history).

jeffvli:
Did you forget to push your latest changes?

shawnsquire:
New version should be up now. 😀

svetixoxo:
Thank you guys for your effort!

PR #2024, opened 2026-05-11
Adds a hotkey to navigate to the currently playing song

nova77:
Adds a hotkey "show playing song" (`mod+l` by default) which jumps to the playing/selected song in the current view.

Note that this only works in "infinite" pagination, or on the the page where the track is being played.
Adding full support for pagination (i.e. jumping across pages) would require substantial more work. @jeffvli I'm happy to do it if you think it's a good idea, but I wanted to keep the PR as minimal as possible.

### DIFF
--- a/src/i18n/locales/en.json
+++ b/src/i18n/locales/en.json
@@ -873,6 +873,7 @@
         "hotkey_listPlayLast": "List play last",
         "hotkey_listPlayNext": "List play next",
         "hotkey_listPlayNow": "List play now",
+        "hotkey_listShowPlayingSong": "Show playing song in list",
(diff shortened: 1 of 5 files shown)


jeffvli:
> Adding full support for pagination (i.e. jumping across pages) would require substantial more work. @jeffvli I'm happy to do it if you think it's a good idea, but I wanted to keep the PR as minimal as possible.

I guess *you could*, but any solution to this would not be scalable, as it involves performing an infinite fetch to eventually find the page that the song is on.

I'm happy to merge this as-is.

nova77:
Yeah, I figured it would be very messy. Thanks for the feedback and feel free to merge at your convenience! :)

PR #2025, opened 2026-05-11
Highlight the playlist in the left panel on play

nova77:
When a song within a playlist is selected, highlights the playlist on the left panel, making it easier to identify in which playlist the current song is playing.

playlist_highlight

---

When a track from a non-playlist view is selected (e.g. artist/track/album/etc). The entry is cleared.


nova77:
Since #2017 has been merged (yay!), I'll need to sync, resolve the conflicts, and make sure it works on folded playlists as well.

nova77:
Okay, merged and verified it works with the new playlist folders.

jeffvli:
The implementation is a bit fragile as it doesn't fully match the queue with the playlist, as it is only watching for specific playback triggers. It can be confusing since queue deviations do not necessarily cancel out the highlight.

Examples of how this breaks:
1. Start playback from a playlist which sets the currentPlaylistContextId
2. Navigate to the album or song list
3. Drag an album or multiple songs into your queue
4. Playlist highlight is still active even though queue has now deviated from the playlist

You could resolve this by adding clear functions to all the other handlers that handle queue mutations, but that might end up being a bit messy as then the logic would end up in multiple places. I need to do some thinking about what approach would be best to refactor.

nova77:
Hi Jeff! Thanks for taking the time to look at my PR.
You're right: my first attempt touched the queue mutation handlers, but it quickly became messy.
I didn't consider queue deviations, though :(.

I'll be away for the next two weeks, hence I won't have much time to look into it so feel free to take over or leave it be. Any suggestion anyway is highly appreciated!

nova77:
Okay, I think I have an alternative approach which should hopefully address (most) of your concerns: each queue song is tagged with the playlist it was queued from, so that it can be used to highlight whenever necessary. This should work across all dragging methods.

The one gap is the shared `PlayAction` menu's append actions (e.g. from the menu Play/Next, or more specifically the append action for songs) whose `PlayActionProps` is very minimal and I wasn't sure we should "pollute" it by adding playlist tracking stuff. Note that the only side effect is that the "play next" is song doesn't  highlight the source. Once we move to the next one (from a playlist), it's will highlight it again.

What do you think? I could add the additional wiring in this PR or in a follow up. Or leave it be as an "edge" case that doesn't warrant the additional complexity.

svetixoxo:
Thank you! @nova77

src/renderer/features/player/context/player-context.tsx
```diff
@@ -187,9 +211,18 @@ export const PlayerProvider = ({ children }: { children: React.ReactNode }) => {
...
+            type: AddToQueueType,
+            playSongId?: string,
+            contextPlaylistId?: null | string,
+        ) => {
             const filters = useSettingsStore.getState().playback.filters;
             const filteredData = filterSongsByPlayerFilters(data, filters);
+            const resolvedContextId =
+                contextPlaylistId ??
+                (isReplaceQueueType(type) ? inferPlaylistContextFromUrl() : null);
+            const tagged = tagPlaylistContext(filteredData, resolvedContextId);
```
kgarner7: Rather than unconditionally call this, why not just override `filteredData` conditionally?

src/renderer/features/player/context/player-context.tsx
```diff
@@ -281,11 +314,17 @@ export const PlayerProvider = ({ children }: { children: React.ReactNode }) => {
                 const filters = useSettingsStore.getState().playback.filters;
                 const filteredSongs = filterSongsByPlayerFilters(sortedSongs, filters);
 
+                const explicitId =
+                    itemType === LibraryItem.PLAYLIST && id.length === 1 ? id[0] : null;
```
kgarner7, line 321: What happens if you queue up multiple playlists at the same time?

src/renderer/features/player/context/player-context.tsx
```diff
@@ -281,11 +314,17 @@ export const PlayerProvider = ({ children }: { children: React.ReactNode }) => {
...
+                const resolvedContextId =
+                    explicitId ?? (isReplaceQueueType(type) ? inferPlaylistContextFromUrl() : null);
+                const taggedSongs = tagPlaylistContext(filteredSongs, resolvedContextId);
```
kgarner7: Same note here, I would suggest conditionally reassigning filteredSongs

kgarner7 (changes requested):

src/renderer/features/player/context/player-context.tsx
nova77: Good point. Switched to `let` and tagged only if `resolvedContextId` is not null.

src/renderer/features/player/context/player-context.tsx
nova77: Well, songs from multiple playlists are flat-mapped in `fetchSongsByItemType`, so per-song origin is lost. Hence I deliberately highlight nothing rather than guess. Per-playlist tagging is possible, but it would be a much larger change.

src/renderer/features/player/context/player-context.tsx
nova77: Same as above with `filteredSongs`.

kgarner7:
Thanks!

PR #2030, opened 2026-05-13
feat: add 10-band graphic equalizer

jlamons:
## Summary

Adds a 10-band graphic equalizer accessible from the playerbar gear popover.
Works on both audio backends (mpv `af` filter chain and Web Audio
BiquadFilterNode chain) and persists with other playback settings. The Web
Audio backend also exposes its post-EQ signal to the visualizer, so the
spectrum tracks what the user actually hears.

## Why

This is one of the most-requested features for self-hosted music players in
the Subsonic / Navidrome / Jellyfin ecosystem, and Feishin already has the
audio plumbing for it on both backends — it just needed the DSP, state, and
UI wired in surgically.

## What's included

- 10-band ISO graphic EQ (31 Hz – 16 kHz), ±12 dB per band
- Manual pre-amp slider (±12 dB)
- Bypass / Enabled toggle
- 7 built-in presets (Flat, Rock, Pop, Jazz, Classical, Bass Boost, Vocal)
- User-savable named custom presets, with delete
- Mantine modal launched from the playerbar gear popover
- Both backends:
  - **mpv**: builds an `af` chain of FFmpeg `volume` + 10 `equalizer`
    filters, dispatched via the existing `mpvPlayer.setProperties` IPC
    with `requestAnimationFrame` coalescing. Clears `af` on bypass and on
    LOCAL→WEB switch.
  - **Web Audio**: inserts 10 `peaking` BiquadFilterNodes per source gain
    + a shared preamp GainNode between gains and destination. Uses
    target-specific disconnects so the visualizer analyser fan-out is
    preserved.
- Per-source unity-gain tap nodes exposed via `webAudio.visualizerInputs`
  so the visualizer reads post-EQ instead of pre-EQ.
- All visible strings via `t()`; English locale shipped (i18next-parser
  picks up the keys for other locales).
- Persistence via the existing settings store (Zod schema + persist
  middleware), so EQ settings round-trip through app restart.

## Implementation notes

- All new code lives under `src/renderer/features/player/equalizer/`. The
  upstream-file diff is **surgical**:
  - `settings.store.ts`: one new schema field + one default value
  - `audio-players.tsx`: one new `<EqualizerHook />` mount
  - `player-config.tsx`: one new option entry
  - `utils/get-visualizer-audio-nodes.ts`: prefer `visualizerInputs` when
    populated regardless of playback type (existing LOCAL+system-audio
    and WEB-only-gains behaviour preserved)
  - `i18n/locales/en.json`: new `equalizer` keys
- No changes under `src/main/`, `src/preload/`, or `src/remote/`.

## Out of scope

- Remote-control surface EQ (`src/remote/`)
- Auto preamp / clip detection
- Parametric EQ, variable Q, additional band counts
- Spectrum-analyzer overlay inside the EQ modal
- Sharing / importing presets

Happy to follow up on any of those in a separate PR if there's interest.

## Test plan

- [ ] Web backend: open gear → "Open…" under Equalizer, enable, slide bands → audible.
- [ ] Apply Bass Boost preset → audible bass.
- [ ] Toggle Enabled off → audio returns to flat in ~10 ms, no click.
- [ ] Save + delete a custom preset.
- [ ] Crossfade two tracks with EQ on → no glitches.
- [ ] mpv backend (if mpv installed): repeat.
- [ ] Backend switch mid-playback (LOCAL → WEB) → audio continues, mpv `af` cleared.
- [ ] Restart app → EQ state restored.
- [ ] Visualizer still works with EQ enabled, and visibly tracks band changes.

### DIFF
--- a/src/i18n/locales/en.json
+++ b/src/i18n/locales/en.json
@@ -209,6 +209,31 @@
         "trackWithCount_one": "{{count}} track",
         "trackWithCount_other": "{{count}} tracks"
     },
+    "equalizer": {
+        "bandLabelHz": "{{frequency}} Hz",
+        "bandLabelKHz": "{{frequency}} kHz",
+        "custom": "Custom",
+        "deletePresetLabel": "Delete preset",
+        "enabled": "Enabled",
+        "menuLabel": "Equalizer",
+        "openLabel": "Open…",
+        "preamp": "Pre-amp",
+        "preset": {
+            "bassBoost": "Bass Boost",
+            "classical": "Classical",
+            "flat": "Flat",
+            "jazz": "Jazz",
+            "pop": "Pop",
+            "rock": "Rock",
+            "vocal": "Vocal"
+        },
+        "presetLabel": "Preset",
+        "reset": "Reset",
+        "saveAs": "Save as preset…",
+        "saveDialogName": "Name",
+        "saveDialogTitle": "Save preset",
+        "title": "Equalizer"
+    },
     "error": {
         "apiRouteError": "Unable to route request",
         "audioDeviceFetchError": "An error occurred when trying to get audio devices",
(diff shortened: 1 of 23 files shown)


jlamons:
Closing this — I missed [#1972](https://github.com/jeffvli/feishin/pull/1972) when I started work and would not have opened this if I'd known. Apologies for the noise. Will leave #1972 to keebsxd and the maintainers.

PR #2043, opened 2026-05-19
feat: persist player timestamp

TiagoSimionato:
Hey, amazing app. I would like just to add this little detail, sometimes I close the player and when I open again the song restarts, but it would be great to continue where I left off. This should persist the song timestamp after closing the app, then restore the timestamp when it's open again

### DIFF
--- a/src/renderer/features/player/hooks/use-queue-restore.ts
+++ b/src/renderer/features/player/hooks/use-queue-restore.ts
@@ -1,6 +1,6 @@
 import { useMutation, useQueryClient } from '@tanstack/react-query';
 import { t } from 'i18next';
-import { useCallback } from 'react';
+import { useCallback, useEffect, useRef } from 'react';
 
 import { api } from '/@/renderer/api';
 import { usePlayerEvents } from '/@/renderer/features/player/audio-player/hooks/use-player-events';
@@ -9,10 +9,14 @@ import { songsQueries } from '/@/renderer/features/songs/api/songs-api';
 import {
     setTimestamp,
     useCurrentServerId,
+    usePlayerHydrated,
+    usePlayerSong,
+    usePlayerStatus,
     usePlayerStore,
     useTimestampStoreBase,
 } from '/@/renderer/store';
 import { toast } from '/@/shared/components/toast/toast';
+import { PlayerStatus } from '/@/shared/types/types';
 
 export const useQueueRestoreTimestamp = () => {
     const player = usePlayerStore();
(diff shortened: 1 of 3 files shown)


src/renderer/features/player/hooks/use-queue-restore.ts
```diff
@@ -37,6 +41,72 @@ export const QueueRestoreTimestampHook = () => {
...
+    );
+};
+
+export const InitialTimestampRestoreHook = () => {
+    useInitialTimestampRestore();
+    return null;
+};
+
 export const useSaveQueue = () => {
     const serverId = useCurrentServerId();
```
jeffvli, line 113: You should use `usePlayerActions()` here instead of subscribing to the entire player store.

src/renderer/features/player/hooks/use-queue-restore.ts
```diff
@@ -37,6 +41,72 @@ export const QueueRestoreTimestampHook = () => {
     return null;
 };
 
+export const useInitialTimestampRestore = () => {
+    const player = usePlayerStore();
```
jeffvli: Same as above.

src/renderer/features/player/hooks/use-queue-restore.ts
TiagoSimionato: replaced `usePlayerStore()` with `usePlayerActions()` at line 45, just confirming, should I replace somewhere else?

TiagoSimionato:
Quick question, @jeffvli is it okay to make this change like this or is it better to add this as a config option?

src/renderer/features/player/hooks/use-queue-restore.ts
```diff
@@ -9,10 +9,15 @@ import { songsQueries } from '/@/renderer/features/songs/api/songs-api';
...
+    usePlayerSong,
+    usePlayerStatus,
     usePlayerStore,
     useTimestampStoreBase,
 } from '/@/renderer/store';
 import { toast } from '/@/shared/components/toast/toast';
+import { PlayerStatus } from '/@/shared/types/types';
 
 export const useQueueRestoreTimestamp = () => {
     const player = usePlayerStore();
```
jeffvli: Sorry, this should be replaced with usePlayerActions as well.

jeffvli:
> Quick question, @jeffvli is it okay to make this change like this or is it better to add this as a config option?

I'm ok with this not requiring a config since this was technically the intended behavior, but had some issues in the past which prevented the timestamp restore from working.

src/renderer/features/player/hooks/use-queue-restore.ts
TiagoSimionato: thanks, changes made!

PR #2045, opened 2026-05-19
feat(artists): preserve artist detail scroll position on back navigation

### LINKED ISSUE #2042: [Feature]: Preserve artist detail scroll position when navigating back from album

- [x] Yes

Not server-specific

When browsing an artist with a long list of albums, opening an album and then navigating back returns to the top of the artist page. It would be helpful to return to the previous position instead, so browsing can continue smoothly.
### Requested behavior
- On back navigation (`POP`), restore the previous scroll position for that artist page history entry.
- On fresh navigations (`PUSH`/`REPLACE`), keep current behavior (start at top).
### Why this would help
For artists with many albums, users often open multiple albums in sequence. Returning to top each time requires repeated scrolling and makes exploration slower.
### Proposed approach
Use the existing scroll persistence pattern already used by list/discography views:
- persist offset keyed by `location.key`
- restore only when `useNavigationType() === 'POP'`
Scope can be limited to artist detail first (no global behavior change).

Happy to open a PR for this if it aligns with project direction.

SAY-5:
Closes #2042. The maintainer commented: "A PR would be appreciated!"

Adds a small `useNativeScrollPersist` hook that mirrors the existing `useItemListScrollPersist` pattern (persist by `location.key`, restore only on `POP`) and wires it into the album artist detail route. Scope is limited to that route per the issue; PUSH/REPLACE navigations still start at the top.

### DIFF
--- a/src/renderer/components/native-scroll-area/use-native-scroll-persist.ts
+++ b/src/renderer/components/native-scroll-area/use-native-scroll-persist.ts
@@ -1,0 +1,53 @@
+import { RefObject, useEffect, useLayoutEffect } from 'react';
+import { useLocation, useNavigationType } from 'react-router';
+
+import { useScrollStore } from '/@/renderer/store/scroll.store';
+
+interface UseNativeScrollPersistProps {
+    enabled: boolean;
+    scrollRef: RefObject<HTMLDivElement | null>;
+}
+
+// Persists vertical scroll offset for a NativeScrollArea, keyed by react-router
+// location.key. Restores the saved offset only on POP navigation; PUSH/REPLACE
+// continue to start at the top.
+export const useNativeScrollPersist = ({ enabled, scrollRef }: UseNativeScrollPersistProps) => {
+    const location = useLocation();
+    const navigationType = useNavigationType();
+    const setOffset = useScrollStore((s) => s.setOffset);
+    const getOffset = useScrollStore((s) => s.getOffset);
+
+    useLayoutEffect(() => {
+        if (!enabled) return;
+        if (navigationType !== 'POP') return;
+        const node = scrollRef.current;
+        if (!node) return;
+
+        const saved = getOffset(location.key);
+        if (typeof saved !== 'number') return;
+
+        const applyOffset = () => {
+            if (!scrollRef.current) return;
+            scrollRef.current.scrollTop = saved;
+        };
+
+        applyOffset();
+        const raf = requestAnimationFrame(applyOffset);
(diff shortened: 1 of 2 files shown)


src/renderer/features/artists/routes/album-artist-detail-route.tsx
```diff
@@ -28,6 +29,8 @@ const AlbumArtistDetailRouteContent = () => {
     const serverId = useCurrentServerId();
     const { artistBackground, artistBackgroundBlur } = useArtistBackground();
 
+    useNativeScrollPersist({ enabled: true, scrollRef: scrollAreaRef });
```
kgarner7, line 32: Unfortunately, the node that is selected here is _not_ the correct one. The node which actually sees the scroll event the _child_ of `scrollAreaRef.current` (`node.children[0]`, in the `use-native-scroll-persists.ts`)

src/renderer/components/native-scroll-area/use-native-scroll-persist.ts
```diff
@@ -0,0 +1,53 @@
...
+// continue to start at the top.
+export const useNativeScrollPersist = ({ enabled, scrollRef }: UseNativeScrollPersistProps) => {
+    const location = useLocation();
+    const navigationType = useNavigationType();
+    const setOffset = useScrollStore((s) => s.setOffset);
+    const getOffset = useScrollStore((s) => s.getOffset);
+
+    useLayoutEffect(() => {
+        if (!enabled) return;
+        if (navigationType !== 'POP') return;
```
kgarner7: These conditions can be combined into one if

kgarner7 (changes requested):
Thanks! With fixes, this does appear to work properly with modifications. Might be worth investigating why `scrollAreaRef` is not the right target for scroll events/scroll top.

src/renderer/features/artists/routes/album-artist-detail-route.tsx
SAY-5: You are right, OverlayScrollbars moves the content into a viewport child of the container, so the ref node itself never scrolls and the listener on it never fired. The hook now resolves scrollRef.current.children[0] for both saving and restoring (2a43a2a).

src/renderer/components/native-scroll-area/use-native-scroll-persist.ts
SAY-5: Combined in 2a43a2a, in both effects.

PR #2056, opened 2026-05-23
chore: support installing dependencies using pnpm v11

### LINKED ISSUE #2114: [Bug]: Does not compile anymore with new pnpm 11

- [x] Yes

### App Version

1.13.0

### Music Server and Version

Navidrome 0.60.3

Desktop Linux

### What happened?

Due to changes in default configs in pnpm, the creation of a 'pnpm-workspace.yaml' is needed to allow build scripts to do their things.
The file should be automatically generated when first run on pnpm 11, and then either doing ``pnpm approve-builds`` or editing the file works

### Steps to reproduce

1. Clone the feishin repo
2. Compile
3. Errors out

### Relevant log output

```shell
[ERR_PNPM_IGNORED_BUILDS] Ignored build scripts: abstract-socket@2.1.1, electron-winstaller@5.4.0, electron@41.7.0, esbuild@0.25.12, esbuild@0.27.7
```

oHaoDaio:
# Summary
The `pnpm` field in `package.json` is no longer used after pnpm 11, it ignore `pnpm.onlyBuiltDependencies` settings and causes Docker builds to fail in clean environments:

```dockerfile
# Dockerfile

# Install pnpm 11.x
RUN npm install -g pnpm

# Throws ERR_PNPM_IGNORED_BUILDS exception
RUN pnpm install
```

Based on the pnpm migration guide, this PR fixes the issue by:
- Moving the allowed build packages to the new build policy configuration in `pnpm-workspace.yaml`.
- Keeping the legacy `pnpm.onlyBuiltDependencies` in `package.json` for now so we still support backward compatibility.

# Reference
[pnpm | Migrating from v10 to v11](https://pnpm.io/migration)
[pnpm | Settings (pnpm-workspace.yaml) | allowbuilds](https://pnpm.io/settings#allowbuilds)

### DIFF
--- a/.github/workflows/publish-alpha.yml
+++ b/.github/workflows/publish-alpha.yml
@@ -45,7 +45,7 @@ jobs:
             - name: Install Node and PNPM
               uses: pnpm/action-setup@v4
               with:
-                  version: 10
+                  version: 11
 
             - name: Install dependencies
               run: pnpm install
@@ -130,7 +130,7 @@ jobs:
             - name: Install Node and PNPM
               uses: pnpm/action-setup@v4
               with:
-                  version: 10
+                  version: 11
 
             - name: Install dependencies
               run: pnpm install
--- a/.github/workflows/publish-beta.yml
+++ b/.github/workflows/publish-beta.yml
@@ -20,7 +20,7 @@ jobs:
             - name: Install Node and PNPM
               uses: pnpm/action-setup@v4
               with:
-                  version: 10
+                  version: 11
 
             - name: Install dependencies
               run: pnpm install
@@ -124,7 +124,7 @@ jobs:
             - name: Install Node and PNPM
               uses: pnpm/action-setup@v4
               with:
-                  version: 10
+                  version: 11
 
             - name: Install dependencies
               run: pnpm install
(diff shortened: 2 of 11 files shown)


tippfehlr:
Seeing that https://github.com/jeffvli/feishin/blob/development/package.json#L190 only lists electron and esbuild, I think this should also only allow those two.

Edit: nevermind, it needs all four to build.

jeffvli:
Can you also push the updated lockfile?

jeffvli:
You're going to have to update the workflow files as well since they pin pnpm v10.

oHaoDaio:
> Can you also push the updated lockfile?

Hi jeffvli,

I have pushed the updated lockfile using pnpm@11.

According to the failed GitHub Action, it looks like the new version of pnpm is ignoring the `resolutions` section in `package.json`.

Maybe we should copy these settings into the `overrides` section of `pnpm-workspace.yaml` so pnpm can pick them up and apply them to `pnpm-lock.yaml`.

jeffvli:
I just pinned the docker build to use pnpm 10 in the meantime. This is somewhat low priority for me atm, but will return to this in a bit.

jeffvli:
I'm ready to push these changes for pnpm 11.

1. I need you to rebase development onto your branch
2. Remove the `RUN corepack...` line which is resolving to pnpm 10 in the Dockerfile
3. Add `"packageManager": "pnpm@11.5.2",` to the package.json

jeffvli:
I ended up fast tracking it in https://github.com/jeffvli/feishin/commit/d23f7619ec240e8854798152b16be3f157d65de4

Issue #1238, opened 2025-11-02
[Feature]: Share link in discord rpc


xNovyz:
### I have already checked through the existing feature requests and found no duplicates

- [x] Yes

### Is this a server-specific feature?

Navidrome

### What do you want to be added?

Add an option to automatically generate a share link and add it in the discord presence

BotBlake:
Hey, not stale Mr. GitHub Actions.
I'd love to have this as well! (sorry for useless offtopic comment)

PR #2061, opened 2026-05-26
Exclude Feature Requests from stale workflow

BotBlake:
Adds the enhancement label to the stale action exemption list.

Feature requests generally do not become "stale" on their own, since they often remain valid until someone decides to implement them. This change prevents enhancement requests from being automatically marked as stale or closed due to inactivity.

The Idea for this came from https://github.com/jeffvli/feishin/issues/1238#issuecomment-4362966275

Love Feishin, keep it up :heart_hands:

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -42,6 +42,6 @@ jobs:
 
 
                   stale-issue-label: 'stale'
-                  exempt-issue-labels: 'keep,security'
+                  exempt-issue-labels: 'keep,security,enhancement'
                   stale-pr-label: 'stale'
                   exempt-pr-labels: 'keep,security'


jeffvli:
Funnily enough, if you look at the [commit history](https://github.com/BotBlake/feishin/commits/f622c52bccf1ed5fc318f954e3e6c68478c2cf66/.github/workflows/stale.yml) for that file, enhancements *were* actually exempted from being stale originally, but I removed that exmeption about a year ago. The problem was that enhancements were starting to pile up, especially very old ones which I was never going to touch.

I'll have to think about if I want to re-add this, because it starts to become unmaintainable if there are hundreds of years-old or inactive feature requests from users who no longer use the app.

PR #2064, opened 2026-05-26
fix: set RESOURCES_PATH relative to app.getAppPath()

tippfehlr:
I’m packaging feishin for Arch Linux, and the tray icon is not shown. We run feishin
with system electron, so process.resourcesPath points to the resources
folder of the system electron, not the one from feishin.

```
Error: Failed to load image from path '/usr/lib/electron39/resources/assets/icons/icon.png'
    at createTray (/usr/lib/feishin/app.asar/src/main/index.ts:410)
```

This PR fixes the issue for Arch, I haven’t tested whether the app.getAppPath()
approach works on other platforms.

### DIFF
--- a/src/main/index.ts
+++ b/src/main/index.ts
@@ -335,7 +335,7 @@ if (isDevelopment) {
 }
 
 const RESOURCES_PATH = app.isPackaged
-    ? path.join(process.resourcesPath, 'assets')
+    ? path.join(path.dirname(app.getAppPath()), 'assets')
     : path.join(__dirname, '../../assets');
 
 const getAssetPath = (...paths: string[]): string => {


Slashbunny:
I'm getting the following error on Arch Linux with version 1.12 that is using this patch:

```
Error: Failed to load image from path '/usr/lib/feishin/app.asar/assets/icons/icon.png'
    at createTray (/usr/lib/feishin/app.asar/src/main/index.ts:415:12)
    at /usr/lib/feishin/app.asar/src/main/index.ts:1050:17
```

I can look at the binary file (`/usr/lib/feishin/app.asar`) and see that the patch has been applied, so I don't know what's going on. I assume `app.getAppPath()` is doing something strange for bundled applications. Perhaps `path.dirname(app.getPath('exe'))` would accomplish what was intended here?

https://www.electronjs.org/docs/latest/api/app#appgetpathname

Slashbunny:
`app.getAppPath()` is supposed to return a directory, so I assume it's returning the path within the bundle to where `index.ts` is, so something like `/usr/lib/feishin/app.asar/src` and then `path.dirname()` on that returns `/usr/lib/feishin/app.asar`. (note: I have zero experience with Electron development, so there might be a preferred way to do this instead of `app.getPath('exe')`)

tippfehlr:
That is weird, for me feishin 1.12.0-1 does not print an error and the icon is shown in the statusbar. I’m using sway with swaybar, what’s your setup?

I uploaded a modified app.asar to https://pkgbuild.com/~tippfehlr/feishin/, which just adds

```js
console.log('isDevelopment:', isDevelopment);
console.log('getAppPath():', require$$0.app.getAppPath());
console.log("getAppPath('assets'):", require$$0.app.getAppPath('assets'));
console.log("getAppPath('exe'):", require$$0.app.getAppPath('exe'));
console.log("getAppPath('modules'):", require$$0.app.getAppPath('modules'));
```

If your issue persists, could you replace /usr/lib/feishin/app.asar with this one and paste the output?

For me, it’s
```
isDevelopment: false
getAppPath(): /usr/lib/feishin/app.asar
getAppPath('assets'): /usr/lib/feishin/app.asar
getAppPath('exe'): /usr/lib/feishin/app.asar
getAppPath('modules'): /usr/lib/feishin/app.asar
```

Slashbunny:
I am just using a stock Plasma setup from `extra` that was updated with `-Syu` yesterday. Here is the output with your patched version:

```$ feishin
isDevelopment: false
getAppPath(): /usr/lib/feishin/app.asar
getAppPath('assets'): /usr/lib/feishin/app.asar
getAppPath('exe'): /usr/lib/feishin/app.asar
getAppPath('modules'): /usr/lib/feishin/app.asar
Error: Failed to load image from path '/usr/lib/feishin/app.asar/assets/icons/icon.png'
    at createTray (/usr/lib/feishin/app.asar/src/main/index.ts:421:13)
    at /usr/lib/feishin/app.asar/src/main/index.ts:1057:21
```

Slashbunny:
Well I identified the difference between our setups. I have `electron41` installed and it's using that by default, per `ps` output. 

If I instead run this command: `electron39 /usr/lib/feishin/app.asar`, I no longer get the tray icon error and the icon shows up and works in the plasma system tray. 

```$ electron39 /usr/lib/feishin/app.asar 
isDevelopment: false
getAppPath(): /usr/lib/feishin/app.asar
getAppPath('assets'): /usr/lib/feishin/app.asar
getAppPath('exe'): /usr/lib/feishin/app.asar
getAppPath('modules'): /usr/lib/feishin/app.asar
```

This still doesn't make any sense😵‍💫 Is `path.dirname()` behaving differently between versions?

tippfehlr:
That is weird. 

```
> pacman -Q electron41 feishin
electron41 41.7.0-1
feishin 1.12.0-1
```

Could you also post the output of `pacman -Q electron41 feishin`?

I also updated the testing app.asar to include more debug prints, maybe try that as well.

Slashbunny:
For some reason I thought the package was depending on a different version (`obsidian` is requiring `39`). It's still a different version on my PC, which explains the difference in behavior:

```
$ pacman -Q electron41 feishin
electron41-bin 41.7.0-1
feishin 1.12.0-1
```

I replaced `electron41-bin` with `electron41` and the tray icon works (I assume something on AUR incorrectly required `-bin` at some point). 

To reproduce the bug with the binary distribution of `electron41`:

```sh
mkdir '/tmp/electron41-bin' && cd '/tmp/electron41-bin' && curl -LO 'https://github.com/electron/electron/releases/download/v41.7.0/electron-v41.7.0-linux-x64.zip' && unzip 'electron-v41.7.0-linux-x64.zip' && ./electron /usr/lib/feishin/app.asar
```

tippfehlr:
Your reproducer works, but it’s strange that (with the new app.asar) the debug print shows

```js
path.join(path.dirname(getAppPath()), "assets"): /usr/lib/feishin/assets
```

I don’t know what else could be different.

But if it works with electron41, I’d leave it at that :)

Slashbunny:
On Arch, it works with the `electron41` package, but if there are other systems/platforms that use the upstream binary electron version, they may have the same issue. 

I tried the latest upstream, `42.2.0` (https://github.com/electron/electron/releases/download/v42.2.0/electron-v42.2.0-linux-x64.zip), and then an older version, `39.1.2` (https://github.com/electron/electron/releases/download/v39.1.2/electron-v39.1.2-linux-x64.zip), and got the same bug, so it seems consistent, even if it doesn't make much sense.

tippfehlr:
@Slashbunny I was a little surprised by the merge, and tested `pnpm dev` and `pnpm run package:linux`, but both work (show the icon). @jeffvli thanks!

PR #2081, opened 2026-05-29
Feature: Add sleep timer for end of current album

Overdrivendev:
Adds an 'End of current album' option to the sleep timers. A feature I've wished for in a music player for a long time. Seemed simple enough to add in this case, it uses the same basic flow as the 'End of current song' mode, but just keeps track of the albumId of the song that was active when the mode was selected. I also added in code to disable this new mode when Shuffle is active, as it would behave exactly as end of current song would in that case. 

Tested the electron build with a jellyfin server.

### DIFF
--- a/src/renderer/features/player/components/sleep-timer-button.tsx
+++ b/src/renderer/features/player/components/sleep-timer-button.tsx
@@ -3,7 +3,11 @@ import { useTranslation } from 'react-i18next';
 
 import { usePlayerEvents } from '/@/renderer/features/player/audio-player/hooks/use-player-events';
 import { usePlayer } from '/@/renderer/features/player/context/player-context';
-import { usePlayerStatus, usePlayerStoreBase } from '/@/renderer/store/player.store';
+import {
+    usePlayerShuffle,
+    usePlayerStatus,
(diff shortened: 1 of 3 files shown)


src/renderer/features/player/components/sleep-timer-button.tsx
```diff
@@ -50,12 +55,38 @@ function formatRemaining(totalSeconds: number): string {
...
-    const { cancelTimer, setRemaining } = useSleepTimerActions();
+    const { cancelTimer, setRemaining, setTargetAlbumId } = useSleepTimerActions();
     const { mediaPause } = usePlayer();
 
     const mediaPauseRef = useRef(mediaPause);
     mediaPauseRef.current = mediaPause;
 
+    // End of album mode. Set the pauseOnNextSongEnd flag whenever the current track
+    // is the last one of the target album.
+    const evaluteEndOfAlbum = useCallback(() => {
```
jeffvli: Small typo here: evaluate is spelled incorrectly.

Overdrivendev: Ah and I thought I caught them all, fixed.

PR #2082, opened 2026-05-30
Fix playback hotkeys by preventing browser default

york9675:
## Problem
Playback hotkeys could trigger browser default behavior (for example, spacebar was not reliably triggering play/pause, and the page would scroll instead) instead of only controlling playback.

## Fix
Update playback hotkey handling to always call `preventDefault` for every playback hotkey.

### DIFF
--- a/src/renderer/features/player/hooks/use-playback-hotkeys.ts
+++ b/src/renderer/features/player/hooks/use-playback-hotkeys.ts
@@ -13,11 +13,19 @@ export const usePlaybackHotkeys = () => {
         const bindingHandlers: Array<{
             binding: (typeof bindings)[keyof typeof bindings];
             handler: () => void;
+            options?: {
+                preventDefault?: boolean;
+                usePhysicalKeys?: boolean;
+            };
         }> = [
             { binding: bindings.next, handler: () => player.mediaNext() },
             { binding: bindings.pause, handler: () => player.mediaPause() },
             { binding: bindings.play, handler: () => player.mediaPlay() },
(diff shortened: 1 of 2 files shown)


kgarner7:
I understand that space is typically the hotkey for toggling play/pause, but is there _ever_ a reason why _any_ of the hotkeys should not prevent browser default?
If I were to assign space to say, play, this would still happen.

york9675:
Thanks for the review and the great point!

Since the immediate problem was the spacebar causing the page to scroll instead of reliably toggling playback, I initially set `preventDefault` just for the play/pause binding.

However, to your point, it would probably be better to update the logic to automatically call `preventDefault()` for any action as long as the triggered key is Space key.
If you think this is a good approach, I’d be happy to try to implement it in this PR. Alternatively, let me know if you have another preferred way to handle this!

kgarner7:
I think it would make sense to prevent default for _any_ hotkey, _regardless_ of the actual keys pressed. Aside from space, tab is another key (not sure _why_ someone would do it, but it counts) that could cause issues (and it's possible other keys might as well). So, I think it'd probably just make sense to always preventDefault?

kgarner7:
Sorry for the ask here, but I wonder, would it make sense to just make _every_ hotkey (the `withPhysicalKeys` function) prevent default?

york9675:
> would it make sense to just make _every_ hotkey (the `withPhysicalKeys` function) prevent default?

Thanks for the point! Now it will prevent browser default for all hotkey.

kgarner7:
Thanks!

PR #2084, opened 2026-05-30
fix(player): step correctly when seeking rapidly

### LINKED ISSUE #2083: [Bug]: Rapidly seeking forward/backward repeatedly seeks to the same time
### Prerequisites

- [x] I have checked the existing (open and closed) bug reports and found no duplicates.

### App Version

1.12.1 (also present in the current `development` branch, commit `def1b1e`)

### Music Server and Version

Server-agnostic — this is purely client-side seek logic and reproduces regardless of backend (Navidrome/Jellyfin/Subsonic) and of the audio player (web / MPV / wavesurfer).

Desktop Linux (the seek logic is platform-independent)

### Current Behavior

When seeking forward/backward by rapidly pressing the seek shortcut, the position gets "stuck" and only one seek effectively registers per ~half-second window.

Example with a 5s skip step, starting at 0:30:
1. Press seek-back twice in quick succession → time lands at 0:25 (expected 0:20).
2. Keep mashing seek-back → it keeps re-seeking to 0:25 on every press instead of stepping further back.
3. After ~0.5–2s it jumps to 0:20 on the next press, and the cycle repeats.

### Expected Behavior

Each press should step by the configured skip offset regardless of how fast the shortcut is pressed (0:30 → 0:25 → 0:20 → 0:15 …).

### Steps to reproduce

1. Play a track and let it reach, say, 0:30.
2. Rapidly press the seek-backward (or seek-forward) shortcut several times within ~1 second.
3. Observe that the time repeatedly snaps to the same value instead of stepping on each press.

### Possible Solution

`mediaSkipBackward` / `mediaSkipForward` in `src/renderer/store/player.store.ts` compute the new target relative to `useTimestampStoreBase.getState().timestamp`. That timestamp store is only refreshed by a ~500ms `setInterval` engine poll (`web-player.tsx`, `mpv-player.tsx`, `wavesurfer-player.tsx`), so rapid presses within that window all read the same stale position and recompute the same target. Updating the timestamp store immediately when a seek is issued makes subsequent presses compute from the new position; the poll then just reconciles to the same value. (Will open a PR soon.)

### Relevant log output

```shell
N/A
```

fiso64:
Fixes #2083

## Problem

Rapidly pressing the seek shortcut repeatedly seeks to the same time instead of stepping. `mediaSkipBackward` / `mediaSkipForward` in `player.store.ts` compute the target relative to `useTimestampStoreBase`, which is only updated by the ~500ms engine poll (`web-player.tsx`, `mpv-player.tsx`, `wavesurfer-player.tsx`). Within that window every press reads the same stale position and recomputes the same target, so the time appears stuck for ~0.5–2s before advancing.

## Fix

Update the timestamp store immediately when a seek is issued, so subsequent presses compute from the new position. The engine poll then just reconciles to the same value. Also applied to `mediaSeekToTimestamp` so absolute seeks (slider / keyboard) reflect in the UI without the poll lag.

## Testing

- `pnpm run lint` (typecheck + eslint + stylelint) — passes
- `pnpm run build` — passes
- Verified manually: seeking now steps correctly (e.g. 0:30 → 0:25 → 0:20 → 0:15) when the shortcut is mashed, on the web backend.

---

🤖 Developed with AI assistance (Claude).

### DIFF
--- a/src/renderer/store/player.store.ts
+++ b/src/renderer/store/player.store.ts
@@ -1185,8 +1185,8 @@ export const usePlayerStoreBase = createWithEqualityFn<PlayerState>()(
                     });
                 },
                 mediaSeekToTimestamp: (timestamp: number) => {
-                    // Optimistically reflect the new position immediately so the UI
-                    // (seek bar, lyrics) doesn't lag behind the ~500ms engine poll.
+                    // See mediaSkipBackward: update the timestamp store right away to
+                    // avoid the stale-read left by the ~500ms engine poll.
                     setTimestampStore(timestamp);
                     set((state) => {
                         state.player.seekToTimestamp = uniqueSeekToTimestamp(timestamp);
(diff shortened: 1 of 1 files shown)


kgarner7 (changes requested):
The comment for each of the changes is different. Please have the same comment (or, just have the other two refer to one) that is correct.

kgarner7:
Also of note, this does _not_ solve the issue for web player (where when paused, it will seek backwards but some state is not updated)

fiso64:
Fixed the comments in the latest commit. 

As for the web player, that looks like a separate issue.

kgarner7:
Thanks!

PR #2097, opened 2026-06-01
fix: preserve infinite list cache on component remount (fixes random sort reshuffling)

pedrovieira:
## Problem

When browsing artists/albums with `sortBy` set to **Random**, clicking into an item and navigating back causes the list to reshuffle — every item is in a different position. This happens because:

1. Navigating to a detail view completely unmounts the list route component
2. On remount, `useItemListInfiniteLoader` resets its internal `useRef` guard (`previousDataQueryKeyRef`), causing the reset `useEffect` to fire
3. All cached pages are cleared and re-fetched from the server with `sortBy: random`, which returns a different ordering each time

## Changes

**`src/shared/types/domain-types.ts`**

Added a shared `SortKeyRandom = 'random'` const that all four sort enums (`AlbumListSort`, `SongListSort`, `AlbumArtistListSort`, `ArtistListSort`) reference:

```ts
export const SortKeyRandom = 'random';

export enum AlbumListSort {
    ...
    RANDOM = SortKeyRandom,
    ...
}
```

**`src/renderer/components/item-list/helpers/item-list-infinite-loader.ts`**

Both guards are now scoped to random sort only (checked via `query.sortBy === SortKeyRandom`):

1. **`fetchPage`:** When random sort is active, skip network fetch if the page is already loaded — prevents re-fetching a different random ordering
2. **Reset `useEffect`:** When random sort is active, check if React Query already has cached data for `dataQueryKey` (survives unmount via stable key hashing). If data exists, skip the reset entirely

Non-random sorts are completely unaffected — both guards are skipped.

**`src/renderer/components/item-list/helpers/item-list-paginated-loader.ts`**

When random sort is active, `staleTime` and `gcTime` are extended to 10 minutes (from the default 15s), so the cached page survives navigation away and back without re-fetching. Explicit Refresh still works — `invalidateQueries()` forces re-fetch regardless of `staleTime`.

## Why this works

- React Query hashes query keys deterministically — loaded pages in the infinite loader survive unmount (within cache lifetime)
- Scroll position is already restored via `useScrollStore` on `POP` navigation
- The explicit Refresh button still works — it intentionally resets data and re-fetches
- Filter/sort changes produce a new query key, naturally triggering a fresh fetch

## Testing

- [x] Navigate to Artists with sort set to Random
- [x] Click an artist → detail view loads
- [x] Navigate back → the same random order is preserved
- [x] Click Refresh → a new random order is fetched
- [x] Switch sort to Name → normal deterministic order
- [x] Switch back to Random → new random order

*AI-assisted Pull Request*

### DIFF
--- a/src/renderer/components/item-list/helpers/item-list-paginated-loader.ts
+++ b/src/renderer/components/item-list/helpers/item-list-paginated-loader.ts
@@ -12,7 +12,7 @@ import { useListContext } from '/@/renderer/context/list-context';
 import { eventEmitter } from '/@/renderer/events/event-emitter';
 import { UserFavoriteEventPayload, UserRatingEventPayload } from '/@/renderer/events/events';
 import { getListRefreshMutationKey } from '/@/renderer/features/shared/components/list-refresh-button';
-import { LibraryItem } from '/@/shared/types/domain-types';
+import { LibraryItem, SortKeyRandom } from '/@/shared/types/domain-types';
 
 const getQueryKeyName = (itemType: LibraryItem): string => {
     switch (itemType) {
@@ -76,7 +76,7 @@ export const useItemListPaginatedLoader = ({
     const fetchRange = getFetchRange(currentPage, itemsPerPage);
     const startIndex = fetchRange.startIndex;
 
-    const isRandomSort = query?.sortBy === 'random';
+    const isRandomSort = query?.sortBy === SortKeyRandom;
 
     const queryParams = useMemo(
         () => ({
--- a/src/renderer/components/item-list/helpers/item-list-paginated-loader.ts
+++ b/src/renderer/components/item-list/helpers/item-list-paginated-loader.ts
@@ -88,7 +88,7 @@ export const useItemListPaginatedLoader = ({
     );
 
     const { data } = useQuery({
-        gcTime: isRandomSort ? 1000 * 60 * 60 : 1000 * 15,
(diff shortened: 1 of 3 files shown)


jeffvli:
I would prefer if this is restricted only to random sort, which you can determine by the `dataQueryKey`.

You may also want to handle this case on the paginated loader.

pedrovieira:
@jeffvli thanks for the feedback, should be fixed now

src/renderer/components/item-list/helpers/item-list-paginated-loader.ts
```diff
@@ -97,7 +99,7 @@ export const useItemListPaginatedLoader = ({
             return result;
         },
         queryKey: queryKeys[getQueryKeyName(itemType)].list(serverId, queryParams),
-        staleTime: 1000 * 15,
+        staleTime: isRandomSort ? 1000 * 60 * 10 : 1000 * 15,
```
jeffvli, line 102: I see that you increased the gcTime and staleTIme if isRandomSort, which makes sense so that you don't lose the cached data for 10 minutes if you don't navigate back.

This should probably also be included for the infinite loader as well since it's not configured there and the defaults use:
```        
gcTime: 1000 * 20, // 20 seconds
staleTime: 1000 * 10, // 10 seconds
```

pedrovieira: that's a good point. I also added them to the infinite loader and kept the `1000 * 15` as the default, same as the paginated loader.

PR #2102, opened 2026-06-01
Fix: Use getCoverArt instead of getAlbumInfo2

Quinnca77:
# What does this fix?

I personally use Navidrome, and embed my cover arts directly into the files. I then instruct Navidrome to use these embeds as the first priority for viewing the cover art, rather than the associated album image. This does not interface well with Feishin, as Feishin asks for the album image directly on the album id. Rather than doing this, we could ask the cover art on the song itself and the server should handle automatically whether it should fetch embed or album, depending on how the user has configured the server.

# Verification steps

- [x] I have ran linting.
- [x] There doesn't seem to be any tests, so I haven't ran any.
- [x] Checked for myself with a song on Navidrome of which the embedded cover art is different from the album cover art that this works.

# Additional notes

I would like to say that I have been enjoying Feishin a lot and do not plan to return to any other platform after experiencing this one. Thank you all for making such a great app!
Second, this is my first open-source contribution. I do not have much experience working in big codebases like these. Feel free to let me know what I could've done better :)


kgarner7:
No. The reason why this is not implemented is that `getCoverArt` requires including credentials to access.

Quinnca77:
Aha, I did not think of this when implementing this. Would it be better to give the user the ability to configure a proxy URL? They could then configure this proxy to return an image link that does not contain any credentials. If you have any better ideas please let me know as well :)

kgarner7:
I suppose you could do something like the ND discord RPC plugin and upload it to some temporary service (https://github.com/navidrome/discord-rich-presence-plugin/blob/main/coverart.go). I believe a Feishin fork (I can't find it currently) also supported a custom image proxy.
If you have better ideas, feel free to discuss in the discussions (or if you're in Discord, come by there!)

The reason _specifically_ for `getAlbumInfo2`, is that ND behavior returns a public URL that doesn't have user credentials embedded, whereas `getCoverArt` requires you to share what is effectively your username/password.

(Sorry I'm grumpy)

PR #2109, opened 2026-06-03
feat: add jukebox player engine for server-side playback

mannansainidev:
## Summary
Adds a Jukebox player engine that enables server-side playback via the Subsonic `jukeboxControl` API.

This allows Feishin to control playback on the server (e.g. a tablet or remote device running Navidrome) instead of streaming audio locally.

## Changes
- Adds `PlayerType.JUKEBOX` to the player type enum
- Adds `JukeboxPlayerEngine` component that communicates with the server via `jukeboxControl` API
- Adds `JukeboxPlayer` wrapper component
- Wires jukebox into `audio-players.tsx` alongside Web and MPV engines
- Adds Jukebox option to the playback settings dropdown

## How it works
- Play/pause/seek/volume all call the `jukeboxControl` API endpoint
- Polls the server every 1 second to update the progress bar and detect track end
- Uses the existing server credential from the auth store — no extra configuration needed

## Testing
Tested with Navidrome. Select Jukebox in Settings → Playback → Audio Player and audio plays on the server device.


mannansainidev:
sorry im kinda new

mannansainidev:
some bugs like music resets if u switch mid music idk why

src/renderer/features/player/audio-player/engine/jukebox-player-engine.tsx
```diff
@@ -0,0 +1,165 @@
...
+        if (!serverUrl || !credential) return null;
+        const url = `${serverUrl}/rest/jukeboxControl?${credential}&v=1.16.1&c=Feishin&f=json&action=${action}${extra ? '&' + extra : ''}`;
+        try {
+            const r = await fetch(url);
+            const d = await r.json();
+            return d['subsonic-response'];
+        } catch {
+            return null;
+        }
+    };
```
jeffvli: Instead of directly calling the Subsonic API endpoints within the player, this should instead be typed and added as a  `subsonic-api.ts` contract and then implemented as an item in the interfaced ControllerEndpoint.

In addition, the app does not use the native `fetch` API, but instead uses the `axios` library to handle http requests.

src/renderer/features/player/audio-player/engine/jukebox-player-engine.tsx
```diff
@@ -0,0 +1,165 @@
...
+
+    // 1. Safe Track Transitioning
+    useEffect(() => {
+        const syncTrack = async () => {
+            if (!currentTrackId) {
+                await callApi('stop');
```
jeffvli: In addition to refactoring into the app fetching layer, you should create a new hook which uses a react-query mutation (similar to how other mutations are implemented in the app) which allows us to track the state of these http requests and also perform error handling.

This applies to all the existing invocations of `callApi`.

src/renderer/features/player/audio-player/engine/jukebox-player-engine.tsx
```diff
@@ -0,0 +1,165 @@
...
+
+                // Small delay to ensure Navidrome registration settles completely
+                setTimeout(() => {
+                    isChangingTrackRef.current = false;
+                }, 300);
+            }
+        };
+
+        syncTrack();
+        // eslint-disable-next-line react-hooks/exhaustive-deps
```
jeffvli, line 295: What is the purpose of all of the `// eslint-disable-next-line react-hooks/exhaustive-deps` for the useEffects?

mannansainidev: > What is the purpose of all of the `// eslint-disable-next-line react-hooks/exhaustive-deps` for the useEffects?

It prevents ESLint from forcing us to add variables (like playerStatus or callApi) to the dependency array, which would otherwise trigger api loops or incorrectly restartthe player whenever the user pauses/resumes a song.

mannansainidev:
Fixed: Linting

jeffvli:
Tested the branch. There seems to be an issue with the Jukebox player spawning a stale mpv player instance which is causing duplicated playback with the non-current song.

mannansainidev:
yo uh jeff can u check this out

src/renderer/features/settings/components/playback/audio-settings.tsx
```diff
@@ -90,23 +91,38 @@ export const AudioSettings = memo(() => {
...
     const status = usePlayerStatus();
     const playbackType = usePlaybackType();
+    
+    // Cast the store hook result to handle the structural list mismatch smoothly
+    const { servers, activeServerId } = useServerList() as any;
+    
+    // Ensure servers is treated safely as an array for the lookup
+    const serversArray = Array.isArray(servers) ? servers : Object.values(servers || {});
+    const currentServer = serversArray.find((s: any) => s.id === activeServerId);
+    const isJukeboxSupported = !!(currentServer as any)?.features?.[ServerFeature.JUKEBOX];
```
jeffvli: Use the existing useCurrentServer hook to get the current server.
Use the existing hasFeature utility function to determine if isJukeboxSupported.

mannansainidev: alright

mannansainidev:
done 
i think

PR #2110, opened 2026-06-03
fix: prevent lyrics desyncing due to scroll issues

iiPythonx:
Prevents lyric autoscrolling lagging behind current lyric progress by improving the source detection used for scrolling. Smooth scrolling is now tracked based on scroll activity rather than timeout, stopping it from being detected as a manual user-induced scroll.

Everything seems to work from my preliminary testing, although further testing is welcomed.


jeffvli:
Can you provide instructions on how to reproduce the issue you were experiencing?

pyxfluff:
> Can you provide instructions on how to reproduce the issue you were experiencing?

Mostly just play a song with fast moving lyrics (rap god by eminem works well). Eventually they should become desynced and frequently do for myself.

iiPythonx:
Reproducing requires nothing but patience:

https://github.com/user-attachments/assets/f054c2a6-6af5-4ff6-aa4a-6f678e61168e

It will also happen more then just once, but this is just the first one I recorded.

Issue #1066, opened 2025-08-12
[Feature]: Ability to download albums and artists discography


Norde:
### I have already checked through the existing feature requests and found no duplicates

- [x] Yes

### Is this a server-specific feature?

Jellyfin

### What do you want to be added?

Hello there,

Thanks to https://github.com/jeffvli/feishin/commit/c4677a63f6e90aac6b430df795277698ea3bf3fa, it is possible to download individual song.

It would be great to be able to download a complete album or even an artist's entire discography.

(As Jellyfin doesn't offer this function natively, it might be a bit tricky).

ullmannJan:
In a first step one could implement the download functionality for multi-selections of songs. That would only require a for loop to download all selected songs individually and then save them into a directory

RayChimera:
The issue is not stale, it still is very much wanted! :3 (aka bump)

PR #2129, opened 2026-06-06
feat(downloads): offline downloads for songs, albums and playlists

gossip31:
## Summary

Adds an offline download system to the Electron build. Songs, albums and playlists can be downloaded via the existing context menu or via a download button in album/playlist detail headers, and play back from disk when available.

## Design

**Main process** owns a queue (3 concurrent) backed by a manifest under `{userData}/downloads`. Files stream to `<serverId>/<songId>.<ext>` via electron `net.request`; failed jobs retry up to twice on a 4 s backoff before surfacing as terminal failures.

A new `feishin-local://` protocol scheme serves songs back to the renderer. The handler streams via `createReadStream` (no whole-file reads) and rejects any manifest entry whose resolved path escapes the downloads root. The scheme is registered with `corsEnabled`, `secure`, `standard`, `stream`, `supportFetchAPI` and `bypassCSP`; CSP allows `media-src` and `connect-src` for `feishin-local:`.

**Player integration**: `useSongUrl` prefers the local URL when a song is in the manifest. `getSongUrl` returns the absolute file path so the MPV engine can play it directly.

**Renderer state**: a zustand store mirrors the manifest. `useDownloadBridge` hydrates on mount, reconciles, and listens to `downloads-progress` events.

## UI

- Floating downloads panel (bottom-right) with per-song progress (% / bytes / speed / ETA) plus an aggregate progress bar tracked against a session total so the bar fills monotonically across a playlist.
- `DOWNLOAD_STATUS` table column rendering a green check, a ring progress, a queued clock or a failed glyph per row with hover tooltips. Wired into song / playlist-song / queue-song / album-detail / full-screen / side-queue tables.
- Album and playlist detail headers gain a download button that shows `N/M` progress and disables itself mid-download.
- `/downloads` route lists downloaded songs grouped by album or artist with search, sort, multi-select delete, per-group delete, per-row hover delete and a top-N storage breakdown bar.
- Toasts coalesce per session: one debounced start notice and one summary on drain (`N songs downloaded` / `Download failed — title (+M)` / `N downloaded · M failed`). `(will retry)` failures are suppressed.

## Migration

Settings version `v28 → v29` (guard is loose through `v30`) injects the new sidebar entry and the `DOWNLOAD_STATUS` column into existing persisted column layouts. Default `albumDetail` enabled-columns also includes the new column for fresh installs.

## i18n

New keys live under `common.*`, `setting.*` and `page.sidebar.*` in `en.json`. Other locales fall back to English at runtime until translated.

## Known limitations / follow-ups

- Electron build only. The web build hides the UI via `isElectron()` and falls back to the existing OS-download path for single songs.
- Subsonic credentials live in the download URL and are passed at enqueue time. Long-queued jobs will fail if credentials rotate before they start; the auto-retry covers transient cases, not credential rotation.
- Grouping in the `/downloads` route is case-sensitive on album/artist strings.
- Localisation only covers `en.json`; the `i18next` extractor config in this repo writes to a non-existent path, so keys were added manually.

## Test plan

- [ ] Right-click a song, an album and a playlist → download enqueues, panel shows progress, files land in `{userData}/downloads/<serverId>/`.
- [ ] Disable the network → downloaded songs still play.
- [ ] Restart the app → manifest re-hydrates, downloaded songs still play.
- [ ] Cancel a mid-flight download → `.part` is removed, no manifest entry.
- [ ] Delete a downloaded song from the `/downloads` page (single, group, multi-select) → file removed from disk, manifest updated, badge clears.
- [ ] Change the downloads folder from settings → new downloads land in the new path.
- [ ] Verify against Subsonic, Navidrome and Jellyfin backends.
- [ ] Verify MPV playback as well as the web audio player.
- [ ] Verify web build still loads and the row indicator stays hidden.


gossip31:
Addresses #1066

jeffvli:
Thanks for the PR. This might take me a bit of time to review!

PR #2133, opened 2026-06-07
chore: upgrade depdencencies

kgarner7:
Run `pnpm update` and remove the `react-router` forced resolution. 
.-. apparently tanstack type changes everywhere (everything is `unknown` now???)


kgarner7:
Downgrading react-tanstack because that introduced too many type changes for me to be willing to deal with. For some reason, it looks like eslint is now complaining about `use-server-authenticated` (it takes issue with the recursion).

src/renderer/hooks/use-server-authenticated.ts
```diff
@@ -97,7 +97,7 @@ export const useServerAuthenticated = () => {
                         isAdmin: userInfo.isAdmin,
                     });
 
-                    // Fetch and update server version and features
+                    // Fetch and update server version and featuresauthenticateServer
```
jeffvli: ?

PR #2135, opened 2026-06-08
fix(sharing): copy share link from within the user gesture (Firefox/Safari clipboard)

elushaX:
## Problem

The **Share** button never copies the link to the clipboard in **Firefox and Safari**.

The share URL only exists *after* the create-share request resolves, and the code called `navigator.clipboard.writeText(shareUrl)` from the mutation's `onSuccess` callback — i.e. after an `await`, outside the originating click's user activation. Firefox/Safari reject clipboard writes that aren't tied to a user gesture:

> DOMException: Clipboard write was blocked due to lack of user activation.

So nothing was copied, yet the success toast still claimed the link was copied. (Chrome is more lenient about this, which is why it appeared to work there.)

## Fix

Call `navigator.clipboard.write()` **synchronously within the submit gesture**, passing a `ClipboardItem` whose `text/plain` value is a `Promise` that resolves to the share URL once the share has been created. This preserves the user activation across the async create-share request — the spec'd pattern for copying asynchronously-produced data.

Fallbacks, in order:
- If `ClipboardItem` is unavailable, fall back to `writeText` after awaiting the URL.
- If the clipboard write is blocked/unavailable, fall back to the existing "click to open" toast.
- The success toast now reflects whether the copy actually succeeded (instead of always claiming success).

Only `src/renderer/features/sharing/components/share-item-context-modal.tsx` is touched.

## Testing

- `pnpm run build:web` builds clean.
- Verified against a live Navidrome server: the Share button now **copies the link in Firefox**, where it was previously blocked; Chromium still works.

---

🤖 This change was **AI-assisted** (Claude). Happy to adjust the approach or wording.

### DIFF
--- a/src/renderer/features/sharing/components/share-item-context-modal.tsx
+++ b/src/renderer/features/sharing/components/share-item-context-modal.tsx
@@ -43,8 +43,18 @@ export const ShareItemContextModal = ({
     });
 
     const handleSubmit = form.onSubmit(async (values) => {
-        shareItemMutation.mutate(
-            {
+        const canUseClipboard = Boolean(navigator.clipboard) && window.isSecureContext;
+
+        // The share URL only exists once the create request resolves. Calling
+        // navigator.clipboard.writeText() from that async callback runs outside
+        // the click's user activation, so Firefox/Safari reject it ("Clipboard
+        // write was blocked due to lack of user activation") and nothing is
+        // copied. Instead, call clipboard.write() synchronously within this
+        // gesture with a ClipboardItem whose value is a promise that resolves to
+        // the URL — this preserves the activation while the share is created.
+        // Falls back to writeText, then to the "click to open" toast.
+        const shareUrlPromise = shareItemMutation
+            .mutateAsync({
                 apiClientProps: { serverId: server?.id || '' },
                 body: {
                     description: values.description,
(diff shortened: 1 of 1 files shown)


jeffvli:
What environment are you basing this on?

I tested Firefox (Windows) on the latest stable build and I'm not having any issues copying to the clipboard. I'm unable to test on Safari though.

elushaX:
Environment
- OS: Linux (Arch)
- Browser: Firefox 150.0.1 — did not test Safari
- Feishin: self-hosted via Docker, served over HTTPS
- Backend: Navidrome

History behind the MR:
- I noticed that creating a share link copied nothing to my clipboard, so I tried on other clients and it worked there. I also recall web working at some point on Firefox (Windows or Linux, I can't remember which).
- I asked Claude to find the issue, and it produced a patch to the source that fixed it for me.
- I then asked it to package the patch into this MR and explain it.

Reproducing:
After your comment, I tested again: in a private window on the previous (unpatched) code it worked fine. But once I started playing with timing, I found that throttling the network makes the copy fail — load the site normally, then enable throttling in dev tools, then click to create a share link. I switched back to the MR version, did the same throttled test, and it copied correctly.

PS:
I didn't dig into the changes themselves — I just asked Claude to report this and prepare the MR.
But it seems that the timing matters and when I faced that issue i probably had bad internet connection.

Thank You for Your work BTW!

PR #2144, opened 2026-06-10
Fetch discord image from lastfm if showServerImage is disabled & we have a key

### LINKED ISSUE #2143: [Bug]: Discord presence image not working

- [x] Yes

### App Version

1.13.0

### Music Server and Version

Jellyfin 10.11.8

Desktop Windows

### What happened?

The presence image isnt working. Ideally if there is a last.fm key and Serve Discord Images from server is disabled then it should use last.fms api.

This is useful in a few situations including my one of the jellyfin server being a local home ip that isnt accessible by the internet meaning discord cannot use my images

### Steps to reproduce

1. enable discord rich presense
2. Add a last.fm api key
3. Ensure Serve images from server is off

### Relevant log output

```shell

```

ChaoticLeah:
Fixes #2143

If showServerImage  is disabled (which is the setting to say if discord should use the image from your server or not), and we have a lastfm key then pull from lastfm to get an image.

### DIFF
--- a/src/renderer/features/discord-rpc/use-discord-rpc.ts
+++ b/src/renderer/features/discord-rpc/use-discord-rpc.ts
@@ -15,7 +15,6 @@ import {
     DiscordLinkType,
     useAppStore,
     useDiscordSettings,
-    useGeneralSettings,
     useLastfmApiKey,
     usePlayerSong,
     usePlayerStore,
@@ -48,7 +47,6 @@ const truncate = (field: string) =>
 
 export const useDiscordRpc = () => {
     const discordSettings = useDiscordSettings();
-    const generalSettings = useGeneralSettings();
     const lastfmApiKey = useLastfmApiKey();
     const privateMode = useAppStore((state) => state.privateMode);
     const [lastUniqueId, setlastUniqueId] = useState('');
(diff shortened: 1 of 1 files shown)


src/renderer/features/discord-rpc/use-discord-rpc.ts
```diff
@@ -285,7 +285,7 @@ export const useDiscordRpc = () => {
             }
 
             if (
-                activity.largeImageKey === undefined &&
+                (activity.largeImageKey === undefined || !discordSettings.showServerImage) &&
```
jeffvli, line 288: > The presence image isnt working. Ideally if there is a last.fm key and Serve Discord Images from server is disabled then it should use last.fms api.
>
> This is useful in a few situations including my one of the jellyfin server being a local home ip that isnt accessible by the internet meaning discord cannot use my images

I'm confused, since this if block condition already triggers regardless if `discordSettings.showServerImage` is enabled or not. 

The previous if block will trigger only if `discordSettings.showServerImage` is true, which sets activity.largeImageKey. If you already have it disabled, then this block already triggers correctly as the largeImageKey will be undefined.

ChaoticLeah: I had a whole other thing written but I reread what was said here and it makes sense what you are saying. Perhaps something else was going wrong?

Regardless if "showServerImage" isnt used for anything else perhaps this workaround is good for debugging?

kgarner7: `showServerImage` is used: https://github.com/jeffvli/feishin/pull/2144/changes#diff-6ca7571d5825ad8af846ea8c0c475406cef3882d346b36d4547e37692bbde126L260

ChaoticLeah: Yeah? Isnt this the intended use case of this toggle? Perhaps im missing something

PR #2152, opened 2026-06-13
feat: Give the user the option to specify an image proxy for Discord RPC

Quinnca77:
# What does this fix?

I personally use Navidrome, and embed my cover arts directly into the files. I then instruct Navidrome to use these embeds as the first priority for viewing the cover art, rather than the associated album image. This does not interface well with Feishin Discord RPC, as Feishin asks for the album image directly on the album id. Rather than doing this, we can do the following:
1. Use getCoverArt endpoint on music server
2. Get image and send it to an image proxy
3. Image proxy sends back a link pointing to the image resource, which is accessible from the internet
4. Send link to Discord RPC

This feature works out of the box if `Image proxy` is selected for the server type, and the link `https://uguu.se/upload` is given.

# Verification steps

- [x] I have ran linting.
- [x] Checked for myself with a song on Navidrome of which the embedded cover art is different from the album cover art that this works. Tested with `uguu.se`.

# Things I'm not too sure about

- How the image proxy returns the image link can differ per service, and this is not configurable at the moment.
- The way the configuration is changed for `Serve discord images from server` and its text.

# Additional notes

I would like to say that I have been enjoying Feishin a lot and do not plan to return to any other platform after experiencing this one. Thank you all for making such a great app!
Second, this is my first open-source contribution. I do not have much experience working in big codebases like these. Feel free to let me know what I could've done better :)


Quinnca77:
I discarded the renamed fields in different locales as I felt this change was unwarranted

src/main/features/core/discord-rpc/index.ts
```diff
@@ -39,6 +39,27 @@ const quit = () => {
...
+    });
+
+    if (!fileUploadResponse.ok) {
+        return null;
+    }
+
+    const json = await fileUploadResponse.json();
+
+    // Location within json for uguu.se
+    return json.files[0].url;
```
jeffvli: This is too rigid, as it will only work for uguu.se or any other uploaders that follow that response format.

You would need to allow the user to customize the uploader (maybe get some inspiration from [ShareX custom uploader](https://getsharex.com/docs/custom-uploader#sxcu-file)?).

The other option is to specifically add pre-configured file share hosts (e.g. imgur, uguu.se, etc) and interface it so that they can be plugged into the same interface.

Quinnca77: I would like to give the user the option to configure this as much as possible. Supporting everything ShareX would support sounds overkill though, so what if we add config fields for:
- The endpoint 
- The file field name in the POST request
- The json path in the server response for feishin to extract the URL from 

Does this sound good?

jeffvli: That works.

Quinnca77:
The upload is now more configurable :)
I added the above bullet points as configuration options that only appear when the user has selected to use an image proxy for Discord RPC.

jeffvli:
Unfortunately even with the three configuration options it's still too rigid. I did a quick analysis (using AI) to determine what formats other image hosts expect for the uploader. 

I think the proper approach here would be to do what I mentioned in my previous comment:
  > The other option is to specifically add pre-configured file share hosts (e.g. imgur, uguu.se, etc) and interface it so that they can be plugged into the same interface.

You would create an interface which accepts an uploader callback as well as a group of custom settings which are depicted as an object but parsed to the interface.

---


### Works (or should) with current config

| Host | Config | Notes |
|---|---|---|
| **uguu.se** (default) | URL `https://uguu.se/upload`, field `files[]`, path `files.0.url` | Designed for this |
| **Self-hosted uguu / pomf clones** | Same pattern if response is JSON with a URL field | Best bet for “bring your own” |
| **ImgBB** | URL `https://api.imgbb.com/1/upload?key=YOUR_KEY`, field `image`, path `data.url` (or `data.display_url`) | Key in **query string** avoids needing an extra form field |
| **tmpfiles.org** | URL `https://tmpfiles.org/api/v1/upload`, field `file`, path `data.url` | Upload works, but response URL is often a **viewer** page (`tmpfiles.org/...`), not a direct image URL Discord may want |

### Does **not** work without extending the client

| Host | Blocker |
|---|---|
| **catbox.moe** | Needs extra form field `reqtype=fileupload`; response is **plain text** URL, not JSON |
| **litterbox.catbox.moe** | Needs `reqtype` + `time`; plain text response |
| **0x0.st** | Multipart `file` is fine, but response is **plain text**, not JSON |
| **Pixeldrain** | Upload can be multipart `file`, but JSON returns an **`id`**, not a full URL — you’d need a URL template like `https://pixeldrain.com/api/file/{id}` |
| **Imgur** | Needs `Authorization: Client-ID …` header |
| **Freeimage.host** (Chevereto-style) | Needs at least `key` + `action=upload` as extra fields (unless those can live entirely in the query string — not guaranteed) |
| **transfer.sh / similar** | Often **PUT** with raw body, not multipart POST |

Quinnca77:
So you'd like me to provide some pre-defined working settings for different image hosts. That's okay, but which image hosts would you like me to implement then? Your first table depicts the hosts that already work with the current implementation.

I could then show different config settings per selected image host, depending on what they need. For example, for `litterbox.catbox.moe` we could give the user the option to configure the `time` field, and pick a default setting if they choose not to configure it.

jeffvli:
Just implement your preferred host to start (uguu.se).

> I could then show different config settings per selected image host, depending on what they need. For example, for litterbox.catbox.moe we could give the user the option to configure the time field, and pick a default setting if they choose not to configure it.

Exactly.

Quinnca77:
I won't have a lot of time the coming days so it'll take a while for me to implement, but when I do I'll also include `litterbox.catbox.moe` as a PoC.

Thank you for the quick review :)

Quinnca77:
I implemented interfaces for `litterbox.catbox.moe` and `uguu.se`. They each specify their own defaults. I kept the URL as a user config item because I felt that users should have the option to point it somewhere else, for example at their own APIs that conform to the same spec. The URLs for each image proxy have their defaults point to the URL you'd expect, so users can simply choose an image proxy while not configuring anything, and it should work for them.

Adding another image proxy should be as easy as exposing the user-configurable config items in Feishin UI, defining its interface, passing the user config items through IPC and implementing its upload function where it uses the config items.

Could you review the code once more? If you have any questions, feel free to ask me through my Discord at the same username.

src/i18n/locales/en.json
```diff
@@ -883,8 +884,15 @@
         "discordPausedStatus": "Show rich presence when paused",
         "discordRichPresence": "{{discord}} rich presence",
         "discordRichPresence_description": "Enable playback status in {{discord}} rich presence. Image keys are: {{icon}}, {{playing}}, and {{paused}}",
-        "discordServeImage": "Serve {{discord}} images from server",
-        "discordServeImage_description": "Share cover art for {{discord}} rich presence from server itself, only available for Jellyfin and Navidrome. {{discord}} uses a bot to fetch images, so your server must be reachable from the public internet",
+        "discordServerType": "Serve {{discord}} images from server",
+        "discordServerType_description": "Choose where to share cover art for {{discord}} rich presence. From music server itself is only available for Jellyfin and Navidrome. {{discord}} uses a bot to fetch images, so your server must be reachable from the public internet",
+        "discordServerType_none": "$t(common.none)",
+        "discordServerType_uguu": "uguu.se",
+        "discordServerType_litterbox": "litterbox.catbox.moe",
```
jeffvli, line 891: URLs do not need a translation string.

src/renderer/features/settings/components/window/discord-settings.tsx
```diff
@@ -213,27 +225,104 @@ export const DiscordSettings = memo(() => {
...
+                    }}
+                    value={settings.litterboxTime}
+                />
+            ),
+            description: t('setting.discordLitterboxTimeField', {
+                context: 'description',
+            }),
+            isHidden: !isElectron() || settings.serverType !== DiscordServerType.LITTERBOX,
+            title: t('setting.discordLitterboxTimeField'),
+        },
```
jeffvli, line 325: I'd rather not allow specific settings like this as it adds additional overhead especially if additional uploaders are added in the future each with their own specific settings.

I would prefer you refactor this into a JSON setting and input which allows the user to free-form a settings config (JSON) with a link to external documentation added to the repo docs as markdown for each uploader.

Formatted as `<uploader>_<setting>`, you could then add an infinite number of settings that could then be passed to the uploader to parse and perform actions on. Now we can add specific settings per uploader without having to add settings UI and translation strings to each one.
```json
{
  "litterbox_time": "72h"
}
```

src/renderer/store/settings.store.ts
```diff
@@ -1238,10 +1249,12 @@ const initialState: SettingsState = {
         clientId: '1165957668758900787',
         displayType: DiscordDisplayType.FEISHIN,
         enabled: false,
+        imageProxyServerLink: 'https://uguu.se/upload',
```
jeffvli, line 1252: > They each specify their own defaults. I kept the URL as a user config item because I felt that users should have the option to point it somewhere else, for example at their own APIs that conform to the same spec. The URLs for each image proxy have their defaults point to the URL you'd expect, so users can simply choose an image proxy while not configuring anything, and it should work for them.

From what you said, this is treated as an override? In this case shouldn't the default setting be empty?

src/renderer/features/discord-rpc/use-discord-rpc.ts
```diff
@@ -256,8 +264,83 @@ export const useDiscordRpc = () => {
...
+                                        break;
+                                    }
+                                    config = {
+                                        time: discordSettings.litterboxTime as LitterboxImageProxyConfig['time'],
+                                    };
+                                    break;
+                                case DiscordServerType.UGUU: {
+                                    config = {};
+                                    break;
+                                }
```
jeffvli, line 317: In addition to the JSON settings change, the raw config should be passed to the uploader which handles parsing and consuming the config rather than at the discord rpc hook level.

jeffvli (changes requested):

Issue #2139, opened 2026-06-09
[Feature]: Link "Album Group" column to the album/artist page


svetixoxo:
### I have already checked through the existing feature requests and found no duplicates

- [x] Yes

### Is this a server-specific feature?

Not server-specific

### What do you want to be added?

I really like the column `Album Group`, but unfortunately it is text only. Is it possible maybe, to link it to link the album name to the album page and maybe the artist name to the artist page too? Like in the columns of `album` and `artist`?

And would it be also possible to show maybe the `favorite` heart or (even better) the `rating` of the album too?

PR #2153, opened 2026-06-14
feat: album group has a config and can set the image size

nova77:
Added a new album group configuration which allows to define the image size of the album displayed in the album group.

album image cfg

album image larger

A couple notes:
1. If the album size (in pixels) is larger than the current configuration for image resolution (settings->Application->Image Resolution->Table), it will be updated to match it.
2. I expect more settings to be added for the album group, hence the choice of a toggle. E.g. displaying album favorite or rating as discussed in https://github.com/jeffvli/feishin/issues/2139.


src/renderer/store/settings.store.ts
```diff
@@ -2480,10 +2482,17 @@ export const useSettingsStore = createWithEqualityFn<SettingsSlice>()(
...
+                        state.general.albumGroupImageSize =
+                            initialState.general.albumGroupImageSize;
+                    }
+                }
+
                 return persistedState;
             },
             name: 'store_settings',
-            version: 28,
+            version: 29,
```
jeffvli: This does not need a migration as it is only a new property which will be merged automatically.

src/renderer/components/item-list/item-table-list/item-table-list-column.tsx
```diff
@@ -402,6 +435,106 @@ export function isLastInAlbumGroup(
...
+        props.getRowItem,
+        props.enableHeader,
+        props.data.length,
+    );
+
+    // Only clamp when the row was actually grown to fit the image.
+    if (albumImageSize <= groupRowCount * baseHeight) return null;
+
+    return baseHeight;
+}
```
jeffvli, line 521: You should probably also be adding extra height based on the padding used in the table rows otherwise the album groups are clamped together.

src/renderer/store/settings.store.ts
nova77: ha, good to know! Reverted! :)

src/renderer/components/item-list/item-table-list/item-table-list-column.tsx
nova77: Of course. `var(--theme-spacing-xs)` hits again! :D
Okay, This should look a bit nicer now.

Issue #1958, opened 2026-04-15
[Feature]: Tag editing dialogue


coltercarson:
### I have already checked through the existing feature requests and found no duplicates

- [x] Yes

### Is this a server-specific feature?

Not server-specific

### What do you want to be added?

Correcting file metadata (genre, artist, album, etc.) currently requires leaving Navidrome/Feishin and using a separate tag editor. It would be great to have a tag editing dialogue accessible via right-click context menu on one or more songs, like the existing info dialogue but with editable fields. On desktop this could be handled via main process IPC using something like node-taglib-sharp, though web/Docker deployments would need a companion backend service with filesystem access.

I know write access to the filesystem is typically a no-go with music server related stuff, but it would be extremely helpful for me. I'd be happy to have a go at developing this - would you consider pulling the changes back into the repo if I got this working well?

coltercarson:
[Audiobookshelf](https://github.com/advplyr/audiobookshelf) has functionality similar to what I suggest - enabling tag editing, lookup from online tag sources, browser custom covers etc:
<img width="1025" height="888" alt="Image" src="https://github.com/user-attachments/assets/8fdbfcc0-1617-4b46-8fde-9d0eb9dd63de" />

jeffvli:
> I know write access to the filesystem is typically a no-go with music server related stuff, but it would be extremely helpful for me. I'd be happy to have a go at developing this - would you consider pulling the changes back into the repo if I got this working well?

I don't mind if you take a go at it.

nova77:
I'd be _very_ interested to see how this is going.

ChrisScott9456:
@coltercarson @jeffvli @nova77 

I have a PR for this ready for review:
https://github.com/jeffvli/feishin/pull/2159

Let me know what you think

ChrisScott9456:
Merged PR:
https://github.com/jeffvli/feishin/pull/2252

PR #2159, opened 2026-06-15
feat: Add batch metadata tag editor with artwork support

ChrisScott9456:
Feature for https://github.com/jeffvli/feishin/issues/1958

New metadata editor accessible via "Edit Metadata" in the song, album, and playlist song context menus. Supports editing one or more files at once.

Tags:
- Displays all metadata fields for the selected files
- Fields differing across files are shown as "(Multiple Values)"
- Fields can be added from a dropdown or removed individually
- Only modified fields are written back to disk

Artwork:
- Displays shared cover art across all selected files
- Supports setting new artwork via file picker or drag-and-drop
- Supports clearing existing artwork
- Shows a placeholder when selected files have different covers

Reading and writing:
- Tags are read via `music-metadata`, for performance and to preserve full ISO date strings
- Writes use `taglib-wasm` WASI in-place editing via setProperty, only touching fields that were explicitly changed
- Writability (permissions) of all files is checked before any writes begin
- Partial read failures show a warning but still allow editing

Other:
- Per-file batch progress is reported during reads and writes
- In-flight reads are cancelled when the modal is closed
- An optional server library rescan can be triggered after saving

---

Metadata Editor Menu (1 song selected)

image

Metadata Editor Menu (Multiple selected)

image

---

Artwork Tab

image

Artwork Tab (Multiple selected)

image

---

Loading Modal

image

The loading modal live updates the count as files are read

### DIFF
--- a/src/renderer/api/subsonic/subsonic-controller.ts
+++ b/src/renderer/api/subsonic/subsonic-controller.ts
@@ -2239,41 +2239,35 @@ export const SubsonicController: InternalControllerEndpoint = {
                 positionMs: Math.round(query.position ?? 0),
             };
 
-            const reportPlayback = (state: 'paused' | 'playing' | 'starting' | 'stopped') => {
-                return ssApiClient(apiClientProps).reportPlayback({
+            const reportPlayback = async (state: 'paused' | 'playing' | 'starting' | 'stopped') => {
+                const res = await ssApiClient(apiClientProps).reportPlayback({
                     query: {
                         ...defaultParams,
                         state,
                     },
                 });
-            };
-
-            const promises: Promise<any>[] = [];
-
-            switch (query.event) {
-                case 'pause':
-                    promises.push(reportPlayback('paused'));
-                    break;
-                case 'start':
-                    promises.push(reportPlayback('starting'));
-                    promises.push(reportPlayback('playing'));
-                    break;
-                case 'stop':
-                    promises.push(reportPlayback('stopped'));
-                    break;
-                case 'unpause':
-                    promises.push(reportPlayback('playing'));
-                    break;
-                default:
-                    break;
-            }
-
-            for (const promise of promises) {
-                const res = await promise;
 
                 if (res.status !== 200) {
                     throw new Error('Failed to report playback');
                 }
+            };
+
+            switch (query.event) {
+                case 'pause':
+                    await reportPlayback('paused');
+                    break;
+                case 'start':
+                    await reportPlayback('starting');
+                    await reportPlayback('playing');
+                    break;
+                case 'stop':
+                    await reportPlayback('stopped');
+                    break;
+                case 'unpause':
+                    await reportPlayback('playing');
+                    break;
+                default:
+                    break;
             }
 
             return null;


nova77:
Wow, this is really cool!
Quick Q: what kind of "library scan trigger" does it run? Is it the standard or full scan?

ChrisScott9456:
> Wow, this is really cool! Quick Q: what kind of "library scan trigger" does it run? Is it the standard or full scan?

Thanks!

I'm leaning on the available controller functionalities so I didn't confirm if they were full scans or not. I think for Jellyfin it will be. For subsonic, I'm unsure. In my experience I was able to save and refresh a few seconds later and the scan had the updates already reflected in the UI - so I assume it's not a full library scan.

coltercarson:
This looks fantastic - great work!

coltercarson:
This might not be the place to suggest, but a logical next step once tag read/write is solid would be searching online databases like MusicBrainz or Discogs (similar to how audiobookshelf allows you to 'match' local files to online metadata - see screenshot)
image

nova77:
> I'm leaning on the available controller functionalities so I didn't confirm if they were full scans or not. I think for Jellyfin it will be. For subsonic, I'm unsure. In my experience I was able to save and refresh a few seconds later and the scan had the updates already reflected in the UI - so I assume it's not a full library scan.

Got it, makes sense. Perhaps there's a way with subsonic/navidrome to specify which to run.

One more question: I'm not entirely clear on how taglib works. Does it rely on a service hosted in the same place the server is running, or is it only local (hence the client must run in the same location)?

ChrisScott9456:
> > I'm leaning on the available controller functionalities so I didn't confirm if they were full scans or not. I think for Jellyfin it will be. For subsonic, I'm unsure. In my experience I was able to save and refresh a few seconds later and the scan had the updates already reflected in the UI - so I assume it's not a full library scan.
> 
> Got it, makes sense. Perhaps there's a way with subsonic/navidrome to specify which to run.
> 
> One more question: I'm not entirely clear on how taglib works. Does it rely on a service hosted in the same place the server is running, or is it only local (hence the client must run in the same location)?

No, unfortunately this iteration of the metadata editor relies on you having local access from Feishin. If we wanted to implement it server side we would have to build a separate service to host. And I wasn't sure if that goes beyond the scope of what Feishin looks to accomplish.

But, if someone ever did go to hook that up, it should be fairly easy to do with all of this preliminary work done.

src/main/index.ts
```diff
@@ -39,6 +40,10 @@ import { autoUpdaterLogInterface, createLog, hotkeyToElectronAccelerator } from
 import { disableAutoUpdates, isLinux, isMacOS, isWindows } from '/@/main/env';
 import { PlayerRepeat, PlayerStatus, PlayerType, TitleTheme } from '/@/shared/types/types';
 
+if (isWindows() && typeof (global as any).require !== 'function') {
+    (global as any).require = createRequire(__filename);
+}
```
jeffvli: What is this being used for?

src/renderer/api/jellyfin/jellyfin-api.ts
```diff
@@ -365,6 +365,16 @@ export const contract = c.router({
             400: jfType._response.error,
         },
     },
+    startScan: {
+        body: z.null(),
+        method: 'POST',
+        path: 'Library/Refresh',
```
jeffvli: This should be using the [`/Items/{itemId}/Refresh`](https://api.jellyfin.org/#tag/Library/operation/RefreshItem) endpoint which can target specific music libraries as opposed to triggering a full refresh.

src/renderer/features/context-menu/actions/edit-metadata-action.tsx
```diff
@@ -0,0 +1,65 @@
...
+        openModal({
+            children: <SongEditModal songs={resolvedSongs} />,
+            size: 'xl',
+            styles: { body: { paddingBottom: 'var(--theme-spacing-xl)' } },
+            title:
+                trackCount > 1
+                    ? `${t('page.contextMenu.editMetadata')} (${trackCount} ${t('common.tracks', 'tracks')})`
+                    : t('page.contextMenu.editMetadata'),
+        });
+    }, [albumIds, server, songs, t]);
```
jeffvli: Instead of using openModal here, you should add a global context modal and use openContextModal instead. This will lete you move all the fetching logic inside of the component instead of performing an async fetch before we can open the modal.

src/renderer/features/tag-editor/utils/utils.ts
```diff
@@ -0,0 +1,29 @@
...
+
+export const filterTagSummary = (
+    tagSummary: Record<string, null | string>,
+): Record<string, null | string> => {
+    const filtered: Record<string, null | string> = {};
+    for (const [k, v] of Object.entries(tagSummary)) {
+        if (KNOWN_TAG_MAP.has(k)) filtered[k] = v;
+    }
+    return filtered;
+};
```
jeffvli: Only showing tags based on a hard-coded KNOWN_TAG_MAP is a pretty big UX issue, as it is misleading if written tags are hidden from the user. If taglib-wasm supports editing all tags despite not being part of the known tags that should be handled.

src/renderer/features/tag-editor/hooks/use-metadata-editor.ts
```diff
@@ -0,0 +1,342 @@
...
+    }, [artworkOp, browser, editedFields, removedKeys, rescan, resolvedSongs, server, t, utils]);
+
+    /** Known tags not yet present in `displayFields`, sorted alphabetically for the add-field dropdown. */
+    const availableToAdd = useMemo(
+        () =>
+            KNOWN_TAGS.filter((tag) => !(tag.key in displayFields))
+                .map((tag) => ({ label: tag.label, value: tag.key }))
+                .sort((a, b) => a.label.localeCompare(b.label)),
+        [displayFields],
+    );
```
jeffvli: In addition to the other comment, the user should be able to add any tag as opposed to only KNOWN_TAGS.

src/renderer/features/tag-editor/hooks/use-metadata-editor.ts
```diff
@@ -0,0 +1,342 @@
...
+
+    const [isLoading, setIsLoading] = useState(true);
+    const [loadProgress, setLoadProgress] = useState<BatchProgress | null>(null);
+    const [error, setError] = useState<null | string>(null);
+    const [readWarning, setReadWarning] = useState<null | string>(null);
+    const [resolvedSongs, setResolvedSongs] = useState<Song[]>([]);
+    const [rescan, setRescan] = useState(true);
+    const [isSaving, setIsSaving] = useState(false);
+    const [tagSummary, setTagSummary] = useState<Record<string, null | string>>({});
+    const [editedFields, setEditedFields] = useState<Record<string, string>>({});
```
jeffvli: editedFields should also be returned from this hook so that you can apply styles or an indicator if a field has been edited, otherwise it's not clear to the user which fields are actually being updated/applied when saving.

src/main/index.ts
ChrisScott9456: This was added to resolve issues resolving Windows drive paths. I ran into an issue I noticed while testing. My application is running on my C:/ drive, but my music is stored at a different drive path (W:/ in my case)

https://github.com/CharlesWiltgen/TagLib-Wasm/blob/main/src/runtime/unified-loader/module-loading.ts#L30

This line in taglib-wasm shows where it creates the function in global scope, and so we're exposing `require` at the global scope so it has access and can preopen (access all Windows drive paths) without issue.

src/renderer/api/jellyfin/jellyfin-api.ts
ChrisScott9456: Yeah I agree. I will update this.

src/renderer/features/context-menu/actions/edit-metadata-action.tsx
ChrisScott9456: Yeah, that makes sense. Will do.

src/renderer/features/tag-editor/utils/utils.ts
ChrisScott9456: Good callout, I agree. Will implement

src/renderer/features/tag-editor/hooks/use-metadata-editor.ts
ChrisScott9456: I'm still thinking through this one. I think ideally we should still keep a list of "actual" tags that the user can reference, but also allow them to write whatever tag they want, even if it's not a standard supported tag. (Such as if you wanted to create a tag called "ajskdflasdf", or whatever)

An Autocomplete that relies on the list of tags from taglib-wasm but accepts freeform input sounds like the best option.

Wanted to get your thoughts on that before I make those changes

ChrisScott9456: @jeffvli Thoughts?

jeffvli: I'm ok with that. Pre-set list of tags but allow user-entry.

ChrisScott9456:
@jeffvli All changes you requested have been merged in for you to review. Thanks!

src/renderer/features/tag-editor/hooks/use-metadata-editor.ts
```diff
@@ -0,0 +1,357 @@
...
+            const songs = (songsProp ?? []).filter((s) => s.path);
+
+            if (songs.length === 0) {
+                setError(t('page.itemDetail.noLocalSongs', 'No songs with local file paths found'));
+                setIsLoading(false);
+                return;
+            }
+
+            setResolvedSongs(songs);
+            const paths = songs.map((s) => s.path!);
```
jeffvli: This should be using the `resolveSongPath` utility function. 

I think you had it there previously but it got removed?

ChrisScott9456: > This should be using the `resolveSongPath` utility function.
> 
> I think you had it there previously but it got removed?

It wasn't there before, but I added it now.

jeffvli:
Oh shoot, I accidentally messed up your branch when I was trying to push my changes onto it.

Can you re-push your changes from local and re-open the PR?

ChrisScott9456:
> Oh shoot, I accidentally messed up your branch when I was trying to push my changes onto it.
> 
> Can you re-push your changes from local and re-open the PR?

No worries. I opened the PR. It also includes one final commit to update taglib-wasm to 1.5.1 that contains a fix for the windows drive paths code I had in the main/index.ts 

https://github.com/jeffvli/feishin/pull/2252

PR #2170, opened 2026-06-20
feat(player): add Listen Together synchronized playback

alsoGAMER:
## What

Adds an opt-in **Listen Together** feature: multiple Feishin clients join a room
and follow one host's playback (play / pause / seek / track / queue changes) in
near-real-time — a SyncPlay/SharePlay-style "listening party".

Coordination is handled by a small standalone open-source server,
[listen-together](https://git.alsogamer.com/alsog/listen-together), over a plain
WebSocket + JSON protocol. It is intentionally **not** a Navidrome plugin
(Navidrome plugins can only be outbound WebSocket clients and can't host a
coordinator), and authentication reuses each user's existing Subsonic
credentials — so it works with Navidrome and any Subsonic-compatible server.

## How

- `src/renderer/api/sync/sync-client.ts` — `SyncSocket`: connection lifecycle,
  Subsonic-credential auth, NTP-style clock sync, and reconnect with backoff.
  Modeled on the existing remote-control client (`src/remote/store/index.ts`).
- `src/renderer/store/sync.store.ts` — Zustand session store. Only the enable
  flag and server URL are persisted; the live socket is kept outside persisted
  state.
- `src/renderer/features/sync/use-sync-session.tsx` — applies inbound room state
  to the player through existing `player.store` actions (works for both the mpv
  and web engines) with clock-corrected drift handling, and emits the host's
  transport on local changes. Mounted as `<SyncSessionHook />` alongside the
  other player hooks in `audio-players.tsx`, mirroring `<RemoteHook />`.
- `src/renderer/features/sync/listen-together-control.tsx` — player-bar control
  (added to `right-controls.tsx`) to create/join rooms by code, view members,
  and hand off control. Uses the shared component wrappers.
- `src/shared/types/sync-types.ts` — wire types.

Single host holds transport authority with explicit pass-control; the host is
reassigned automatically if it leaves. Echo is prevented with an
applying-remote guard, and stale frames are dropped via a monotonic sequence
number.

## Privacy

Only track ids and position are synced — **stream URLs are never shared**. Each
client resolves songs and streams with its own credentials.

## Notes

- Fully opt-in: inert until enabled in the control, where the sync-server URL is
  also configured. No behavior changes for existing users.
- This PR is the client side only; the sync server is a separate project.

### DIFF
--- a/package.json
+++ b/package.json
@@ -32,6 +32,8 @@
         "postinstall": "electron-builder install-app-deps",
         "lint": "pnpm run typecheck && pnpm run lint-code && pnpm run lint-styles",
         "lint:fix": "pnpm run lint-code:fix && pnpm run lint-styles:fix",
+        "test": "vitest run",
+        "test:watch": "vitest",
         "lint-code": "eslint --max-warnings=0 --cache .",
         "lint-code:fix": "eslint --cache --fix .",
         "lint-styles": "stylelint --max-warnings=0 'src/**/*.{css,scss}'",
@@ -183,7 +185,8 @@
         "vite-plugin-conditional-import": "^0.1.7",
         "vite-plugin-dynamic-import": "^1.6.0",
         "vite-plugin-ejs": "^1.7.0",
-        "vite-plugin-pwa": "^1.3.0"
+        "vite-plugin-pwa": "^1.3.0",
+        "vitest": "^3.2.6"
     },
     "packageManager": "pnpm@11.5.2",
     "productName": "feishin"
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -343,6 +343,9 @@ importers:
       vite-plugin-pwa:
         specifier: ^1.3.0
         version: 1.3.0(vite@7.3.5(@types/node@24.13.1)(jiti@2.7.0)(sugarss@5.0.1(postcss@8.5.15))(terser@5.48.0))(workbox-build@7.3.0(@types/babel__core@7.20.5))(workbox-window@7.3.0)
+      vitest:
+        specifier: ^3.2.6
+        version: 3.2.6(@types/debug@4.1.13)(@types/node@24.13.1)(jiti@2.7.0)(sugarss@5.0.1(postcss@8.5.15))(terser@5.48.0)
 
 packages:
 
@@ -2113,8 +2116,14 @@ packages:
   '@types/cacheable-request@6.0.3':
     resolution: {integrity: sha512-IQ3EbTzGxIigb1I3qPZc1rWJnH0BmSKv5QYTalEwweFvyBDLSAe24zP0le/hyi7ecGfZVlIVAg4BZqb8WBwKqw==}
 
+  '@types/chai@5.2.3':
+    resolution: {integrity: sha512-Mw558oeA9fFbv65/y4mHtXDs9bPnFMZAL/jxdPFUpOHHIXX91mcgEHbS5Lahr+pwZFR8A7GQleRWeI6cGFC2UA==}
+
   '@types/debug@4.1.13':
     resolution: {integrity: sha512-KSVgmQmzMwPlmtljOomayoR89W4FynCAi3E8PPs7vmDVPe84hT+vGPKkJfThkmXs0x0jAaa9U8uW8bbfyS2fWw==}
+
+  '@types/deep-eql@4.0.2':
+    resolution: {integrity: sha512-c9h9dVVMigMPc4bwTvC5dxqtqJZwQPePsWjPlpSOnojbor6pGqdk541lfA7AqFQr5pB1BRdq0juY9db81BwyFw==}
 
   '@types/electron-localshortcut@3.1.3':
     resolution: {integrity: sha512-D+CRdDTRZ4/9UmcSaZ5qvW4uq2VyyVmqsH9cdNReB4CL6MSIgyhr9w2PKeNEb0J/ZS7db7irJM/+ZiA5uSQsLw==}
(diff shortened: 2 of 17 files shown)


alsoGAMER:
### Follow-up: sync hardening (`af9fe5d`)

Pushed a robustness pass over the Listen Together sync session, paired with the listen-together server **v1.0.1** release:

- **Clock offset keeps re-converging** over long sessions instead of locking onto an early min-RTT sample, so drift correction stays accurate as machine clocks drift.
- **Out-of-order protection:** each outbound transport is now stamped with a monotonic logical clock (`clientTimeMs`); the server drops frames that arrive out of order. Field is optional/back-compatible.
- **Followers skip re-resolving the whole queue** when it is already loaded (e.g. the host skips within the shared queue), avoiding a burst of `getSongDetail` calls on every track change.
- Replaced the magic echo-guard timeout with a named, documented constant.

Server side of these changes lives in the standalone sidecar at https://git.alsogamer.com/alsog/listen-together (v1.0.1).

alsoGAMER:
### Update — UX & resilience pass (commit `7866b91`)

A follow-up to the earlier correctness hardening, paired with a matching server release (**listen-together v1.0.2**).

**Client (this commit)**
- **Reconnect resilience:** if a room is reaped while a follower is briefly disconnected, the socket now surfaces a *room-closed* event so the session resets cleanly instead of leaving a dead room stuck in the UI.
- **Sync health indicator:** the player-bar popover now shows the measured clock offset plus an *in sync / correcting* badge for followers.
- **Pause-sync for followers:** a **Follow host** toggle lets a follower detach from the host's transport temporarily (e.g. to take a call) without leaving the room; it re-syncs to the latest state on resume.
- **Settings integration:** a *Listen Together* section under **Settings → Playback** (enable toggle + sync-server URL), backed by the same store as the popover.

**Server (listen-together v1.0.2)** — optional, back-compatible hardening for public hosting: room/member caps (`LT_MAX_ROOMS`, `LT_MAX_MEMBERS_PER_ROOM`), a browser `Origin` allowlist (`LT_ALLOWED_ORIGINS`), and an opt-in, token-gated `GET /stats` (`LT_STATS_TOKEN`).

alsoGAMER:
### Update — efficiency, i18n & tests (commits `f35d8ef`, `ab52bb1`, `57857e2`)

Paired with server release **listen-together v1.0.3**.

- **Queue diffing (`feat`).** The host now omits the queue from a transport when it's unchanged since the last send, so play/pause/seek no longer resend the full track-id list. The v1.0.3 server keeps the room's current queue when the field is omitted (wire- and back-compatible). Adds a shared `idsEqual` helper, reused for the follower's same-queue fast path.
- **i18n (`refactor`).** The feature's hardcoded UI strings (popover, the Settings → Playback section, and toasts) now go through `t()` / `en.json`, matching the rest of Feishin. Other locales fall back to English until translated.
- **Tests (`test`).** Adds a small **vitest** setup (`pnpm test`) and unit tests for the sync layer — `idsEqual` plus `SyncSocket`'s authenticate-on-open, clock-offset math, `clientTimeMs` stamping, pending-join failure handling, and reconnect-after-close — driven by a fake WebSocket, so no network or DOM is required.

**Server (listen-together v1.0.3)** — transport `queue` is now optional (omit = keep current), and the rarely-used password auth fallback is sent in the Subsonic `enc:` (hex) form instead of plaintext.

alsoGAMER:
### Update — follow-state fix (commit `443c441`)

Small correctness fix paired with server release **listen-together v1.0.5**.

**Client (`fix`).** A follower who toggled *Follow host* off and was then handed control stayed flagged as detached; after handing control back, they'd remain silently detached (the flag was only reset on join/leave/create, not on the host transition). It now resets when becoming host, so a control hand-back resumes following.

**Server (listen-together v1.0.5).** Connections that never authenticate are now dropped after a short deadline (~20s) — browsers auto-answer WS pings, so an unauthenticated socket could otherwise linger and accumulate. Complements the existing auth backoff and room/member caps.

alsoGAMER:
### Update — follower sync correctness fixes (commits `38c7c94`, `6f73106`)

Fixes for several playback-sync issues found in testing:

- **Rewind-to-start broadcast a stale position (`38c7c94`).** `mediaPrevious`'s restart branch (when >10s into a track) bumped `seekToTimestamp` without updating the timestamp store — the only seek path that didn't. The host's debounced transport then read the stale pre-restart position, so followers ignored the rewind. (General player fix, not sync-specific.)
- **Track-change seek raced the load (`6f73106`).** A follower switching within the shared queue seeked immediately after `mediaPlayByIndex`, before the new source was seekable, so it could start from 0. The post-track-change seek is now deferred (mirroring `use-queue-restore`).
- **Followers could drift / desync themselves (`6f73106`).** Added a follower reconciliation loop (while *following*): it re-projects the host's **live** position and re-applies track / play-pause / position every 500 ms. This (a) lands the correct position once a slow-to-load track becomes seekable — recomputed live, so you join where the host *is now* — and (b) makes a follower effectively read-only: a local play/pause/seek/skip is reverted so it can't desync the room. The **Follow host** toggle detaches for manual control.

**Known limitation:** large files on slow connections (e.g. big FLACs) have inherent catch-up latency on join/track-change — the follower converges as soon as the file is buffered enough to seek.

alsoGAMER:
### Update — Listen Together UI refinements (commit `03c779b`)

- **Sync badge no longer twitches.** It was deriving *in sync* / *correcting* from raw drift crossing a single threshold on every reconcile tick, but the player reports position coarsely (rounded / ~500 ms polled) so the drift jitters around the boundary. Replaced with a hysteretic `syncing` flag (enter "correcting" past 1.5 s of divergence, return to "in sync" under 0.6 s) — steady while following, one clean transition during real catch-up.
- **Control requests are now actionable.** A follower's *Request control* raises a modal for the host with **Give control** / **Cancel**, instead of a passive toast (backed by `pendingControlRequest` state, so it shows regardless of popover state).
- **Configuration is settings-only.** The enable toggle and sync-server URL are removed from the player-bar popover (now room-management only); it's hidden until the feature is enabled in **Settings → Playback**, and shows a hint to set the URL there when it's blank.

alsoGAMER:
### Update — jittered reconnect backoff (commit `2f67635`)

`SyncSocket` reconnect now uses **full jitter**: it waits a random time in `(0, backoff]` instead of the exact backoff, while the backoff ceiling keeps doubling to the 15s cap. Without this, a server restart drops every client at once and they all retry in lockstep — a thundering-herd reconnect that re-hammers the instance as it comes back. Unit test updated (jitter pinned via `Math.random` to keep it deterministic).

This is the optional reference-client side of the reconnect-jitter recommendation already documented server-side in `CLIENT_GUIDE.md`.

jeffvli:
Unfortunately this would not be a good candidate to be merged into base Feishin as it relies on a lesser known external service (I'm also assuming it's your own project?).

An alternative that I would accept would be utilizing a P2P framework like Trystero or PeerJS to handle playback and state synchronization between users.

alsoGAMER:
Oh I see. Yes it's my project, though it uses an open protocol that anyone can write clients for.

Would it be an issue for you if I maintain a Feishin fork with these changes?

> P2P framework

I don't think that'd be a desirable solution for everyone though (eg: people that can't open ports or have heavy cgnat applied, we'd need to use an external stun server anyway, for network punching)

jeffvli:
> Would it be an issue for you if I maintain a Feishin fork with these changes?

This would be the perfect situation for maintaining a fork. GL with your project!

Issue #1923, opened 2026-04-07
[Bug]: Can't play any song on PC if Transcoding is disabled


PaperCutBad:
### I have already checked through the existing bug reports and found no duplicates

- [x] Yes

### App Version

1.11.0

### Music Server and Version

Navidrome 0.61.1

### What local environments are you seeing the problem on?

Desktop Windows

### What happened?

Updated Navidrome to latest and feishin to latest version. Can't play any song if transcoding is disabled in feishin settings. Navidrome settings are at default. All music files are FLACs.

### Steps to reproduce

1. Play any song
2. Doesn't play-wavebar stays at zero
3. Enable transcoding leave transcoding option empty
4. Song plays

### Relevant log output

```shell

```

emanoellorenzzutti:
Same issue here, same solution!

jeffvli:
Can you disable transcoding again and check the browser console logs to see what errors you're seeing?

DaveKai:
I can confirm, nothing is playable, the songs get stuck at 0:00 unless if I enable transcoding

emanoellorenzzutti:
> Can you disable transcoding again and check the browser console logs to see what errors you're seeing?

<img width="1395" height="2445" alt="Image" src="https://github.com/user-attachments/assets/e428396b-84ea-4ad0-90b8-68a307cf95ae" />

127.0.0.1:4533/api/album?_end=20&_order=DESC&_sort=recently_added&_start=0&library_id=1&missing=false:1  Failed to load resource: net::ERR_CONNECTION_FAILED

create-storage.mjs:89 Uncaught RangeError: Maximum call stack size exceeded.
    at create-storage.mjs:89:18
    at Object.o (release-notes-modal.tsx:411:9)
    at Object.onClose (release-notes-modal.tsx:434:24)
    at IW (reducer.mjs:6:3)
    at ModalsProvider.mjs:54:9
    at Array.forEach (<anonymous>)
    at Object.closeAllModals (ModalsProvider.mjs:53:50)
    at i.<computed> (create-use-external-events.mjs:10:64)
    at gSe (create-use-external-events.mjs:5:10)
    at create-use-external-events.mjs:24:28

Hey, so I got this. Anything else I could do to help?

jeffvli:
[...]

Issue #187, opened 2023-07-20
Fails to play any song after windows sleeps


CrepeTF:
## Expected Behavior

Playback should work

## Current Behavior

When windows wakes up from sleep state and there's an attempt at playing a song. The song is not played and the program has to be restarted.

## Steps to Reproduce

1. Leave feishin in the background - song paused

2. Let windows enter sleep state

3. Wake up windows

4. Try to play a song

## Possible Solution (Not obligatory)

N/a

## Context

There's times when I wish to leave my pc idle but when I get back to it windows enters sleep mode to conserve energy. However upon coming back I always have to restart feishin if I want to play a song as playback would not work otherwise.

## Your Environment

-   Application version (e.g. v0.1.0) : 0.2.0
-   Operating System and version (e.g. Windows 10) : Windows 11  (22621)
-   Server and version (e.g. Navidrome v0.48.0) : Jellyfin 10.8.10
-   Node version (if developing locally) : N/a

kgarner7:
I tested latest build on w11 and I'm not able to reproduce this, is it still an issue? 
What version of MPV are you using, and where did you get it? (winget/sourceforge/choco?)
Does it also happen with web player?

CrepeTF:
Hey, quick update, it is still happening but I've actually had a look at the network requests for the issue. It was stuck waiting for server response

kgarner7:
Just to be clear, where is your Jellyfin server hosted? Is it also on windows? What specifically is Feishin waiting on?

kgarner7:
Since there have not been updates to this issue for quite a while, I'm going to go ahead and close it. If this is still an issue, please feel free to reopen.

PR #2172, opened 2026-06-21
fix: recover mpv playback after the OS resumes from sleep

### LINKED ISSUE #2171: [Bug]: Playback fails to start/resume after Windows wakes from sleep (MPV engine)

- [x] Yes

### App Version

1.13.0

### Music Server and Version

Navidrome 0.62.0 (1b46b977)

_No response_

### What happened?

Windows version: 25H2 (OS Build 26200.8655).

After my Windows desktop wakes from sleep, I'm unable to play any music in Feishin. The playback bar/progress stays stuck and nothing plays. This happens at least 90%+ of the time after a sleep/wake cycle, if not every time.

**Workaround:** Fully restarting the Feishin app (no Windows reboot needed) immediately fixes playback. This points to Feishin holding onto a stale connection or player/session state across the sleep cycle, rather than a deeper OS networking/driver issue.

**Server-side investigation (ruling this out before filing):**
I administer the Navidrome server myself and double-checked it wasn't a server/transcoding-decision issue similar to #1923:
- Navidrome version: 0.62.0 (already well past the 0.61.2 fix for the `getTranscodeDecision` fallback issue)
- `ffmpeg` and `ffprobe` both present and on PATH on the server (v6.1.1)
- Checked the Navidrome systemd journal across the period the issue occurred: no errors, no restarts, clean "Streaming file" / "Scrobbled" log entries from this same Feishin client both before and after the failure window
- Server has been continuously up for multiple days without interruption

Given the server shows no errors and keeps serving other clients fine, this looks like a Feishin-side (client) issue specific to the sleep/resume transition on Windows, not a server or ffmpeg/ffprobe problem.

I'm using the MPV playback engine in Feishin's settings (not Web Audio).

I haven't yet been able to capture Feishin's DevTools network tab (Ctrl+Shift+I) at the exact moment of failure. I'll attach that the next time I catch it before restarting, since the original reporter on #187 found their request was "stuck waiting for server response" in a similar scenario.

### Steps to reproduce

1. Have Feishin running on Windows desktop with music paused, connected to a Navidrome server, with MPV as the playback engine
2. Hibernate in Windows, leave for some time
3. Wake the desktop
4. Attempt to play any track (resume or start a new track)
5. Playback does not start; progress bar stays at the current position
6. Restarting Feishin resolves it until the next sleep/wake cycle

Note that an immediate hibernate -> wake cycle does not reproduce the issue, it only comes up when I hibernate overnight and come back to the computer

### Relevant log output

```shell

```

ryankupk:
## Problem

After Windows wakes from sleep/hibernate, Feishin can't play any track on the mpv engine — neither resuming a paused track nor starting a new one. The only fix is to fully restart Feishin (not Windows). See #2171 for the original report, including server-side investigation that ruled out Navidrome/ffmpeg/ffprobe as the cause (server logs clean, ffmpeg+ffprobe present, already past the unrelated fix in #1923).

There's also prior art in #187 (2023, against a Jellyfin server) with the same symptom — a request "stuck waiting for server response" — which went stale without resolution and is now locked.

## Root cause

Traced the mpv engine path (`src/main/features/core/player/index.ts`) and the underlying `node-mpv` fork it depends on (`jeffvli/Node-MPV`, pinned at commit `32b4d64`):

1. mpv is launched with no `--network-timeout` and no ffmpeg reconnect options (`DEFAULT_MPV_PARAMETERS`). A stream left open across a sleep/hibernate cycle sits on a TCP connection the OS silently drops; without these flags, mpv/ffmpeg blocks indefinitely on the dead socket instead of failing or reconnecting.
2. That hang isn't contained to the stream. Node-MPV's IPC layer makes every command — including `quit()` — a promise that only resolves when mpv sends a matching reply over the socket, with no timeout anywhere in that path:
   - `quit()` just `await`s `this.command("quit")` and nothing else: https://github.com/jeffvli/Node-MPV/blob/32b4d64395289ad710c41d481d2707a7acfc228f/lib/mpv/_startStop.js#L242-L253
   - the request wrapper backing every command resolves/rejects only from an IPC reply, with no timeout fallback: https://github.com/jeffvli/Node-MPV/blob/32b4d64395289ad710c41d481d2707a7acfc228f/lib/ipcInterface/ipcRequest.js#L11-L41

   So once mpv's command queue is wedged on the dead network read, `quit()` — and therefore Feishin's existing "restart mpv" path — can also hang forever. The only way out is killing the whole app, which matches the reports.
3. Electron's `powerMonitor` is only used in Feishin today to *prevent* sleep during playback (`powerSaveBlocker`). Nothing listens for *resume* to recover playback afterward.

## Fix

- Add `--network-timeout=10` and `--stream-lavf-o=reconnect=1,reconnect_streamed=1,reconnect_at_eof=1,reconnect_delay_max=5` to mpv's default parameters (skipped if the user already set either via the existing extra-parameters setting), so a stalled stream fails fast or reconnects instead of hanging forever.
- Make the existing `quit()` helper resilient to a wedged mpv process: race it against a 3s timeout and fall back to force-killing the process, so both manual and automatic recovery can't hang even if mpv itself is stuck.
- Listen for `powerMonitor`'s `resume` event in the main process and tell the renderer to reload mpv (reusing the existing `MPV_RELOAD` mechanism already used when playback settings change), so playback recovers automatically after sleep/wake instead of requiring a manual app restart.

## Testing

I want to be upfront: **I have not yet been able to verify this fix against a real Windows sleep/wake cycle.**

What's been verified:
- `pnpm run lint` (typecheck + eslint + stylelint — the exact CI check) passes clean.
- Manually traced the code paths above against the actual `node-mpv` source to confirm this targets the root cause rather than papering over the symptom.

Why conclusive testing will take a bit longer:
- I can only reproduce this after hibernating overnight and waking up later — an immediate manual sleep/wake from my desk doesn't reproduce it.
- A wake that *doesn't* fail isn't conclusive by itself (the original report is ~90%, not 100%, reproducible), so I'd want a few overnight hibernation cycles before calling it solid either way.

I'm continuing to test this in parallel on my own machine and will report back results. If you're comfortable reviewing the code and merging on that basis given how narrowly scoped the change is (default mpv parameters + a timeout fallback + a resume listener, no behavior change on the happy path), I'm fine with that — but didn't want this to merge under the assumption it was already verified end-to-end.

Closes #2171

### DIFF
--- a/src/main/features/core/player/index.ts
+++ b/src/main/features/core/player/index.ts
@@ -1,5 +1,5 @@
 import console from 'console';
-import { app, ipcMain } from 'electron';
+import { app, ipcMain, powerMonitor } from 'electron';
 import { access, rm } from 'fs/promises';
 import uniq from 'lodash/uniq';
 import MpvAPI from 'node-mpv';
@@ -83,6 +83,19 @@ const DEFAULT_MPV_PARAMETERS = (extraParameters?: string[]) => {
 
     if (!extraParameters?.some((param) => prefetchPlaylistParams.includes(param))) {
         parameters.push('--prefetch-playlist=yes');
+    }
+
+    // Without these, mpv/ffmpeg will block indefinitely on a dead TCP connection
+    // instead of failing or reconnecting. This commonly happens when the OS network
+    // adapter resets after the system wakes from sleep while a stream is open.
+    if (!extraParameters?.some((param) => param.startsWith('--network-timeout'))) {
+        parameters.push('--network-timeout=10');
+    }
+
+    if (!extraParameters?.some((param) => param.startsWith('--stream-lavf-o'))) {
+        parameters.push(
+            '--stream-lavf-o=reconnect=1,reconnect_streamed=1,reconnect_at_eof=1,reconnect_delay_max=5',
+        );
     }
 
     return parameters;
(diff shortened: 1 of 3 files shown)


ryankupk:
I left my computer hibernating with a Feishin client built from this branch for 2 days. After waking the computer from hibernation, playback worked perfectly fine without having to restart Feishin. Again, this isn't conclusive since the issue only arose for me at least ~90% of the time, but in this test at least, there was no issue.

I'll leave it hibernating for a few more days and check again, if it works well twice I feel confident that this change resolves the issue.

ryankupk:
Left it for another couple of days and had no issues with playback after waking from hibernation. I had the current release of Feishin open at the same time with music paused, and it *did* have a playback issue, so I think that these changes should fix it.

jeffvli:
I personally haven't had this issue but the changes make sense and don't seem harmful to merge in. Thanks.

PR #2161, opened 2026-06-17
feat: add Japanese Furigana support to lyrics display

york9675:
This PR introduces support for Furigana (pronunciation guides) in the lyrics display.

This is *highly* beneficial for users who want to sing along but have difficulty reading Kanji.

## New Dependencies
Added the `kuroshiro` and `kuroshiro-analyzer-kuromoji` packages to handle Japanese text processing.

### Working well on both fullscreen and queue side panel
<table>
  <tr>
    <td width="50%">
      Fullscreen View
    </td>
    <td width="50%">
      Queue Side Panel View
    </td>
  </tr>
</table>

### New toggle in both main settings panel and lyric settings form
<table>
  <tr>
    <td width="50%">
      Main Settings Toggle
    </td>
    <td width="50%">
      Lyric Settings Form Toggle
    </td>
  </tr>
</table>

PR #2175, opened 2026-06-22
fix: lyrics export issue when furigana is enabled

york9675:
## Problem
The Japanese Furigana support introduced in #2161 broke the in-app lyrics export feature: when Furigana is enabled, the raw HTML tags and readings are accidentally included in the exported text.

### Example of the issue in lyrics export
```
[00:01.180]<ruby>考<rp>(</rp><rt>かんが</rt><rp>)</rp></ruby>えたってわからないし
[00:02.890]<ruby>青空<rp>(</rp><rt>あおぞら</rt><rp>)</rp></ruby>の<ruby>下<rp>(</rp><rt>した</rt><rp>)</rp></ruby>、<ruby>君<rp>(</rp><rt>きみ</rt><rp>)</rp></ruby>を<ruby>待<rp>(</rp><rt>ま</rt><rp>)</rp></ruby>った
``` 

## Changes
The formatter now removes `rt` and `rp` nodes before extracting text, so `<ruby>考<rt>かんが</rt></ruby>え` becomes `考え` instead of carrying readings into the file.

### DIFF
--- a/src/renderer/features/lyrics/components/lyrics-export-form.tsx
+++ b/src/renderer/features/lyrics/components/lyrics-export-form.tsx
@@ -20,16 +20,6 @@ interface LyricsExportFormProps {
     synced: boolean;
 }
 
-const stripLyricsMarkup = (text: string) => {
-    if (!text.includes('<')) return text;
-
-    const template = document.createElement('template');
-    template.innerHTML = text;
-    template.content.querySelectorAll('rt, rp').forEach((node) => node.remove());
-
-    return template.content.textContent ?? text;
-};
-
 export const LyricsExportForm = ({ lyrics, offsetMs, synced }: LyricsExportFormProps) => {
     const { t } = useTranslation();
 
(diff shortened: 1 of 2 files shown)


src/renderer/features/lyrics/components/lyrics-export-form.tsx
```diff
@@ -20,6 +20,16 @@ interface LyricsExportFormProps {
...
+    if (!text.includes('<')) return text;
+
+    const template = document.createElement('template');
+    template.innerHTML = text;
+    template.content.querySelectorAll('rt, rp').forEach((node) => node.remove());
+
+    return template.content.textContent ?? text;
+};
+
 export const LyricsExportForm = ({ lyrics, offsetMs, synced }: LyricsExportFormProps) => {
```
jeffvli: It's a bit counterproductive to manually strip the furigana when you can just pass the raw lyrics to the `openLyricsExportModal` and `LyricsExportForm` instead of the `displayLyrics`.

jeffvli (changes requested):

york9675:
Updated, now it will pass the raw lyrics into the export modal instead of the furigana-rendered display lyrics.

PR #2181, opened 2026-06-28
perf: use crypto.getRandomValues() for higher-quality shuffle in music player

alabsi91:
This update improves shuffle randomness for the music player by using crypto-based randomness instead of Math.random.

### DIFF
--- a/src/renderer/utils/shuffle.ts
+++ b/src/renderer/utils/shuffle.ts
@@ -1,22 +1,22 @@
 export function shuffle<T>(array: T[]): T[] {
-    // Create a copy of the array to avoid mutating the original
     const shuffled = [...array];
-
-    // Loop through the array from the last element to the first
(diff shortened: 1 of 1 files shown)


src/renderer/utils/shuffle.ts
```diff
@@ -1,22 +1,22 @@
...
     for (let i = array.length - 1; i > 0; i--) {
-        const j = Math.floor(Math.random() * (i + 1));
+        const j = Math.floor(cryptoRandom() * (i + 1));
         [array[i], array[j]] = [array[j], array[i]];
     }
     return array;
 }
+
+function cryptoRandom(): number {
+    const buf = new Uint32Array(1);
```
jeffvli: The buf can be initialized outside of cryptoRandom so that you're not creating a new instance on every loop of the shuffle function.

PR #2182, opened 2026-06-28
feat: add setting for static window title

mihawk90:
### Description
Adds a setting whether or not to show the track info in the Window title. Defaults to true (i.e. current state).

### Motivation and Context
For one, I don't really look at the Window title anyway, so the track information conveyed there is of no use.

However, the motivation for this change is that I'm occasionally window capturing Feishin in OBS Studio (but the same applies to other software). Window capture is matched by the Window title, so having it change all the time means the capture is also lost after an application restart, which means having to set it up again.

### How Has This Been Tested?
This PR for artifacts.

- checked the window title stays static
- checked that Private Mode is still appended to window title, which (intentionally) still breaks capture.

Only tested on Linux. Screen capture on Plasma via xdg-desktop-portal. Have no system to test on Windows or Mac.

### Types of changes
- New feature (non-breaking change which adds functionality)

### DIFF
--- a/src/i18n/locales/en.json
+++ b/src/i18n/locales/en.json
@@ -1175,6 +1175,8 @@
         "webAudio": "Use web audio",
         "windowBarStyle_description": "Select the style of the window bar",
         "windowBarStyle": "Window bar style",
+        "windowBarTrackinfo": "Track info in Window Title",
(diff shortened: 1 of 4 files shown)


mihawk90:
Ugh, I goofed... this was meant to go to my fork until it's finished... I'll close this for now.

PR #2184, opened 2026-06-28
Improve Jellyfin playlist loading and modification performance times

BlackHoleFox:
I've been moving a bunch of playlist content into Jellyfin from another music service using Feishin this weekend. At some point though it became _unusably slow_ in edit -> view iterations. While looking into this I also tried updating to Jellyfin 12.0RC1 since the headline feature is a large number of performance improvements. But this didn't help much :(

After updating the Jellyfin WebUI _loading_ playlists was significantly faster then Feishin though so I started to look at how Feishin was making the same API requests. The two things:
- All playlist-related fetches were requesting `People` and did not seem to be using the response data.
    - Requesting `People` seems like an ongoing performance issue in Jellyfin (https://github.com/jellyfin/jellyfin/issues/15346) even when supposedly fixed in 12.0 :/
 - When modifying a playlist with duplicate detection enabled, each add fetched the whole playlist's contents and threw away every response property except the ID. Beyond `People` being unused and slow, these responses were just unnecessarily slow and large.

### People is slow

To attempt to reduce the `People` slowness I stripped it out of what `getPlaylistSongList` returns. This didn't appear to cause any issues locally. It seems like showing a song's composer (if you have that column property shown) might need it per the code but showing them doesn't look to be functional on `development` either?

composer_missing_in_playlist

Without this query parameter loading a large playlist in Feishin goes from ~13s to ~3s so it does make quite a difference in day to day use.

Before:
load_playlist_old

After:
load_playlist

### Minimize playlist addition requests

The more obviously harmless improvement is just optimizing the API requests made for duplicate detection. The two handlers for playlist additions only used the IDs and it makes a large difference when needing to add multiple songs back-to-back into a playlist which otherwise can end up getting out of sync with multiple requests in flight at once.

Before:
add_with_duplicate_detection_old

After:
add_with_duplicate_detection

### DIFF
--- a/src/renderer/features/context-menu/actions/add-to-playlist-action.tsx
+++ b/src/renderer/features/context-menu/actions/add-to-playlist-action.tsx
@@ -211,11 +211,11 @@ export const AddToPlaylistAction = ({ items, itemType }: AddToPlaylistActionProp
                 let songsToAdd: string[] = allSongIds;
 
                 if (skipDuplicates) {
-                    const queryKey = queryKeys.playlists.songList(serverId, playlistId);
+                    const queryKey = queryKeys.playlists.songListIds(serverId, playlistId);
 
                     const playlistSongsRes = await queryClient.fetchQuery({
                         queryFn: ({ signal }) => {
-                            return api.controller.getPlaylistSongList({
+                            return api.controller.getPlaylistSongIds({
                                 apiClientProps: {
                                     serverId,
                                     signal,
@@ -228,7 +228,7 @@ export const AddToPlaylistAction = ({ items, itemType }: AddToPlaylistActionProp
                         queryKey,
                     });
 
-                    const playlistSongIds = playlistSongsRes?.items?.map((song) => song.id);
+                    const playlistSongIds = playlistSongsRes?.items;
                     const uniqueSongIds: string[] = [];
 
                     for (const songId of allSongIds) {
--- a/src/shared/api/jellyfin/jellyfin-normalize.ts
+++ b/src/shared/api/jellyfin/jellyfin-normalize.ts
@@ -10,7 +10,6 @@ import {
     LibraryItem,
     MusicFolder,
     Playlist,
-    RelatedArtist,
     Song,
 } from '/@/shared/types/domain-types';
 import { ServerListItem, ServerType } from '/@/shared/types/types';
@@ -18,42 +17,6 @@ import { ServerListItem, ServerType } from '/@/shared/types/types';
 const TICKS_PER_MS = 10000;
 
 type AlbumOrSong = z.infer<typeof jfType._response.album> | z.infer<typeof jfType._response.song>;
-
-const KEYS_TO_OMIT = new Set(['AlbumArtist', 'Artist']);
-
-const getPeople = (item: AlbumOrSong): null | Record<string, RelatedArtist[]> => {
-    if (item.People) {
-        const participants: Record<string, RelatedArtist[]> = {};
-
-        for (const person of item.People) {
-            const key = person.Type || '';
-            if (KEYS_TO_OMIT.has(key)) {
-                continue;
-            }
-
-            const item: RelatedArtist = {
-                // for other roles, we just want to display this and not filter.
-                // filtering (and links) would require a separate field, PersonIds
-                id: '',
-                imageId: null,
-                imageUrl: null,
-                name: person.Name,
-                userFavorite: false,
-                userRating: null,
-            };
-
-            if (key in participants) {
-                participants[key].push(item);
-            } else {
-                participants[key] = [item];
-            }
-        }
-
-        return participants;
-    }
-
-    return null;
-};
 
 const getTags = (item: AlbumOrSong): null | Record<string, string[]> => {
     if (item.Tags) {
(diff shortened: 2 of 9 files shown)


kgarner7:
`People` is actually used, although it isn't as visible in Jellyfin. If you do context menu > get info on a track, you can get a list of people used (for whatever reason, the IDs are different).

BlackHoleFox:
Oh yep, I see what you mean. On my branch the `Composer` field there no longer shows up. I guess I'll remove that commit from this branch then since it removes functionality?

Interestingly loading the song in the Jellyfin web UI directly (`https://example.com/Users/<uid>/Items/e0a8276e9354a01d108d419344654efb` returns the `People` object unconditionally but with no performance impact. I guess because its not a query parameter like the reported issue mentions. Do you think its too complicated to make `Get Info` fetch it directly instead of using the list data?

kgarner7:
I'll be honest, the inclusion of `People` in Jellyfin is mostly a worse version of the `participants` exposed by some OpenSubsonic servers and Navidrome (those can be linked to the actual artist). Those _are_ used otuside of `get info` (in particular, there's the ability to promote a `remixer` to an `artist).

For whatever reason (my understanding of the Jellyfin API is limited), the `People` ID for the same name is different than the artist, so it isn't used for linking in the same way. I only see the `composer`, even though I have other roles in the tags, so I would not be entirely opposed to just removing it.

BlackHoleFox:
I wouldn't be opposed to removing it outright if you're fine with that. It makes Feishin generally more responsive in local testing. Loading the `Tracks` view becomes faster as well (~3s to ~1s), for example. I pushed up a commit with the removal from the Jellyfin controllers if you're interested in that path.

src/shared/api/jellyfin/jellyfin-normalize.ts
```diff
@@ -106,11 +70,8 @@ const getPlaylistImageId = (item: z.infer<typeof jfType._response.playlist>): nu
     return null;
 };
 
-const getArtists = (
-    item: z.infer<typeof jfType._response.song>,
-    participants?: null | Record<string, RelatedArtist[]>,
-): RelatedArtist[] => {
-    if (!item?.ArtistItems?.length && !item.AlbumArtists && !participants) {
+const getArtists = (item: z.infer<typeof jfType._response.song>): RelatedArtist[] => {
+    if (!item?.ArtistItems?.length && !item.AlbumArtists) {
```
kgarner7: At this point, it's simple enough that you could just move it into `normalizeSong`

BlackHoleFox: Ah fair, should be gone now.

src/renderer/features/context-menu/actions/add-to-playlist-action.tsx
```diff
@@ -215,7 +215,7 @@ export const AddToPlaylistAction = ({ items, itemType }: AddToPlaylistActionProp
 
                     const playlistSongsRes = await queryClient.fetchQuery({
                         queryFn: ({ signal }) => {
-                            return api.controller.getPlaylistSongList({
+                            return api.controller.getPlaylistSongIds({
                                 apiClientProps: {
                                     serverId,
                                     signal,
```
kgarner7, line 221: The query key will need to be updated to playlistsongids or similar (otherwise it will fetch the wrong details).

src/shared/api/jellyfin/jellyfin-normalize.ts
```diff
@@ -189,9 +121,16 @@ const normalizeSong = (
...
+    (item?.ArtistItems?.length ? item.ArtistItems : item.AlbumArtists)?.forEach((entry) => {
+        artists.push({
+            id: entry.Id,
+            imageId: null,
+            imageUrl: null,
+            name: entry.Name,
+            userFavorite: false,
+            userRating: null,
+        });
+    });
```
kgarner7: Rather than push, you can just assign this entire value to `artists` via `map`

kgarner7 (changes requested):

src/renderer/features/context-menu/actions/add-to-playlist-action.tsx
BlackHoleFox: Oops, my bad. I didn't realize there was caching happening here. Should be fixed now. It doesn't look like the `id` parameter needs to be optional either unlike with the old key.

kgarner7:
One other small note, but the Jellyfin API itself just doesn't seem to allow duplicates _at all_ in a playlist. I'm not really sure the best way to handle that in the interface, because even if you have skip duplicates _off_, it will still do that server side. That being said, I am fine with the implementation as-is (having Feishin check for duplicates and showing count _can_ have more value than the server silently dropping it).

BlackHoleFox:
> I'm not really sure the best way to handle that in the interface, because even if you have skip duplicates off, it will still do that server side.

It would be nice if the endpoint returned the number of changes actually persisted so Feishin could show that even with local de-duplication instead of [silently logging](https://github.com/jellyfin/jellyfin/blob/feef2403c49003761010656cfadaba55a278acd7/Emby.Server.Implementations/Playlists/PlaylistManager.cs#L241-L248).

> showing count can have more value than the server silently dropping it

Would agree with this. Moving over my playlists I've found several "oh whoops I added that track twice in my old platform" which is nice to know.

PR #2189, opened 2026-07-01
feat(player): right-click volume icon to switch audio output device

LoganRupe:
## What

Adds a right-click context menu to the volume/speaker icon in the player bar for quickly switching the audio output device.

- Right-clicking the speaker icon opens a menu of available output devices.
- The currently active device is checkmarked; a **System default** entry resets the selection.
- Works for both the **Web** player (`audioDeviceId`) and the **MPV** player (`mpvAudioDeviceId`).

## Why

Changing the audio output previously required opening Settings → Playback. This puts it one right-click away from where you're already controlling volume. The selection is applied live and **persists across restarts** via the existing settings store, and — unlike the settings-page selector — the menu is **not disabled during playback**.

## Implementation notes

- Reuses the existing `useAudioDevices` hook and the `audioDeviceId` / `mpvAudioDeviceId` playback settings, so device application (web `setSinkId` effect, mpv `--audio-device` param) and persistence are already wired — no new plumbing.
- The icon is wrapped in a plain `<div>` trigger inside `ContextMenu.Target` so the `onContextMenu` handler reliably reaches a DOM element (the tooltip-wrapped `ActionIcon` is not itself a suitable `asChild` target). Left-click (mute) and scroll-to-adjust-volume are unchanged.
- Adds one new i18n string, `setting.audioDeviceDefault` ("System default"), to `en.json`.

## Known limitation

On the **Web** player type with the Web Audio pipeline turned *off*, output switching is currently only wired to the `AudioContext` (not the raw `<audio>` element) — so the choice persists but won't take audible effect until Web Audio is enabled or the MPV player is used. This is a pre-existing limitation of the current codebase, not introduced by this change.

## Testing

Manual: right-click the speaker icon, select a different output device, confirm audio routes to it and that the selection survives an app restart.

> Note: I was unable to run `typecheck`/`lint` in my environment (dependencies weren't installed locally). The change was written against the existing hook/component APIs; a maintainer CI run should confirm.

### DIFF
--- a/src/renderer/features/player/components/right-controls.tsx
+++ b/src/renderer/features/player/components/right-controls.tsx
@@ -590,6 +590,12 @@ const VolumeButton = () => {
         <>
             <ContextMenu>
                 <ContextMenu.Target>
+                    {/*
+                     * ActionIcon renders a Mantine Tooltip wrapper, which does not
+                     * forward the onContextMenu/ref that Radix injects via asChild to
+                     * the underlying button. Wrap in a real DOM node so right-click
+                     * reliably opens the menu.
+                     */}
                     <div style={{ alignItems: 'center', display: 'flex' }}>
                         <ActionIcon
                             icon={muted ? 'volumeMute' : volume > 50 ? 'volumeMax' : 'volumeNormal'}
@@ -612,7 +618,6 @@ const VolumeButton = () => {
                     </div>
                 </ContextMenu.Target>
                 <ContextMenu.Content>
-                    <ContextMenu.Label>{t('setting.audioDevice')}</ContextMenu.Label>
                     <ContextMenu.Item
                         isSelected={!currentAudioDeviceId}
                         onSelect={() => handleSelectAudioDevice(null)}
--- a/src/renderer/features/player/components/right-controls.tsx
+++ b/src/renderer/features/player/components/right-controls.tsx
@@ -7,6 +7,7 @@ import { PlayerConfig } from '/@/renderer/features/player/components/player-conf
 import { CustomPlayerbarSlider } from '/@/renderer/features/player/components/playerbar-slider';
 import { SleepTimerButton } from '/@/renderer/features/player/components/sleep-timer-button';
 import { usePlayer } from '/@/renderer/features/player/context/player-context';
+import { useAudioDevices } from '/@/renderer/features/settings/components/playback/audio-settings';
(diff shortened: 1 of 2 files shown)


src/renderer/features/player/components/right-controls.tsx
```diff
@@ -561,24 +588,49 @@ const VolumeButton = () => {
...
+                            tooltip={{
+                                label: muted ? t('player.muted') : volume,
+                                openDelay: 0,
+                            }}
+                            variant="subtle"
+                        />
+                    </div>
+                </ContextMenu.Target>
+                <ContextMenu.Content>
+                    <ContextMenu.Label>{t('setting.audioDevice')}</ContextMenu.Label>
```
jeffvli: You can remove the Label as it's self explanatory what this menu is doing.

Image

src/renderer/features/player/components/right-controls.tsx
```diff
@@ -561,24 +588,49 @@ const VolumeButton = () => {
...
-                size="sm"
-                tooltip={{
-                    label: muted ? t('player.muted') : volume,
-                    openDelay: 0,
-                }}
-                variant="subtle"
-            />
+            <ContextMenu>
+                <ContextMenu.Target>
+                    <div style={{ alignItems: 'center', display: 'flex' }}>
```
jeffvli, line 599: Is there a reason you added a wrapper div for the ActionIcon?

src/renderer/features/player/components/right-controls.tsx
LoganRupe: Good call — removed the label in 86e56334.

src/renderer/features/player/components/right-controls.tsx
LoganRupe: Yes — `ActionIcon` renders a Mantine `Tooltip` wrapper when a `tooltip` prop is set, and Mantine's Tooltip (v9) destructures the `onContextMenu`/`ref` that Radix injects via `asChild` into its `...others` and applies them to the floating label, not the underlying button (see cloneElement in `Tooltip.mjs` — only onClick/onMouseEnter/onMouseLeave/onPointerDown + the child's own props + targetRef reach the child). So without a real DOM node as the trigger, right-click never opens the menu. Wrapping in a div makes the trigger reliable while keeping the tooltip intact. I've added a comment explaining this in 86e56334 — happy to take a different approach if you'd prefer one.

PR #2197, opened 2026-07-03
feat: play/pause icon on the album art toggles when clicked

nova77:
Make the UI play element consistent with the player at the bottom: when the user click on the album art the play icon toggles between play/pause accordingly.

This works across different views including table, detail, and grid, as well as different sections: main, playlist, tracks, etc.

play_pause_playlist
play_pause_main
play_pause_grid
play_pause_detail


nova77:
Note: There's a new `isActiveItem` that derives whether this card's item is the one currently playing for the new toggle. This is yet-another-thingy doing that (4th! Alongside `useIsCurrentSong`, `useCurrentPlaylistContextId`, and `useRowPlayControl`'s switch) and none of the existing ones covers the grid card's polymorphic item types. It's getting itchy. Ideally we'd consolidate into a shared `useIsActiveLibraryItem(item, itemType)` hook, but that refactor is (IMO) outside the scope of this PR.

kgarner7:
This change doesn't really make sense. The `Play` button on an album will _always_ replace the current play queue, regardless of whether the player is paused or not.

nova77:
> This change doesn't really make sense. The `Play` button on an album will _always_ replace the current play queue, regardless of whether the player is paused or not.

I'm not sure I follow. If I click on the play button inside the album art (see screenshot), it does _not_ change icon. I would expect to change from ▶️ to ⏸️ as it does in the player.

kgarner7:
It's not supposed to. That play icon means "replace the current play queue with this album"

nova77:
Got it, thanks for the clarification. I still find it a bit confusing, but I understand your design choice.

kgarner7:
I could see this as a configurable option (since, this is something that clearly benefits you and possibly others as well), but _probably_ not the default. Thoughts @jeffvli ?

jeffvli:
Unfortunately I don't agree with this change.

The playerbar is already a global UI element with all the associated controls only requiring a single click to access.

Like @kgarner7 mentioned, the card play controls are strictly associated to queue add behavior, not playback.

nova77:
Roger that! Happy to retire it :)

PR #2198, opened 2026-07-03
feat(player): set right ctrl vol max to mpv's vol max if player==local

danglingptr0x0:
I got a lot of music that sounds pretty quiet and I find ReplayGain to not work uniformly well across my library, so I added the option to set the volume control max when the player is set to local via `mpv` extra args. It reads the `mpv` extra args passed to `mpv` via the settings instead of querying `mpv` itself, which is something I considered but decided against.

For example: `--volume-max=200`


jeffvli:
The functionality looks good!

The only change I ask for would be for this to be extended to the rest of the volume handlers in the app.
- mpris volume handler: `src\main\features\linux\mpris.ts`, `src\renderer\features\player\hooks\use-mpris.ts`
- (optionally, as this would require a larger amount of work) remote volume handler: `src\main\features\core\remote\index.ts`, `src\renderer\features\remote\hooks\use-remote.tsx`,

danglingptr0x0:
> The functionality looks good!
> 
> 
> 
> The only change I ask for would be for this to be extended to the rest of the volume handlers in the app.
> 
> - mpris volume handler: `src\main\features\linux\mpris.ts`, `src\renderer\features\player\hooks\use-mpris.ts`
> 
> - (optionally, as this would require a larger amount of work) remote volume handler: `src\main\features\core\remote\index.ts`, `src\renderer\features\remote\hooks\use-remote.tsx`, 

hey! will do. thanks

danglingptr0x0:
> The functionality looks good!
> 
> The only change I ask for would be for this to be extended to the rest of the volume handlers in the app.
> 
> * mpris volume handler: `src\main\features\linux\mpris.ts`, `src\renderer\features\player\hooks\use-mpris.ts`
> * (optionally, as this would require a larger amount of work) remote volume handler: `src\main\features\core\remote\index.ts`, `src\renderer\features\remote\hooks\use-remote.tsx`,

did MPRIS. since the remotes are external, I haven't done those yet

jeffvli:
I'm fine merging without the remote implementation.
Thanks!

kowalcj0:
Hi @danglingptr0x0,
How can I control the value of `MPV_VOLUME_MAX_DEFAULT`? ATM it's hardcoded to `130` and it looks like one can't set it in the `~/config/feishin/config.json` nor control it in playback settings.
I'm controlling the volume on my DAC/Headphone amp. Raising volume in the player above 100% is not desirable as it may introduce clipping and/or distortion. 
Thanks.

ps. kinda related issue https://github.com/mpv-player/mpv/issues/18282

mihawk90:
As the PR description says you can pass `--volume-max` to MPV in the Feishin settings.

That being said, I ran into this too right after the update on Discord.

I don't think it makes sense to inherit the default from MPV in this. IMO Feishin should be setting `--volume-max=100` by default (either transparently in the background or as a default in the config field) to match behaviour between MPV and Web backend. It doesn't make sense that suddenly my volume slider can go to 130 with no explanation. No other audio player I know behaves that way.

On that note:
> **It reads the `mpv` extra args passed to `mpv`** via the settings instead of querying `mpv` itself, which is something I considered but decided against.

If that is true, then how come my slider can go to 130 when no extra params are given?

image

Might be worth moving this to a separate issue though. There's not much point discussing in a merged PR.

PS:

> ps. kinda related issue https://github.com/mpv-player/mpv/issues/18282

This is an unrelated issue. That issue is about raising the volume of the output stream in PipeWire, which is separate from the volume increase in MPV.

danglingptr0x0:
> Hi @danglingptr0x0, How can I control the value of `MPV_VOLUME_MAX_DEFAULT`? ATM it's hardcoded to `130` and it looks like one can't set it in the `~/config/feishin/config.json` nor control it in playback settings. I'm controlling the volume on my DAC/Headphone amp. Raising volume in the player above 100% is not desirable as it may introduce clipping and/or distortion. Thanks.
> 
> ps. kinda related issue [mpv-player/mpv#18282](https://github.com/mpv-player/mpv/issues/18282)

hello. I reverted the change. it didn't behave the way it was supposed to. some of my audio is quiet, but the moment the mpv's volume crosses 100%, it's immediately distorted. I will figure out some other way to amplify without distortion. sorry for the chaos

PR #2210, opened 2026-07-10
feat: added playlist from queue creation

### LINKED ISSUE #2204: [Feature]: Playlist from current queue

- [x] Yes

Not server-specific

I think it would be really cool if there was a button to create a server-side playlist from the current queue. Sometimes you put together a really good queue and want to save it as a playlist for the future.

I don't think that this is server-specific, but I am specifically using feishin with Jellyfin.

Keviddle:
I would like to propose this solution for the request #2204
I think having the option to create a playlist from queue is a good idea. I've needed this functionality more than once as well, so I thought I'd give it a shot at implementing it and speculating that this is a feature you would like to see in the app as well

- added functionality to create a playlist from queue
- prefilling is done as an extra function to be api agnostic since navidrome native api for example does not offer a parameter for filling a playlist with songs on creation


jeffvli:
Thanks!

PR #2218, opened 2026-07-11
feat: add support for loading custom themes from files

make-42:
Adds support for loading custom themes from files, it validates files so that it doesn't crash the whole app if the themes files are invalid...


make-42:
as of now, if the user changes the content of a theme file at runtime, it'll not reload the theme automatically, i could add support for that

make-42:
(i forgot to provide example theme files)
[ontake.css](https://github.com/user-attachments/files/29940075/ontake.css)
[ontake.json](https://github.com/user-attachments/files/29940076/ontake.json)

jeffvli:
Thanks!

make-42:
Thanks to you too for such an amazing piece of software!

PR #2219, opened 2026-07-12
fix furigana and romaji issues

york9675:
## Problem

1. When furigana/romaji generation is enabled, it will not show and error in console like:

```
Error occurred in handler for 'lyric-convert-furigana': TypeError: Cannot read properties of undefined (reading 'length')
    at Object.hasKana (/Users/york/Documents/code/feishin-development/node_modules/.pnpm/kuroshiro@1.2.0/node_modules/kuroshiro/lib/util.js:126:27)
    at convertFurigana (/Users/york/Documents/code/feishin-development/out/main/index.js:3591:28)
    at /Users/york/Documents/code/feishin-development/out/main/index.js:4139:16
    at Session.<anonymous> (node:electron/js2c/browser_init:2:116372)
    at Session.emit (node:events:509:28)
Error occurred in handler for 'lyric-convert-romaji': TypeError: Cannot read properties of undefined (reading 'length')
    at Object.hasKana (/Users/york/Documents/code/feishin-development/node_modules/.pnpm/kuroshiro@1.2.0/node_modules/kuroshiro/lib/util.js:126:27)
    at convertRomaji (/Users/york/Documents/code/feishin-development/out/main/index.js:3614:28)
    at /Users/york/Documents/code/feishin-development/out/main/index.js:4148:16
    at Session.<anonymous> (node:electron/js2c/browser_init:2:116372)
    at Session.emit (node:events:509:28)
```

2. The export lyrics form will crash the app when furigana/romaji generation is enabled.

## Fix

- Normalize synced lyrics before furigana and romaji conversion so generated lyric structures display correctly.
- Normalize synced lyrics before export so generated synced lyrics can be exported without crashing.
- Harden furigana/romaji helpers to safely handle empty or non-string input.

### DIFF
--- a/src/main/features/core/lyrics/furigana.ts
+++ b/src/main/features/core/lyrics/furigana.ts
@@ -35,6 +35,10 @@ const getKuroshiro = async () => {
 };
 
 export const convertFurigana = async (text: string): Promise<string> => {
+    if (typeof text !== 'string' || !text) {
+        return text;
+    }
+
     const KuroshiroClass = (Kuroshiro as any).default || Kuroshiro;
 
     // check if the text contains any Japanese kana (to distinguish Japanese from Chinese text, which shares Kanji)
@@ -51,7 +55,11 @@ export const convertFurigana = async (text: string): Promise<string> => {
 };
 
 export const convertFuriganaFragment = async (text: string): Promise<string> => {
-    if (!text || !hasJapanese(text)) {
+    if (typeof text !== 'string' || !text) {
+        return text;
+    }
+
+    if (!hasJapanese(text)) {
         return text;
     }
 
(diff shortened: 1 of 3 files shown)


nova77:
Will this also fix the issue of not having the option available on non-local lyrics (i.e. handled by navidrome)?

jeffvli:
@nova77 There is no difference in functionality depending on where it is fetched from, since it's using the same component.
Are you actually seeing an issue with it being present on one but not the other depending on the provider?

nova77:
> @nova77 There is no difference in functionality depending on where it is fetched from, since it's using the same component. Are you actually seeing an issue with it being present on one but not the other depending on the provider?

@jeffvli Yeah, that's indeed the case. I'm using the web player (not the standalone, as I am not allowed to at work), and this is the settings page I see.
Screenshot 2026-07-14 at 08 20 53

jeffvli:
Furigana/romaji generation is currently only supported in desktop mode as the generation happens in the desktop process as opposed to the browser.

The [kuroshiro](https://www.npmjs.com/package/kuroshiro) library *does* support browser so it would need to be refactored to work in the browser. @york9675 thoughts?

PR #2220, opened 2026-07-12
Support using alt+click on prev/next to jump to previous/next album in queue

eramdam:
What the title says, demo below.

https://github.com/user-attachments/assets/7f25840a-b8ac-48e0-aebe-9a65a30a01ea

Also improves the performance of `sortSongsByFetcherOrder` by using a `Set()` to store `fetchedIds` and use that instead of looping over the ids every time. A quick benchmark with 2,555 albums and 27,301 songs shows a significant speed-up (LMK if you want me to take this one into its own PR)

```
current x 6.71 ops/sec ±0.95% (21 runs sampled)
new x 88.54 ops/sec ±1.63% (76 runs sampled)
Fastest is new
```


jeffvli:
The performance gains are impressive!

I'm wondering if the alt+click behavior to jump is a pattern you've seen in other music players or is it something that you thought up? My main concern is that if this is not a standard behavior that you would normally see in music players, no one would actually think to try alt+clicking the media player buttons.

One thing that you could add is a second-line tooltip on the Next/Previous buttons in the controls which informs the user of the hotkey, similar to how the `Hold to shuffle` tooltip is displayed on the play buttons.

jeffvli:
It also makes sense to also handle this behavior on all the mediaNext and mediaPrev hotkeys, mpris, etc.

eramdam:
> I'm wondering if the alt+click behavior to jump is a pattern you've seen in other music players or is it something that you thought up? My main concern is that if this is not a standard behavior that you would normally see in music players, no one would actually think to try alt+clicking the media player buttons.
> 

I know I've seen that on Music.app (pre macOS 26) and iTunes when shuffling by albums (which is why I miss it elsewhere haha) but I'll admit I'm not aware of any other players doing it.

> One thing that you could add is a second-line tooltip on the Next/Previous buttons in the controls which informs the user of the hotkey, similar to how the `Hold to shuffle` tooltip is displayed on the play buttons.

Yeah makes sense, I'll dig into that.

eramdam:
> It also makes sense to also handle this behavior on all the mediaNext and mediaPrev hotkeys, mpris, etc.

How do you recommend I go about this? AFAICT, most of the non-button handles for next/previous don't really capture the state of the alt key so I'm not sure if I should look into introducing a new hotkey entirely or if we could, somehow, keep track of the state of the alt key 🤔

eramdam:
> > It also makes sense to also handle this behavior on all the mediaNext and mediaPrev hotkeys, mpris, etc.
> 
> How do you recommend I go about this? AFAICT, most of the non-button handles for next/previous don't really capture the state of the alt key so I'm not sure if I should look into introducing a new hotkey entirely or if we could, somehow, keep track of the state of the alt key 🤔

I ended up implementing this as a new (global allowed) hotkey, LMK what you think

src/renderer/features/player/components/mobile-playerbar.tsx
```diff
@@ -223,7 +223,7 @@ export const MobilePlayerbar = () => {
                     icon={<Icon fill="default" icon="mediaNext" size="md" />}
                     onClick={(e) => {
                         e.stopPropagation();
-                        mediaNext();
+                        mediaNext(false);
```
jeffvli: These use onClick, so you should be able to track altKey here?

eramdam: I could but I figured `e.altKey` wasn't very relevant on a mobile page? (I guess you could use the mobile UI with a keyboard...?)

jeffvli:
> > It also makes sense to also handle this behavior on all the mediaNext and mediaPrev hotkeys, mpris, etc.
> 
> How do you recommend I go about this? AFAICT, most of the non-button handles for next/previous don't really capture the state of the alt key so I'm not sure if I should look into introducing a new hotkey entirely or if we could, somehow, keep track of the state of the alt key 🤔
>
> I ended up implementing this as a new (global allowed) hotkey, LMK what you think

I'm fine with the hotkey approach. What you could have done is add global event listeners which track keydown/keyup events for ALT and pass it down in a global context using a ref. That way even if the event is not a click you could still determine whether or not ALT was pressed.

eramdam:
> I'm fine with the hotkey approach. What you could have done is add global event listeners which track keydown/keyup events for ALT and pass it down in a global context using a ref. That way even if the event is not a click you could still determine whether or not ALT was pressed.

Yeah I figured it'd be something like that, I just couldn't find a pattern in the codebase that did that, but I can dig into it.

src/renderer/features/player/components/mobile-playerbar.tsx
jeffvli: The mobile playerbar can still show up on the desktop app. It's just a responsive page so Mouse+Keyboard still applies.

eramdam: Added in https://github.com/jeffvli/feishin/pull/2220/commits/e58a74f581353046e6b06d24a6c26f2509cf1a8b

PR #2223, opened 2026-07-12
Add basic Sonos support

Hayqe:
I have (with a lot of assistance of DeepSeek..) added basic Sonos support. It supports discovery, connecting and playing music. 

There is an 'interesting' design solution in here: As we are **streaming** directly from Navidrome (or any other subsonic client) and not using the Sonos cloud solution, sonos is not able to understand 'seek', nor does sonos understand the next song

- only play/pause/previous/next is available when steaming to sonos. 
- volume control does not work in app (use the sonos device buttons)
- we use a timer (!) that uses the duration of a  track to 'decide' when a song is completed. At that point, Feishin pushes the next song to Sonos.
- When using pause button on sonos, feishin will register the action and stop this timer. When it registers 'play' again, it will resume from the beginning of the song, restarting the counter.

Yes, this is not the prettiest, but it works and doesn't require a https server + api key + app registration on Sonos platform.


jeffvli:
Have you tried https://github.com/jeffvli/feishin/pull/1887?

Hayqe:
How could I not have found that one... Way more elegant solution. I'm closing this one.

PR #2227, opened 2026-07-13
feat: implemented new fullscreen player design

### LINKED ISSUE #2224: [Feature]: Slightly redesigned fullscreen Now Playing screen

- [x] Yes

Not server-specific

Created a slightly redesigned fullscreen Now Playing screen. Up Next, Related, Lyrics, and Visualizer buttons are moved to the bottom-right as icons instead of text, and the song information below the cover art is removed since it's already displayed in the bottom player bar. Thought I'd share the concept. 

Image

Image

### LINKED ISSUE #2260: [Feature]: Text Wrapping  in the fullscreen player

- [x] Yes

Not server-specific

Very long titles (or short ones, depending on the width of your screen) do not get wrapped in the fullscreen player, it would be nice to have that option as some song titles will get cut and can't be read completely in most of the UI.

Image

### LINKED ISSUE #2285: [Feature]: Cover art animation on hover

- [x] Yes

Not server-specific

A subtle hover animation that slightly enlarges the album art in the Now Playing bar, as shown in video. 

https://github.com/user-attachments/assets/a4e9b7ad-8214-457d-83fb-be195441de36

Also, an option to hide expand/collapse button that moves album art into sidebar. It often gets in the way when clicking album art to open Now Playing screen.

### LINKED ISSUE #2286: [Feature]: Bigger cover in fullscreen player

- [x] Yes

Not server-specific

Hi there, it would be nice to add an option to make the cover noticeably bigger, maybe even like in Georgia-ReBORN theme for foobar2000

Currently we have a look like this:
Image

And Georgia-ReBORN looks like this:
Image

I don't mean "copy that", it's just a reference of how good it would look if cover would be bigger. 

Maybe we could hide duplicate information about a currently playing track (we already see it below and in a queue)  and use that space to stretch a cover in there?

Keviddle:
I would like to propose this solution for the request #2224
I like the mockup/concept @jagrit2u proposed, and took the liberty to give it a shot at implementing this design

- Implemented icon tab navigation in bottom right
- Increased cover art
- Removed unneeded song information

closes #2224, closes #2260, closes #2285, likely closes #2286


jagrit2u:
Looks nice!
One small refinement that I didn't include in the issue, opacity slider could control the entire Now Playing area instead of only the section. I've attached screenshots below to show what I mean.

Before:
Screenshot 2026-07-14 at 19 26 25

Without:
Screenshot 2026-07-14 at 19 21 25


With: Improved text readability.
Screenshot 2026-07-14 E

Keviddle:
@jagrit2u good idea! I take it that its just a mockup again right? 
I will have a look at a possible implementation of that later today

york9675:
I wanted to share one concern about removing the song information from the fullscreen player.

The larger title, artist, and album text under the cover art makes it much easier to identify the currently playing song at a glance. The customizable track metadata is also useful for users who want to see details such as the release year, format, or something else they’ve chosen to display.

With both sections removed, the only remaining song information is in the small control bar at the bottom, which is much less visible in fullscreen mode.

Could this be made optional instead? A toggle to show or hide the title/artist/album and track metadata would allow users to choose between a cleaner layout and a more informative one.

jagrit2u:
@Keviddle Yep, just another mockup. 

@york9675 I agree, this could be optional. Track title, artist, album, etc. could be added to the Player Items configuration in Settings (where codec, year, etc. are already configurable, but title, artist, and album aren't since they've always been visible). If no metadata items are selected, the artwork could use full available space. As metadata items are enabled, the artwork could scale down accordingly to make room for them.

jeffvli:
Also agree with adding back configurable metadata items.
You can use the existing drag/drop settings component to add those.

*also misclicked the close pull request button instead of the comment button, oops*

Keviddle:
Soooo... I finally had the time to implement the suggested changes

 - Metadata is now toggleable and uses the old drag/drop settings
 - Metadata design is the same as beforehand but aligned to the left (looks better in my opinion but feel free to argue differently)
 - Drag/Drop settings are now available in the popup settings menu of the fullscreen player
 - Blur affects the whole background now. New default value is 25
 - I aligned the top left buttons to match the style of the bottom right buttons
 
 
image

image

image

jagrit2u:
- Instead of a separate **"Show song info"** toggle, it might be better to add **Title**, **Artists**, and **Album** inside **Player item configuration** alongside other configurable items. This removes an extra toggle and lets user choose exactly what to display (e.g. **Title** only, or **Title + Artists**). Cover art could automatically scale based on how many items are enabled.

- I think metadata looked better when centered.

- Having lyrics end at the line I marked would align them with the bottom of **Queue** and **Related** tabs. Also, **active lyric line** is a bit off-center relative to album art.



q

Keviddle:
I have another round of updates cooked up and ready to test :D
- Removed show info toggle. If nothing is selected inside of player item config, then the reserved space is automatically freed up for the image. The image also scales with the amount of metadata shown, but strictly speaking that was already in place before this update
- Added alignment setting for the metadata. Default value is center
- Fixed visual alignment of "Player Item Config" text in the popup settings
- Extended lyrics display (by decreasing the shadow on the bottom)

jagrit2u:
- With metadata enabled, album art is smaller than before, text wraps sooner, and gap between artwork and metadata got larger than before.



- When aligned left or right, metadata feels slightly misaligned.
Screenshot 2026-07-23 at 19 46 18
Use the available space so album art follows left/center/right metadata alignment, keeping everything visually aligned.


- Synced lyrics fade at both top and bottom, while the bottom edge of unsynced lyrics area is positioned higher and only fades at the top.
Screenshot 2026-07-23 at 20 36 25

jeffvli:
Agreed with the review from @jagrit2u. 

I'm also wondering if it would look better if the tab buttons were centered underneath the right section instead of absolutely positioned at the bottom right.

jeffvli:
Also related: https://github.com/jeffvli/feishin/issues/2260
It might be worth adding `-webkit-line-clamp` to the title to allow up to 3 lines.

Also, play around with the existing `calculateTitleSize` utility function to see if that works better to autosize the title instead of what's being used now.

Keviddle:
Already working on a fix for the open points of @jagrit2u 
I've also noted that none of these customization are currently available in the mobile/compact view. Neither the metadata selection nor the alignment. I will also work on an implementation for that 

To address the point of the tab buttons: I think they look better absolutely positioned because then they follow the same style and pattern of the upper left buttons. But maybe @jagrit2u can judge that and create a mockup to test that option

jagrit2u:
I think the bottom right placement looked better.

jeffvli:
Bottom right is fine then!

jagrit2u:
### More refinements ###

- Once the visualizer is expanded, shrink icon feels more appropriate. Pressing it to return to Now Playing screen instead of closing it feels more intuitive.


- https://github.com/jeffvli/feishin/issues/2285#issue-4981240483

- I think the outlined lyrics icon looked better than the filled one. The rest look good.

Keviddle:
Next round of updates

- Added a config for metadata placement (left, center, right). Center is the default value
- Metadata will stay in the confines of the image itself, when large enough. When space becomes too small, then the metadata uses up the remaining space
- Added option for scrolling or multi-line titles. Multi-line with 3 rows is the default (#2260)
- Fixed button for visualizer. Button also returns to the correct starting place now
- Changed player buttons to outline buttons
- Added option to deactivate the expand button (#2285)
- Added animation for cover art (#2285)

jagrit2u:
- Layout feels visually heavy now, larger cover art and tighter spacing make the Now Playing screen feel more crowded. 
Screenshot 2026-07-28 at 00 17 08

I think something closer to this would have better visual balance.
With metadata:

or slightly larger.
Without:
Screenshot 2026-07-28 at 02 06 45

- Now that album art expands on hover, **Toggle Fullscreen Player** tip can be removed.

- The hover area extends beyond album art. The animation triggers before reaching it, but clicking there doesn't open Now Playing.

- Now that the button returns to starting place, Now Playing screen no longer needs to swipe down and back up during visualizer expand/collapse.

Keviddle:
Gave it another shot at allowing max customization to accomodate to all feature requests.
The current changes are:
- Option to configure size of the art cover. Defaults to medium
- Fixed unsynced lyrics design
- Fixed a bug with the scrolling title
- Fixed hover trigger point for album cover in player bar

jagrit2u:
- I think refining the default would be better than adding another option. Large already feels good, just reduce the height of tab section slightly so there's a bit more space above tab buttons.
If possible, a subtle top and bottom fade on the **Up next**, **Related** would be a nice touch.

- Center the cover art within the left half and tabs within the right half. It's slightly off-center now.

Keviddle:
Next small update
- Centered the image in its container correctly
- Added fade effect to the play next and similar songs table. Only at the bottom while header is shown, bottom and top if header isn't visible
- Tooltip delay for fullscreen player is slightly larger 

I decided to not remove the tooltip for the fullscreen player and instead just increase the show delay slightly because all in all I think the tooltip serves a valid purpose. There may be people who do not know that the cover art has a function/serves as a button. Therefore only increasing the image is not a sufficient hint in my opinion. Especially while keeping in mind that even buttons with rather obvious or globally accepted icons, like the play button, have a tooltip. But I did increase the tooltip show delay to the default amount for other tooltips as well. The show delay now is in line with other buttons used throughout the fullscreen player

For the second point, I agree that the default should try to get as close as possible to be a default that suffices for most people. But I would argue to still include the setting to choose between small, medium and large for the cover art. We are already looking at an application that offers a maximum number of customization options. If we compare the number of options we have to customize how the lyrics looks like or how the play next table is structured, then we still have a lot of leeway on the number of settings that we could provide for the metadata :D And including for example #2286 in this change, then it seems like there is a need for some for an even bigger album art.

But of course there are just my two cents towards that topic :) I am open to other arguments or discussions

jagrit2u:
- Since we're keeping size setting, I think Small is a bit too small. What about:
Small = current Medium
Medium = current Large
and Large = actual Large


- Fade effect on Up Next and Related covers too much of the first and last items.
Screenshot 2026-07-28 at 18 05 12
Tested locally. Changing fade stops from 15%/85% to 5%/95% seemed to fix this issue.

- Since lyrics are no longer inside a card, I don't think the left/right padding slider is needed anymore.

- Instead of Small, Medium, and Large, can we have a slider?

- Large crops the sides of the album art. I edited a mockup with slightly smaller size. It could use more vertical space on wider screens and automatically reduce its height when space is limited.

jagrit2u:
https://github.com/user-attachments/assets/1753c55d-ac1d-42ee-b704-7b070f2bebf8

Only album art view is good to have. Designed this where clicking active tab icon brings album art front and center, with or without metadata as configured.

src/renderer/features/player/components/left-controls.tsx
```diff
@@ -141,6 +141,7 @@ export const LeftControls = () => {
                                 onContextMenu={handleToggleContextMenu}
                                 role="button"
                                 transition={{ duration: 0.2, ease: 'easeIn' }}
```
jagrit2u: Tested this locally, and easeOut felt snappier here instead easeIn animation.

Keviddle:
Okay I am back with another update
 
 Main points are:
 - I added the cover slide over (and therefore also made the left tabs toggleable by selecting the same option again)
 - I changed the cover size for a slider. I honestly don't know how to make it less confusing. The slider ranges from 40% to 100%, since the setting affects how much of the available container space the image will take up. But if you have metadata activated, then it will happen that the image reaches its max height before actually hitting 100% since the metadata also already take up space inside of the container. 
 - I also reduced the blur on the up next and recommended table
 - Added the newest changes from the main branch
 - Changed image animation to easeOut 
 
 I am not sure if I forgot something from the previously open points. So if I missed an open point, please comment it again and I will have another look :)

jeffvli:
I will note that it's a little bit hard to see the first and last song in the queue/related tables due to the mask-image gradient. Maybe adjust that a bit.

Otherwise, changes are looking good at I think we're close to being able to merge.

jagrit2u:
- Increase the minimum slider range to 55% and remove percentage label.
- Album art starts cropping sides near the upper end of slider or when the app window is narrow. Can cover art resize automatically to fit available space?
- Remove **Lyrics button** from bottom **player bar**. Two visible Lyrics buttons are redundant.
- Add a subtle overshoot/bounce when cover art settles into the center, similar to mockup video above.
- Also, consider using **ease-out**/**easeOut** or only **ease** instead of **ease-in-out**/**easeInOut** for other animations as well. I tested it locally, and it felt noticeably smoother and snappier.
- Change bottom fade of queue and related tabs start from 90% to 95% for better symmetry and visibility, since top fade stops at 5%.

Keviddle:
Okay i had a look at that list and here are my updates so far:
- I increased the slider to 55% and removed labeling
- I added the subtle bounce effect to the animation
- I switched remaining new animations to easeOut
- I changed the fade effect to 95%

I did not remove the lyrics button yet, because I would like to raise the concern that removing it can confuse the user. Because we either have to remove it completely, which would kinda remove functionality that the users are probably already used to, or hide it when opening the fullscreen player. But hiding it would lead to a situation where you can open up the fullscreen player by selecting the lyrics button, but then it would disappear which would look kinda weird and could leave to confusion. 

The album art cropping problem also isn't addressed yet since its fairly complicated. The text size depends on the album art size and for reasons beyond a 15 minutes bugfix scope, the text bugs out when limiting the album art beyond the current limits. But I will investigate that one further later on

jagrit2u:
I think it's ready to merge now. What do you think, @Keviddle @jeffvli?
Just a nitpick: I think Cover Art Size slider should be above Player Item Arrangement.

Keviddle:
Moving the slider makes sense in my opinion to have it in an order that corresponds to the visual order. I moved it.
I also fixed some visual bugs when using the glassy design and changed how the lyrics button in the player bar works with all these changes

Otherwise from my perspective we are good to go for a merge as well

jagrit2u:
- I think this is a bug: tab icons don't dim after mouse leaves Now Playing screen in latest artifact.
- Glassy theme has a bit more space below tab icons, which moves everything above it up.
- I think this would be a nice optional feature with a toggle in settings: automatically return to fullscreen Now Playing view after a period of inactivity when music is playing and the app is left open.

Created an issue
- https://github.com/jeffvli/feishin/issues/2323#issue-5081044101

Keviddle:
I'd argue modifying the sidebar for a feature not in connection with the fullscreen player changes is beyond the scope of this PR.
Especially noticing that there are several bugs/confusing features like the button being named "Queue" instead of sidebar, vertical sidebar not displaying all configured features, vertical sidebar not even visible in glassy design, detached sidebar only showing queue, no matter what, and probably more. If you want, create an issue for reworking/fixing the sidebar and I will take a shot at it after this one 

I will tackle the other open points later today

Keviddle:
I can't reproduce the glassy theme misalignment

image

image

Looks the same to me honestly. And also looking through the html nodes, most have the same width and height

Keviddle:
But i fixed the hover bug and I added the timeout function to automatically open the full-screen player with the default value being disabled

jagrit2u:
Screenshot 2026-08-07 at 03 52 54
Screenshot 2026-08-07 at 03 53 24
The gap below tab buttons is large on Glassy dark theme for me.

jagrit2u:
Screenshot 2026-08-07 at 04 25 11
Album art shadow feels a bit narrow. A wider, soft shadow like in the mockup below would help artwork pop out.

jagrit2u:
Since this PR changes design quite a bit, I think defaults for new installs could be updated as well. A simpler default seems like a better fit, and users who prefer more can always enable it in Settings.
Suggested defaults:
- Player view: No music metadata (it's already shown in bottom bar and queue, so showing it three times feels unnecessary. If someone prefers it, it's only a few click away.)
- Lyrics alignment: Left
- Up Next / Related: Header off, row index off
- Fullscreen player first tap: All tabs closed, with album art centered and tab icons ready to be tapped
- Expand album art: Off
- Visualizer → Sensitivity → Maximum Decibels: -15 dB. Visualizer is a bit too tall now with increased tab height, so lowering it keeps it from getting cut off too much.
- Fullscreen player auto-open timeout (minutes): 1

Maybe I missed something, but here are a couple of screenshots showing what I had in mind for defaults:
Fullscreen Player: When first tapped on **Toggle fullscreen player**
- Cover art size: around 75%


- Enable Background Image: Off, which is current default and not everyone will want it enabled, but when enabled for the first time, **Image Blur Size** should default to **6rem** instead of the current **1.5rem**, which gives the result shown above.

Queue/Related:
Screenshot 2026-08-07 at 15 55 44
Lyrics:
Screenshot 2026-08-07 at 15 58 16

jeffvli:
Agreed on updating the defaults.

Regarding the glassy dark theme - you don't need to make any fixes regarding it. The glassy dark applies a custom css stylesheet which may be conflicting which needs to be fixed on the theme side rather than this PR.

Keviddle:
I changed the defaults as described

jagrit2u:
I think this will be my last ask here
## Couple of previous points that may have been missed: ##

- Album art shadow feels a bit narrow. A wider, soft shadow like in the mockup below would help artwork pop out.



- Since lyrics are no longer inside a card, I don't think left/right padding slider is needed anymore. Everything is aligned in line with other tabs, so why have a slider that misaligns it?

## More suggested defaults: ##
- **Show lyric match, show lyric provider**: **off** (Keeping it simple for new installs. If someone wants more, it's only a few click away)
- **Home page configuration**: I’d suggest changing default order to:
Newly added releases
Explore from your library
Recently released
Recently played
Most played
Home page configuration isn't directly related to Now Playing, but since we're updating defaults, I thought it was worth mentioning. **Most played**(currently the default top item in the list) being first can keep same songs there indefinitely, so I think this order would be better, especially since most users won't change default configuration.
- **Artist, Album background image blur size**: **6rem** (just to keep it consistent)

Keviddle:
I've implemented most of the suggested changes
- Softened shadow. I hope its close to your expectation. Its hard to guess from an image. If not, then please provide a value for the filter on that image.
- Changed the defaults to your recommendation
- Removed the padding for lyrics

I did not change the default order for the home view yet, because I would like to raise beforehand that it might make sense to push "explore from your library" to the top then. For example in my case "newly added" also rarely changes because I do not add music that frequently

jagrit2u:
Yeah, that makes sense. Then **Explore from your library** first and **Newly added releases** second.

Keviddle:
Changed default order of home view as well now.
So feel free to have hopefully a final look around before we can finally merge this :D and therefore for it to be hopefully be included in the next release already

src/renderer/features/player/components/full-screen-player-image.module.css
```diff
@@ -3,7 +3,7 @@
     max-width: 100%;
     height: 100%;
     border-radius: 5px;
-    filter: drop-shadow(0 0 5px rgb(0 0 0 / 40%)) drop-shadow(0 0 5px rgb(0 0 0 / 40%));
+    filter: drop-shadow(0 0 10px rgb(0 0 0 / 20%)) drop-shadow(0 0 10px rgb(0 0 0 / 20%));
```
jagrit2u: filter: drop-shadow(rgba(0, 0, 0, 0.50) 0px 5px 15px);

jagrit2u (changes requested):
Tried this locally and it looked slightly better. Could we use this instead?

src/renderer/features/player/components/full-screen-player-image.module.css
```diff
@@ -3,7 +3,7 @@
     max-width: 100%;
     height: 100%;
     border-radius: 5px;
-    filter: drop-shadow(0 0 10px rgb(0 0 0 / 20%)) drop-shadow(0 0 10px rgb(0 0 0 / 20%));
+    filter: drop-shadow(rgb(0 0 0 / 50%) 0 5px 15px);
```
jagrit2u: 50% feels a little too dark around the artwork. Let's make it 35%?

jagrit2u (changes requested):

src/renderer/features/lyrics/components/lyrics-scroll-content.tsx
```diff
@@ -34,16 +30,12 @@ export const LyricsScrollContent = ({
...
-            '--lyric-padding-right': `${paddingRight}%`,
             gap: gap !== undefined ? `${gap}px` : undefined,
             paddingBottom: bottomScrollPadding,
-            paddingLeft: `${paddingLeft}%`,
-            paddingRight: `${paddingRight}%`,
-            paddingTop: '10vh',
+            paddingTop: '2rem',
             ...style,
         } as React.CSSProperties;
-    }, [bottomScrollPadding, gap, paddingLeft, paddingRight, preview, style]);
```
jeffvli, line 46: Why was the padding configuration removed?

src/renderer/features/player/components/full-screen-player-queue.tsx
```diff
@@ -29,105 +30,182 @@ const ButterchurnVisualizer = lazy(() =>
...
-        <div
-            className={clsx(styles.gridContainer, 'full-screen-player-queue-container')}
-            style={
-                {
-                    '--opacity': opacity / 100,
-                } as CSSProperties
-            }
+        <Group
+            className={[styles.controlsContainer, 'full-screen-player-controls-container'].join(
+                ' ',
```
jeffvli: This should still be using clsx

src/renderer/store/settings.store.ts
```diff
@@ -1359,6 +1387,7 @@ const initialState: SettingsState = {
         showVisualizerInSidebar: true,
         sidebarCollapsedNavigation: true,
         sidebarCollapseShared: false,
+        sidebarImageExpand: false,
```
jeffvli: Should be true . This has been a core feature for a long time and should not be forced to be hidden by default.

src/renderer/features/player/components/left-controls.tsx
```diff
@@ -66,6 +68,12 @@ export const LeftControls = () => {
     const title = currentSong?.name;
     const artists = currentSong?.artists;
 
+    useEffect(() => {
+        if (!sidebarImageExpand && image) {
+            setSideBar({ image: false });
+        }
+    }, [sidebarImageExpand, image, setSideBar]);
+
```
jeffvli, line 76: You do not need an effect here. The condition is already implicit based on the setting.

src/renderer/store/settings.store.ts
```diff
@@ -1309,6 +1336,7 @@ const initialState: SettingsState = {
         externalLinks: true,
         followCurrentSong: true,
         followSystemTheme: false,
+        fullscreenAutoOpenTimeout: 1,
```
jeffvli: I would also say leave this empty by default. I was extremely confused about why the fullscreen player was opening by itself before I saw the setting. 

I think in most cases users would prefer the fullscreen toggle to be explicit rather than automatic.

src/renderer/features/lyrics/components/lyrics-scroll-content.tsx
jagrit2u: Lyrics are no longer inside a card, padding just moves the lyrics out of alignment with the other tabs. So, it didn't seem needed

jeffvli: The primary reason that setting was added was for when using Left Aligned lyrics. That way you could left align but still "center" the line by using padding.

jagrit2u: Makes sense. But wouldn't that leave a large gap between album art and lyrics in this new layout, while also misaligning them with other tabs? I just thought why have a setting that makes design less consistent.

also left-aligned lyrics already start around the center of screen in this new Now Playing layout.

src/renderer/features/player/components/left-controls.tsx
Keviddle: I'd argue that this effect is indeed needed. If you remove it, then the expanded image won't collapse when disabling the functionality to expand the image

The sidebar itself renders its image based only on sidebar.image. So without this effect, disabling the preference while the panel is open would leave it visible. It could also reappear after reload because the app-store state is persisted. The effect clears that now-invalid open state when the setting is disabled.

src/renderer/store/settings.store.ts
Keviddle: I changed it back to null, therefore disabled by default

src/renderer/store/settings.store.ts
Keviddle: Changed it back to true

src/renderer/features/player/components/full-screen-player-queue.tsx
Keviddle: Fixed it

src/renderer/features/player/components/left-controls.tsx
jeffvli: You added a new setting which gates the feature. You just need to consume the setting in the Sidebar as well.

```jsx
    const sidebarImageEnabled = useAppStore((state) => state.sidebar.image);
    const sidebarImageExpand = useSidebarImageExpand();
    const showImage = sidebarImageEnabled && sidebarImageExpand;
```

src/renderer/features/lyrics/components/lyrics-scroll-content.tsx
jeffvli: The Lyrics component is also used in different places in the app (the player sidebar and also the lyrics search screen).

jeffvli: Here's how I was using it in the player sidebar:

Left align with padding:
image

Left align without padding:
image

jagrit2u: My concern is that it just breaks alignment, especially since changing padding in one place affects all other places where lyrics are displayed.

**Without padding**: Everything stays perfectly aligned.
5-small


**With padding**: Alignment starts to fall apart.



Since it's an optional feature, bringing it back does no harm. I just thought, why have a setting that makes design less nice.

src/renderer/features/player/components/left-controls.tsx
Keviddle: 🤦🏼 yeah lets not talk about that one :D I don't know why I didn't think about that. I fixed it. Also renamed the variables a bit to improve the clarity of their purpose

src/renderer/features/lyrics/components/lyrics-scroll-content.tsx
Keviddle: I reverted the removal of the padding options

jeffvli: Ok, I'm fine with removing it.

src/i18n/locales/ca.json
```diff
@@ -147,16 +147,12 @@
...
-                "lyricPaddingLeft": "Espai a l'esquerra de la lletra (%)",
-                "lyricPaddingRight": "Espai a la dreta de la lletra (%)",
                 "followCurrentLyric_description": "Desplaça automàticament per mantenir la lletra activa a la vista",
                 "lyricFollowScrollAlignment_description": "Posició vertical de la línia activa mentre se segueix",
                 "lyricAlignment_description": "Alineació horitzontal de la lletra",
                 "lyricGapSynchronized_description": "Espaiat vertical entre les línies de les lletres sincronitzades",
                 "lyricGapUnsynchronized_description": "Espaiat vertical entre les línies de les lletres no sincronitzades",
                 "lyricLineLeadTime_description": "Avança la línia activa aquests tants milisegons abans de la seva marca temporal",
-                "lyricPaddingLeft_description": "Espaiat a l'esquerra com a percentatge de l'àrea de la lletra",
-                "lyricPaddingRight_description": "Espaiat a la dreta com a percentatge de l'àrea de la lletra",
```
jeffvli, line 159: By the way, you can undo the changes which touch any localization file *except* for en.json. Weblate automatically propagates any changes from the source languages (en).

Keviddle: Alright :) good to know in case I need to remove translation strings again
But due to the revert of the lyrics padding removal, I do not have any translation removals anymore. 
But because this topic is closely related: Do you want me to re-remove the lyrics padding then? I initially reverted the change due to your comment that it might make sense to leave it in for other areas of the application

jeffvli: If you already reverted it that's fine.

Keviddle: Perfect! In that case, there should be no open bugs/changes for now

jeffvli:
Thanks for all your work on this!

PR #2229, opened 2026-07-14
fix: unify scrobbling behavior across subsonic servers

Spyboticsguy:
Small change that modifies the subsonic `scrobble` controller method to always limit sending a `scrobble` payload to the subsonic backend to cases where a scrobble is being submitted or a new song is being started.

This is the current behavior for subsonic backends which support the `reportPlayback` backend (e.g., Navidrome). This change only modifies the behavior for other backends, which would receive `scrobble` notifications whenever a song changed for both the stopped song and the newly-started song.

After this change, all subsonic backends now only receive a `scrobble` when a new song is started or when feishin is attempting to submit a scrobble. This prevents issues where some backends would receive a non-submission scrobble for the previous, stopped song after receiving one for the currently playing song, resulting in the server tracking plays (or now playing information) incorrectly.

### DIFF
--- a/src/renderer/api/subsonic/subsonic-controller.ts
+++ b/src/renderer/api/subsonic/subsonic-controller.ts
@@ -2213,24 +2213,24 @@ export const SubsonicController: InternalControllerEndpoint = {
     scrobble: async (args) => {
         const { apiClientProps, query } = args;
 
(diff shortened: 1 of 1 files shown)


jeffvli:
Thanks!

PR #2238, opened 2026-07-15
feat: add Jellyfin Quick Connect authentication support

### LINKED ISSUE #986: [Feature]: Jellyfin SSO/OAuth or Quick Connect support

- [x] Yes

Jellyfin

I have a jellyfin server with the SSO plugin installed for OAuth logins to my server. It would be great if there was a way to support SSO logins or even quick connect logins to the server.

love the app so far - I'm running it in docker and love that I can have a web client that looks this good. Thank you devs!

ssalonen:
Resolves #986

Allows users to log in to Jellyfin servers using Quick Connect (the built-in Jellyfin passwordless auth flow) instead of a username/password. A 6-digit code is shown which the user confirms in their Jellyfin web UI; the client polls every 5 seconds until authorized and then completes the login automatically.

Tested locally and it seemed to work.

I did get one crash but I think it's unrelated and probably something to do with my audio setup on linux (`[408520:0715/214809.384852:ERROR:media/gpu/vaapi/vaapi_wrapper.cc:1640] vaInitialize failed: unknown libva error`)

I file this now to gauge if is interesting to the project, and would be accepted. I do think the UI needs some polishing, it does not feel very logical now (one should skip filling in the username and password, and instead press the button there). I am also not sure if this success/error communication via `toast` is good 

The code has been heavily assisted by Claude AI.

Screenshots from my manual testing against my jellyfin server.

Connection screen, with quick connect code visible
connection screen

Server side view, accepted the code:
server accepting code

Feishin after logging in:
logged in feishin


src/renderer/features/login/routes/login-route.tsx
```diff
@@ -106,6 +106,124 @@ const LoginRoute = () => {
...
+        try {
+            result = await api.controller.quickConnectInitiate(serverUrl);
+        } catch (err: any) {
+            setIsQuickConnectLoading(false);
+            toast.error({
+                message:
+                    err?.message ??
+                    t('error.quickConnectNotActive', {
+                        defaultValue: 'Quick Connect is not active on this server',
+                    }),
```
jeffvli: You'll need to add all of the new translation strings into the en.json.

src/renderer/features/login/routes/login-route.tsx
```diff
@@ -287,6 +405,34 @@ const LoginRoute = () => {
...
+                                    </Button>
+                                    {quickConnectCode && (
+                                        <Text c="dimmed" size="sm" ta="center">
+                                            {t('form.addServer.input', {
+                                                code: quickConnectCode,
+                                                context: 'quickConnectCode',
+                                                defaultValue: `Enter code {{code}} in your Jellyfin web UI to authorize`,
+                                            })}
+                                        </Text>
+                                    )}
```
jeffvli: UX could be improved by using the CopyButton or similar to allow the user to 1-click copy the code.

src/renderer/features/servers/components/add-server-form.tsx
```diff
@@ -336,6 +434,32 @@ export const AddServerForm = ({ onCancel }: AddServerFormProps) => {
...
+                            </Button>
+                            {quickConnectCode && (
+                                <Text c="dimmed" size="sm" ta="center">
+                                    {t('form.addServer.input', {
+                                        code: quickConnectCode,
+                                        context: 'quickConnectCode',
+                                        defaultValue: `Enter code {{code}} in your Jellyfin web UI to authorize`,
+                                    })}
+                                </Text>
+                            )}
```
jeffvli: Same thing here. Maybe extract to a new component?

ssalonen:
Thanks for the review! I had to rebase the PR as it already had conflicts with `development`. Tested happy flow manually, as well as simple error case of wrong host.

In addition to your review comments, I worked on the UX

Initial view:
image

After clicking the "Quick Connect" tab:
image

After initiating Quick Connect
There is link to open Quick Connect from jellyfin web UI directly

image


Toasts are used for user communication (success & error).

Note that I took liberty of adding some "UI documentation" for the Public URL

src/renderer/features/servers/components/add-server-form.tsx
ssalonen: Components created in follow up PRs

src/renderer/features/login/routes/login-route.tsx
ssalonen: Done

src/renderer/features/login/routes/login-route.tsx
ssalonen: I added all translations now to en.json, and removed the `defaultValue` from code - it was not heavily used in the existing code base.

ssalonen:
OK minor UX polishing still in https://github.com/jeffvli/feishin/pull/2238/commits/c18c127c5bb4245761879aa13fbd8eaf63f05112

- bigger font for the quick connect code
- bright text instead of the gray, pops out more clearly
- visual indicator that "Open Quick connect" opens the server view inside a new window
- animated "progress" indicator before "waiting for approval on your Jellyfin server..."
- 
image

ssalonen:
Forgot re-request a review @jeffvli. Now requested it via UI

src/shared/types/domain-types.ts
```diff
@@ -1766,10 +1771,13 @@ export type InternalControllerEndpoint = {
     getTopSongs: (args: ReplaceApiClientProps<TopSongListArgs>) => Promise<TopSongListResponse>;
     getUserInfo: (args: ReplaceApiClientProps<UserInfoArgs>) => Promise<UserInfoResponse>;
     getUserList?: (args: ReplaceApiClientProps<UserListArgs>) => Promise<UserListResponse>;
+    isQuickConnectEnabled?: (url: string) => Promise<boolean>;
     jukeboxControl?: (
         args: ReplaceApiClientProps<JukeboxControlArgs>,
     ) => Promise<JukeboxControlResponse>;
     movePlaylistItem?: (args: ReplaceApiClientProps<MoveItemArgs>) => Promise<void>;
+    quickConnectInitiate?: (url: string) => Promise<{ code: string; secret: string }>;
+    quickConnectState?: (url: string, secret: string) => Promise<boolean>;
```
jeffvli: After further scrutiny, it doesn't make sense to add this large number of additional controller endpoints. These are Jellyfin-specific, and adds confusion if there were to ever be additional server support added.

Instead, this should all be handled within the authenticate endpoint. You can refactor the `body` property on the authenticate endpoint to allow any number of properties to be passed to it and then conditionally handle everything within authenticate.

For example:
```typescript
authenticate(url: string, body: { action: 'isQuickConnectEnabled' })
authenticate(url: string, body: { action: 'getQuickConnectState', secret: 'zzz' })
authenticate(url: string, body: { action: 'quickConnectInitiate', code: '12345', secret: 'zzz' })
```

ssalonen: https://github.com/jeffvli/feishin/pull/2238/commits/c4c360557d24515caf452dc38c4cdf36bf0697ba

src/shared/types/domain-types.ts
```diff
@@ -240,6 +240,22 @@ export type Artist = Omit<AlbumArtist, '_itemType'> & {
...
+    | { action: 'isQuickConnectEnabled' }
+    | { action: 'quickConnectAuthenticate'; secret: string }
+    | { action: 'quickConnectInitiate' }
+    | { action: 'quickConnectState'; secret: string }
+    | { action?: 'password'; legacy?: boolean; password: string; username: string };
+
+export type AuthenticateResult =
+    | AuthenticationResponse
+    | boolean
+    | { code: string; secret: string };
```
jeffvli, line 257: The goal was moreso just a Record<string, any> so that we don't have to add types for any of the server authenticate functions. Instead, you would add type guards / assertions directly in the authenticate controller for the specific server. That way we can make it completely generic without muddying the domain types.

jeffvli (changes requested):

ssalonen:
@jeffvli can you please check now?

jeffvli:
Looks good, thanks!

PR #2246, opened 2026-07-16
feat(player): add autoplay on open

### LINKED ISSUE #2241: [Feature]: Autoplay on open

- [x] Yes

Not server-specific

This would continue playing the music when opened; if no music was playing, it won't do anything.

Pegoku:
## Summary

- add an opt-in **Autoplay on open** setting under General → Application
- persist the playing state only when autoplay on open is enabled
- automatically enable and lock **Save play queue** while autoplay on open is enabled
- keep paused and stopped sessions idle after reopening

Closes #2241

### DIFF
--- a/src/renderer/store/settings.store.ts
+++ b/src/renderer/store/settings.store.ts
@@ -497,6 +497,7 @@ export const GeneralSettingsSchema = z.object({
     artistItems: z.array(SortableItemSchema(ArtistItemSchema)),
     artistRadioCount: z.number(),
     artistReleaseTypeItems: z.array(SortableItemSchema(ArtistReleaseTypeItemSchema)),
+    autoPlayOnOpen: z.boolean(),
     autoSave: AutoSaveSchema,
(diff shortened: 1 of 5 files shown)


src/renderer/store/settings.store.ts
```diff
@@ -2620,10 +2622,14 @@ export const useSettingsStore = createWithEqualityFn<SettingsSlice>()(
...
 
+                if (version < 32) {
+                    state.general.autoPlayOnOpen = false;
+                }
+
                 return persistedState;
             },
             name: 'store_settings',
-            version: 31,
+            version: 32,
```
jeffvli, line 2632: You can remove the migration here. Not needed as the persist middleware already handles merging new keys.

jeffvli:
Unfortunately it doesn't seem to work for mpv, only the web player. Mpv may require an additional listener to trigger playback start instead of just the player state.

PR #2252, opened 2026-07-17
feat(tag-editor): add batch metadata editor with artwork support (Duplicate)

ChrisScott9456:
This is a duplicate PR for the accidentally closed https://github.com/jeffvli/feishin/pull/2159


ChrisScott9456:
Hm...I think there was some kind of hiccup or something in the build. 

@jeffvli Can you take a look?

jeffvli:
Added my commits to your branch. Let me know if you have any thoughts on the changes, otherwise I think we're pretty much good to merge.

Highlights:
- Embedded lyrics are now parsed 
- Multi-valued tags are no longer flattened to only show the first value. They are now parsed into a TagsInput component where you can add/remove individual items
- Added a settings tab in the song editor modal which allows you to set which fields are multi-value and also set "favorite" values which appear in the TagsInput dropdown
- Reworked the field removal button to be a 3-way toggle between remove, undo, and re-add
- The "Trigger library rescan" checkbox is now persisted to settings
- A few minor styling changes to make it consistent with other parts of the app

ChrisScott9456:
> Added my commits to your branch. Let me know if you have any thoughts on the changes, otherwise I think we're pretty much good to merge.
> 
> Highlights:
> 
> * Embedded lyrics are now parsed
> * Multi-valued tags are no longer flattened to only show the first value. They are now parsed into a TagsInput component where you can add/remove individual items
> * Added a settings tab in the song editor modal which allows you to set which fields are multi-value and also set "favorite" values which appear in the TagsInput dropdown
> * Reworked the field removal button to be a 3-way toggle between remove, undo, and re-add
> * The "Trigger library rescan" checkbox is now persisted to settings
> * A few minor styling changes to make it consistent with other parts of the app

Okay I took a look. I have a some thoughts. 
I tried to look at it as a new user trying to understand how the whole feature works.

---

I think it makes sense to have the Settings tab in the metadata modal so you can actively see the effects of the setting changes, but it wasn't clear to me that the settings were actually permanent unless I clicked "Save" - which made me think "I have to save metadata changes to save settings"

(I realized with testing that they seem to save when edited, but that isn't clear to the user without playing around and figuring it out)

Maybe some sort of visual indicator that settings were saved when editing them near the top of the modal, or just replacing the "Save" button with "Save Settings" that saves the settings and gives a confirmation message, but doesn't close the modal, allowing the user to swap back to the other tabs and continue.

---

I also was very confused by the text: "Fields in this list use a tag input and are saved as separate metadata values. Parsed multivalue fields use it automatically. Lyrics are always handled separately." 

I wasn't really sure what the Multi-Value Fields settings actually did until I played around with it.

I think something like "Fields listed below can receive multiple values that will be saved to the metadata of the file. Values added to the favorites list will appear when adding tags to that field. Lyrics are handled separately." (or something else - idk)

---

Trying to parse the visual information of the favorites list along with the multi-value list made it more confusing for me. I think if we are going to have favorites, it's also something that doesn't necessarily only apply to multi-value fields either.

So maybe the "Favorites" can either be its own section, or we just add an "X" in the autocomplete dropdown as the way of managing the favorites list. Either way, I think if we're gonna keep it, we should allow it for all fields and separate it visually from the multi-value field list.

---

But on the subject of "favorites" - I think a better option is something that I've seen other media players do, which is use existing metadata to form lists to pull from. 

For example, if I have songs with artists like "Beyonce" in my library, when typing in the artist tag field, "Beyonce" (and any other artists I have in my library) will show in the autocomplete. 

But, I guess that also means storing metadata information. Not sure how you feel about that. It would definitely be a future enhancement level effort separate from this PR in my opinion.

---

Also, I understand the different X buttons we have here, but the UI feels really busy with 3 layers of them lol

I think maybe adding tooltips, or even making the X button on the far right some other symbol than an "X" would help the visual clutter. Maybe even coloring the button red would help break things up visually.

I also notice with longer field values, I'm getting ugly left-right scroll wheels. I think maybe we can just extend the modal width instead here to avoid that.

image

---

Let me know what you think about all this. I know that was a lot, so sorry about that lol

jeffvli:
> Maybe some sort of visual indicator that settings were saved when editing them near the top of the modal, or just replacing the "Save" button with "Save Settings" that saves the settings and gives a confirmation message, but doesn't close the modal, allowing the user to swap back to the other tabs and continue.

I don't think this makes sense, as none of the settings in the app utilize a save button. It's intended that all settings are saved automatically.

> I also was very confused by the text: "Fields in this list use a tag input and are saved as separate metadata values. Parsed multivalue fields use it automatically. Lyrics are always handled separately."

I agree this can be worded better.

> Also, I understand the different X buttons we have here, but the UI feels really busy with 3 layers of them lol
> I think maybe adding tooltips, or even making the X button on the far right some other symbol than an "X" would help the visual clutter. Maybe even coloring the button red would help break things up visually.

The buttons already have tooltips when hovered. I do agree with your point that it feels busy. I think I have a decent solution for this by only showing on row hover when the field is unmodified.

> I also notice with longer field values, I'm getting ugly left-right scroll wheels. I think maybe we can just extend the modal width instead here to avoid that.

I can fix this.

> But on the subject of "favorites" - I think a better option is something that I've seen other media players do, which is use existing metadata to form lists to pull from.
> For example, if I have songs with artists like "Beyonce" in my library, when typing in the artist tag field, "Beyonce" (and any other artists I have in my library) will show in the autocomplete.

This is actually possible, but only for certain metadata. Like you mentioned, artists/album artists would be an easy one. Genres as well since these have dedicated endpoints to fetch. Navidrome supports a host of other tags which can be fetched (for example the tags shown in the album/track filters), but Jellyfin and Subsonic do not have this capability. I can give it an attempt (or you can as well).

ChrisScott9456:
> 
> I don't think this makes sense, as none of the settings in the app utilize a save button. It's intended that all settings are saved automatically.
> 

Gotcha. Sounds fine to me then. 👍 

> 
> I agree this can be worded better.
> 

👍 

> 
> The buttons already have tooltips when hovered. I do agree with your point that it feels busy. I think I have a decent solution for this by only showing on row hover when the field is unmodified.
> 

Ah, okay. Yeah I think that's a good solution to try.

> 
> I can fix this.
> 

👍 

> 
> This is actually possible, but only for certain metadata. Like you mentioned, artists/album artists would be an easy one. Genres as well since these have dedicated endpoints to fetch. Navidrome supports a host of other tags which can be fetched (for example the tags shown in the album/track filters), but Jellyfin and Subsonic do not have this capability. I can give it an attempt (or you can as well).

Oh, I see. I don't wanna dive into that now, but if you want to give it a shot, by all means. If not, then I think it would make a good future enhancement.

jeffvli:
One more thing, I switched the labels over to using raw tag names as I want to avoid having to add localization for these.

ChrisScott9456:
> One more thing, I switched the labels over to using raw tag names as I want to avoid having to add localization for these.

The only other thing I would note is I would really prefer that we didn't limit the artwork to 240x240 in the display. It was filling the modal before and provided a nice view of what you're actually uploading.

Other than that, looks good to me!

PR #2257, opened 2026-07-19
fix: keep a seek control when the waveform fails to load

### LINKED ISSUE #2193: [Bug]: progress/seek bar sometimes completely disappear

- [x] Yes

### App Version

1.13.0

### Music Server and Version

Navidrome 0.62.0 (1b46b977)

Desktop Linux

### What happened?

Sometimes the progress/seek bar at the bottom starts disappearing from UI, this happens shortly after switching currently playing track.

It looks like it's trying to switch to a waveform view, but fails. Once in this state, I haven't found a way to get rid of this bug, the only thing that helps is to fully close and restart Feishin. After that the seek bar starts to switch to the waveform view correctly. But can bug out again soon after.

https://github.com/user-attachments/assets/e942cbd1-044a-4fc6-b3ce-fdfb68945b81

_Since this might be somehow relevant, I always keep Feishin on my 4:3 monitor. Maybe it's breaking the UI in some way, I've seen some UI elements not fitting the available space and going "off screen". Unlikely, but I'm mentioning that just in case._

### Steps to reproduce

No idea

mvanhorn:
## Summary

Fixes #2193: the playerbar seek/progress bar can disappear entirely after switching tracks, with the only recovery being a full app restart. This makes the waveform component always keep a working seek control, even when the waveform itself fails to load.

## Background

`PlayerbarWaveform` renders the standard `PlayerbarSeekSlider` as a fallback *only* while `isLoading` is true, and clears `isLoading` on any wavesurfer `ready` event. It has no failure handling: `wavesurfer.load(streamUrl)` is called with no `.catch()`, and there is no `error` listener. So when a load fails, or when a stale `ready` from a superseded load clears the loading flag for the wrong track, the component is left with an empty waveform canvas and no seek slider. The reporter described it as "trying to switch to a waveform view, but fails," and it was confirmed by a contributor across Linux (X11/Wayland) and Windows 11, which points at a logic gap in the component rather than a platform issue. There is no reliable reproduction, but the symptom maps directly to this unhandled-failure path.

## Fix

- Add an explicit `hasError` state and render the fallback `PlayerbarSeekSlider` whenever the waveform `isLoading || hasError`, hiding the empty canvas in that case. This guarantees a seek control is always present.
- Add a `wavesurfer.on('error', ...)` listener and a `.catch()` on `load()` that ignore `AbortError` (the expected result of a rapid track switch) and surface only genuine failures into `hasError`.
- Add a per-load `loadStarted` guard alongside the existing effect cleanup so `ready`/`error` events from a previous, superseded load are ignored and cannot clear the loading state for the wrong track. `hasError` resets on stream-URL change and on a successful `ready`.

The happy path is unchanged: when a load succeeds and `ready` fires, the waveform shows and the fallback hides exactly as before.

## Testing

`pnpm run typecheck` and `eslint` on the changed file both pass. This repository's CI (`test.yml`) runs only `pnpm run lint` (typecheck + eslint + stylelint) and has no unit-test harness, so verification is lint/typecheck plus manual reasoning about the load/ready/error state transitions; the change is scoped to a single component and preserves existing behavior for successful loads.

Closes #2193

AI was used for assistance.

### DIFF
--- a/src/renderer/features/player/components/playerbar-waveform.tsx
+++ b/src/renderer/features/player/components/playerbar-waveform.tsx
@@ -27,6 +27,7 @@ export const PlayerbarWaveform = () => {
     const audioElementRef = useRef<HTMLAudioElement>(document.createElement('audio'));
     const { mediaSeekToTimestamp } = usePlayer();
     const [isLoading, setIsLoading] = useState(true);
+    const [hasError, setHasError] = useState(false);
     const [isDragging, setIsDragging] = useState(false);
     const [tooltipPosition, setTooltipPosition] = useState<null | { x: number; y: number }>(null);
     const [tooltipValue, setTooltipValue] = useState(0);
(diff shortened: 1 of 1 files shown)


mvanhorn:
Appreciate the merge. Nice that a failed waveform no longer costs you the seek bar entirely, thanks for reviewing it.

PR #2262, opened 2026-07-20
Make ratings/favourites a bit more customizable

### LINKED ISSUE #2205: [Feature]: Optionaly rely on ratings over favourites

- [x] Yes

Not server-specific

I prefer star ratings over favourites. My Navidrome library has thousands of 4 and 5 stars ratings, and 0 favourites. Feishin has an option to surface star ratings, but this feels like it could significantly be improved for people who rely on star ratings _only_, by optionally:
 - Hide favourite buttons - in the now playing bottom bar, in the artist and album pages...
 - Rely on star ratings to determine favourite songs (e.g. on the artist page)

Kernald:
As expressed in #2205 , some users might (I know I do) genuinely not use favourites, but rely only on ratings. This changes the current setting controlling whether star ratings are displayed at all, and replaces it by a combination of:
 - Neither favourite nor star rating
 - Only favourite
 - Only star rating
 - Both

This keeps the existing behaviour by default - people who didn't have the star rating displayed still don't have it, and people who had both favorurite and star rating get both. New users get the previous default as well.


src/renderer/features/settings/components/general/application-settings.tsx
```diff
@@ -564,23 +584,25 @@ export const ApplicationSettings = memo(() => {
...
+                    }
                 />
             ),
-            description: t('setting.showRatings', {
+            description: t('setting.favoriteRatingControls', {
                 context: 'description',
             }),
             isHidden: false,
-            title: t('setting.showRatings'),
+            title: t('setting.favoriteRatingControls'),
```
kgarner7: Rather than being a select, I'd prefer it to be two independent switches (preserving the original one).
This also means that you don't have to bump the storage version (new fields don't require such a change).

kgarner7 (changes requested):
1. Please fix your merge conflict.
2. Please have rating/favorite as independent selects, rather than merging them into one field.

src/renderer/features/settings/components/general/application-settings.tsx
Kernald: Done

Kernald:
Split back to two switches, and merged the latest changes in

src/renderer/features/albums/components/album-detail-header.tsx
```diff
@@ -293,7 +294,7 @@ export const AlbumDetailHeader = forwardRef<HTMLDivElement>((_props, ref) => {
                     <LibraryHeaderMenu
                         favorite={detailQuery?.data?.userFavorite}
                         onAlbumRadio={handleAlbumRadio}
-                        onFavorite={handleFavorite}
+                        onFavorite={showFavorites ? handleFavorite : undefined}
```
kgarner7: Rather than do this here, please use the similar scheme that `handleUpdateRating` does: `showFavorites ? () : undefined` for the function definition.

kgarner7 (changes requested):
Small nit, and then looks good!

src/renderer/features/albums/components/album-detail-header.tsx
Kernald: Done!

kgarner7:
Thanks!

PR #2269, opened 2026-07-20
Add a Window menu item that re-opens main window on Mac

eramdam:
image

### DIFF
--- a/src/main/index.ts
+++ b/src/main/index.ts
@@ -708,13 +708,15 @@ async function createWindow(first = true): Promise<void> {
         mainWindow = null;
     });
 
-    mainWindow.on('show', () => {
-        rebuildMainMenu();
-    });
+    if (isMacOS()) {
+        mainWindow.on('show', () => {
+            rebuildMainMenu();
+        });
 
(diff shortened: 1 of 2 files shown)


src/main/index.ts
```diff
@@ -708,6 +708,14 @@ async function createWindow(first = true): Promise<void> {
...
     });
 
+    mainWindow.on('show', () => {
+        rebuildMainMenu();
+    });
+
+    mainWindow.on('hide', () => {
+        rebuildMainMenu();
+    });
+
```
kgarner7, line 720: If the rebuild on show/hide are only necessary fro macOS (which it looks to be the case, as the `subMenuWindow` change is exclusive to `darwin`), it seems like it would make sense to gate this only on macOS.

kgarner7 (changes requested):

src/main/index.ts
eramdam: Done! https://github.com/jeffvli/feishin/pull/2269/commits/bd178893ffffba9c7260cc9d22dae87405588998

kgarner7:
Thanks!

PR #2273, opened 2026-07-22
feat: implemented random button in the playlist view

### LINKED ISSUE #2270: [Feature]: Shuffle/Randomize Albums in Playlist

- [x] Yes

Not server-specific

Feature: When viewing a Playlist, when grouped by Album, when viewing the sorting options, "Random" should be in the list (and implicitly, when I hit play, the tracks should be added to the queue in random album order, but preserve track order within each album).

This is already a sorting option when viewing **all** Albums in my Library, but I would like to be able to create a playlist containing a subset of albums from my library, and randomize the album order only within that playlist.

Keviddle:
This is an implementation that fixes #2270.
Having a random function in many places in the app already, only lead me to the conclusion that it makes sense to implement it in the playlist view as well.
Therefore my changes include
- Random sort order in playlist view for tracks and albums
- Addition of some logic to allow refresh to re-randomize the list

Closes #2270

### DIFF
--- a/src/renderer/features/playlists/components/playlist-detail-album-view.tsx
+++ b/src/renderer/features/playlists/components/playlist-detail-album-view.tsx
@@ -57,7 +57,7 @@ export const PlaylistDetailAlbumView = ({
     const sortBy = (query.sortBy as SongListSort) ?? SongListSort.ID;
     // only re-randomize when the sort is actually random
     const randomRefreshRevision = sortBy === SongListSort.RANDOM ? refreshRevision : null;
-    const randomCacheRef = useRef<{ revision: boolean | null; songs: Song[] }>();
+    const randomCacheRef = useRef<{ revision: boolean | null; songs: Song[] }>(undefined);
 
     const filteredAndSortedSongs = useMemo(() => {
         const raw = data?.items ?? [];
--- a/src/renderer/features/playlists/components/playlist-detail-album-view.tsx
+++ b/src/renderer/features/playlists/components/playlist-detail-album-view.tsx
@@ -1,3 +1,4 @@
+import shuffle from 'lodash/shuffle';
 import { useEffect, useMemo } from 'react';
 
 import { useGridRows } from '/@/renderer/components/item-list/helpers/use-grid-rows';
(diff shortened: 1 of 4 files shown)


stevenre3d:
thanks @Keviddle :)

src/renderer/features/playlists/hooks/use-playlist-track-list.ts
```diff
@@ -103,7 +106,8 @@ export function usePlaylistTrackList(data: PlaylistSongListResponse | undefined)
         const sortBy = (query.sortBy as SongListSort) ?? SongListSort.ID;
         const sortOrder = (query.sortOrder as SortOrder) ?? SortOrder.ASC;
         return sortSongList(filtered, sortBy, sortOrder);
-    }, [data?.items, query, searchTerm]);
+        // eslint-disable-next-line react-hooks/exhaustive-deps -- refreshRevision intentionally triggers a new random order
+    }, [data?.items, query, searchTerm, refreshRevision]);
```
kgarner7: You should only refresh the filter _if_ the sort by is random **and** the sortBy is random.

src/renderer/features/playlists/components/playlist-detail-album-view.tsx
```diff
@@ -59,12 +66,29 @@ export const PlaylistDetailAlbumView = ({ data }: { data: PlaylistSongListRespon
...
+                }
+            });
+
+            return shuffle(Array.from(songsByAlbum.values())).flatMap((albumSongs) =>
+                sortSongList(albumSongs, SongListSort.ALBUM, SortOrder.ASC),
+            );
+        }
+
+        return sortSongList(searched, sortBy, sortOrder);
+        // eslint-disable-next-line react-hooks/exhaustive-deps -- refreshRevision intentionally triggers a new random order
```
kgarner7: Likewise here, this only needs to be refreshed when refresh version changes and the sort order is random.

kgarner7 (changes requested):

src/renderer/features/playlists/hooks/use-playlist-track-list.ts
Keviddle: I updated the deps and only refresh it if the sort type is actually random

src/renderer/features/playlists/components/playlist-detail-album-view.tsx
Keviddle: Same here, I updated the deps and added only refresh it if the sort type is actually random

jeffvli:
I added my own implementation https://github.com/jeffvli/feishin/commit/849356762c40ce17a97ccdd3e57e69469e534189

PR #2276, opened 2026-07-23
Expanded date fields

ermmmwafle:
This PR tries to separate for the DATE and RELEASEDATE fields similar to navidrome which uses ~3 fields for dates and years in addition to some other minor changes.
- DATE/YEAR: The recording date of the track
- RELEASEDATE: The release date of the album
- ORIGINALDATE/ORIGINALYEAR: The original release date of the album (nothing done in this PR)

All instances of year/date in feishin originally referred to the releaseyear/releasedate fields on the servers, so all the labels for these have been changed to 'Release Year' and 'Release Date' for consistency.

This is a list of things I added (or tried to add) to help with reviewing (Note that most of these changes only really apply to navidrome since to my knowledge jellyfin and subsonic don't have an equivalent to the navidrome track date.)
- new 'date' and 'year' fields for Songs. Uses 'releaseyear' as a final fallback if theres no track date/year.
- added date/year fields for fullscreen player, columns on the table, grid, and detail displays for songs, and as client side song filters.
- Added 'year' sort option for songs.
- Song info now shows date and release date with their respective year fields as backups instead of just showing the release year.
image

- date/release date column views now show the full date in english format (e.g. "Oct 12, 2023") instead of just showing the year.
- added the track year range for albums with tracks that span over multiple years from the navidrome ui.
image

- Tried to add the settings version migration to properly add the fields when upgrading, but the version # might need to be changed.

Please let me know about any changes/fixes I should make!


src/shared/api/navidrome/navidrome-normalize.ts
```diff
@@ -93,6 +93,44 @@ const normalizeNavidromeOriginalDate = (item: {
...
+        return null;
+    }
+
+    if (minYear === maxYear) {
+        return null;
+    }
+
+    return { max: maxYear, min: minYear };
+};
+
```
jeffvli: Is there a specific reason you're using a separate function for this? Navidrome's album domain type already returns `minYear` and `maxYear` so this probably isn't needed.

ermmmwafle: Nope. I saw those but completely forgot to check them before the PR! I'll push a commit to fix that in a minute

PR #2280, opened 2026-07-24
fix: f11 fullscreen was completely broken on linux/windows; feat: butterchurn preset expansion + randomized preset arrows

cmyksoda:
i've made these changes:

* **fix**:  f11 previously did nothing on linux or windows, in this fork, f11 properly full-screens the window. additionally, if in the expanded visualizer view (where it is covering the screen except for the now playing section), the visualizer will enter fullscreen mode, covering the entire screen  
***not tested on macOS — that platform keeps its native application menu (unlike linux/windows, where it's disabled entirely), so f11 there is still handled by the existing menu accelerator and won't trigger the new visualizer-specific fullscreen.*** [^1] 

* **feature:** expanded the list of available butterchurn presets to include [extra](https://github.com/jberg/butterchurn-presets/blob/master/extra.js) and [md1](https://github.com/jberg/butterchurn-presets/blob/master/md1.js), adding a total of 243 additional unique presets (30 overlap with the base package already included in the source)

* **feature:** when butterchurn is set to shuffle presets, the left and right arrows on the visualizer now cycle randomly through presets instead of cycling in alphabetical order

i mostly forked this repo to change some things to align with my personal preferences, and understand if one or more of these things aren't to y'alls liking, just let me know if anything here seems like a good change and i'll take out what you don't like and keep the rest locally.

[^1]: first macos ci failure was not due to this, but due to build memory from the imported presets, now fixed by raising the Node heap limit for the renderer build (`--max-old-space-size=4096`)


src/renderer/app.tsx
```diff
@@ -258,6 +259,7 @@ const LanguageEffect = () => {
 
 const NativeMenuSyncEffect = () => {
     useNativeMenuSync();
+    useFullscreenToggle();
```
jeffvli, line 268: This should be moved to its own component instead of being tatcked onto the NativeMenuSyncEffect.

src/renderer/features/visualizer/components/butternchurn/visualizer.tsx
```diff
@@ -40,6 +40,31 @@ export function getButterchurnPresetOptions(presets: Record<string, string>) {
...
+        ...base.default,
+        ...extra.default,
+        ...md1.default,
+    };
+
+    mergedPresetsCache = presets;
+
+    return presets;
+}
+
```
jeffvli, line 67: Good find on the extra butterchurn presets. I didn't even realize these were available.

src/renderer/features/visualizer/components/audiomotionanalyzer/visualizer-settings-form.tsx
```diff
@@ -2113,6 +2116,7 @@ const ButterchurnGeneralSettings = () => {
                     <VisualizerSelect
                         data={presetOptions}
                         label={t('visualizer.selectPreset')}
+                        limit={100}
```
jeffvli: Were you running into performance issues on these Select inputs without limit={100}?
From what I can see it's 349 presets, which should still be fairly reasonable in terms of rendering.

jeffvli:
Thanks!

PR #2282, opened 2026-07-25
feat: add flat layout toggle for song tables

adambenhassen:
One click swaps a song table between its default columns and a flat layout (index, title, artist, album, date added, duration) with no artwork. Reaching that layout previously meant toggling eight columns by hand in the config modal, per list.

Switching back restores the list's factory defaults rather than a hardcoded song order, so album detail keeps its track number column and playlist detail keeps its reorder handle.

Compact row height drops from 40px to 32px; 40px left too much dead space for a single line of text.

Screenshot 2026-07-26 at 00 45 45

This should bring the layout closer to how the classic layout of Spotify looks like right now:

Screenshot 2026-07-26 at 00 49 35

### DIFF
--- a/src/i18n/locales/en.json
+++ b/src/i18n/locales/en.json
@@ -1346,6 +1346,9 @@
                 "trackNumber": "Track Number",
                 "year": "$t(common.year)"
             },
+            "layout": {
+                "toggle": "Switch layout"
+            },
             "view": {
                 "detail": "Detail",
                 "grid": "Grid",
--- a/src/renderer/components/item-list/item-table-list/column-presets.ts
+++ b/src/renderer/components/item-list/item-table-list/column-presets.ts
@@ -1,0 +1,54 @@
+import { ItemTableListColumnConfig } from '/@/renderer/store';
+import { TableColumn } from '/@/shared/types/types';
+
+/**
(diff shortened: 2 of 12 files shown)


jeffvli:
I don't think it makes sense adding a toggle for a predefined column set. Also, 40px was chosen because it is the minimum height at which we can fit the two rows for the `Title (combined)` and `Title (artist)` columns.

I think what you're looking for would actually be user-defined table configuration presets which you can quickly switch between.

daviddavid533:
@jeffvli would you reconsider this one? I've been using the branch for a few weeks and the flat layout is the main reason.

**On the two points raised:**

- **Predefined toggle vs presets**. Fully agree presets are the better end state, but that's a much larger change and nobody has picked it up. This PR is small, self-contained, and covers the most common case (a Spotify-style single-line list). It could land now and be folded into a preset system later as the built-in "Compact" preset, with the toggle button becoming the preset switcher. Nothing here would need to be undone.

- **32px row height**. The 40px minimum only matters when Title (combined) or Title (artist) is in the column set. The flat layout deliberately excludes both, so 32px is safe there. If the concern is the compact setting elsewhere, the height change could be scoped to the flat layout only rather than the global compact value.

Would you take it with the row height change scoped that way?

PR #2283, opened 2026-07-26
feat: add an option to configure default share expiration

yunusey:
First of all, thanks to everyone working on Feishin; I really enjoy using it!

Currently, the default share expiration date is [hardcoded to a year later](https://github.com/jeffvli/feishin/blob/f8a0544cebf375fe5f336817490d79bf91e4a8e5/src/renderer/features/sharing/components/share-item-context-modal.tsx#L31), following [Navidrome's default `DefaultShareExpiration`](https://www.navidrome.org/docs/usage/features/sharing/#default-expiration-for-shares) of `8760h = 1y`.

Navidrome allows this value to be overridden in the configuration, but there is no reliable way to fetch this parameter from the server. (A hack that could solve this is to create a share without passing `expiration` so that it uses the default expiration, find the corresponding expiration date, do the math, and find the default expiration interval. This is quite unorthodox, though.)

Naturally, Feishin defines its own default value to show a preview of what the expiration date of the share will be set to. One potential improvement would be to allow the users to remove the previewed date (by clearing the set expiration date in the share-item-modal) which would then simply omit `expiration` parameter in `shareItem`. In that case, Navidrome would default to the set `DefaultShareExpiration`.

Until I opened up this PR, I wasn't quite familiar with Navidrome's internals much. And I should say that I am still a bit confused as to why this option is even necessary as part of the server configuration. Because, in my opinion, this option is rather a client responsibility (what if the server is shared by multiple people that prefer different default share expiration dates, for instance?). At least, I think it should be stored in a per-user storage, which Navidrome doesn't seem to have.

Right now, for people using Feishin, this flag is practically meaningless as it is never really made use of.

The point is that, I think, the clients should do slightly more than that and allow the users to define their very own expiration intervals. If preferred, they can also make use of Navidrome's flag as a reasonable default.

This PR allows the users to choose a default share expiration interval in their configurations.

_LLM Disclosure: Before doing any reasonable work, I carefully examined the repository to find an LLM policy, but couldn't find one. I then checked the recently merged PRs and encountered some PRs that were LLM-assisted, and I thought this was considered acceptable. I respect the time the maintainers need to spend reviewing these PRs, and I completely understand if you consider this PR low-effort. I used AI tools mainly for reviewing and producing a draft which I then heavily modified and carefully examined each line to the best of my ability. The description of this PR has been written by me without any form of LLM assistance._


src/renderer/store/settings.store.ts
```diff
@@ -2736,10 +2758,20 @@ export const useSettingsStore = createWithEqualityFn<SettingsSlice>()(
...
+                            useServerDefault: false,
+                        };
+                    }
+                }
+
                 return persistedState;
             },
             name: 'store_settings',
-            version: 33,
+            version: 34,
```
jeffvli: Don't need a version migration - this is automatically handled by the persistence middleware as new keys get merged.

src/renderer/features/settings/components/general/sharing-settings.tsx
```diff
@@ -0,0 +1,127 @@
...
+                </Group>
+            ),
+            description: t('setting.shareExpiration', { context: 'description' }),
+            isHidden: shareExpiration.useServerDefault,
+            title: t('setting.shareExpiration'),
+        },
+    ];
+
+    return <SettingsSection options={options} title={t('page.setting.sharing')} />;
+});
```
jeffvli: Instead of putting these in the settings, I think it makes more sense to move these settings directly to the share item modal underneath an accordion pane.

Another nice to have might also add preset buttons underneath the expiration date picker which programatically set the date picker to the selected value (e.g. default, 1 day, 1 week, 1 month, etc).

src/i18n/locales/en.json
```diff
@@ -1148,6 +1151,17 @@
...
         "scrobble_description": "Scrobble plays to your media server",
         "scrobble": "Scrobble",
+        "shareExpiration_description": "How far in the future new share links expire by default. Clear the date in the share dialog to use the server default instead",
+        "shareExpiration_optionSecond": "Seconds",
+        "shareExpiration_optionMinute": "Minutes",
+        "shareExpiration_optionHour": "Hours",
+        "shareExpiration_optionDay": "Days",
+        "shareExpiration_optionWeek": "Weeks",
+        "shareExpiration_optionMonth": "Months",
+        "shareExpiration_optionYear": "Years",
```
jeffvli: These should be moved under the "datetime" i18n key as these are somewhat generic values.

src/renderer/store/settings.store.ts
yunusey: Thanks for the heads up! I have seen similar patterns in `migrate`, and thought we would need to explicitly set the defaults here. Removed in 8f3b323!

src/i18n/locales/en.json
yunusey: I was actually quite unhappy about the new keys I added, and I wanted to move them to somewhere that can later be used should there be a need. So, thanks a lot for pointing to a better place!

Fixed in 1323c7d, but let me know if you have a better naming convention in mind. Currently, under `datetime`, the defined keys are:
```json
{
        "minuteShort": "m",
        "secondShort": "s",
        "hourShort": "h",
        "dayShort": "d"
}
```
So, I simply named the new ones to be the `Long` version of the keys:
```json
{
        "secondLong": "Seconds",
        "minuteLong": "Minutes",
        "hourLong": "Hours",
        "dayLong": "Days",
        "weekLong": "Weeks",
        "monthLong": "Months",
        "yearLong": "Years",
}
```
Of course, maybe it would be unreasonable for this commit, but maybe we could consider adding the `Short` versions for weeks, months, and years in the future?

src/renderer/features/settings/components/general/sharing-settings.tsx
yunusey: I most definitely agree with the second point. I just wasn't sure if that would be too many buttons in the modal. The client I use on mobile, [substreamer](https://github.com/ghenry22/substreamer) has [such preset buttons](https://github.com/ghenry22/substreamer/blob/866fc16a26b06f9d827f447536e196b987786e23/src/components/CreateShareSheet.tsx#L24-L31), by the way. I will write a patch implementing that.

And about the settings in the modal, I have a question: are you suggesting that we move these out of the settings store entirely and just keep the last-used options in local storage (or modal state), or that we still keep them in the store (so they can be exported) but _only_ configure them from the share modal instead of the settings page?

If it is the second one, I am going to need to look into it in detail, probably. I had been treating store settings as something you mainly configure from the settings menu, though I realize `setSettings` can be called from elsewhere too. Though, it still requires that it is accessible through the settings menu?

If it is the first one, I think that works (arguably, it is better than the current approach). I don't really see that much of a use for having the share expiration in the store anyway. I was thinking of it as a good place to configure settings in general. I think I can most definitely do that.

I think a little clarification would be great here; thanks in advance!

jeffvli: >  I had been treating store settings as something you mainly configure from the settings menu, though I realize setSettings can be called from elsewhere too. Though, it still requires that it is accessible through the settings menu?

There's not really much consistency here. The main benefit to sticking everything in the store is that it is part of the settings backup/restore so that users can retain settings.

I don't really have a preference either way. If you think it's a small enough configuration where persistence isn't necessary go ahead and remove it from the store.

yunusey: Upon rereading my comment from yesterday, I see that I did a pretty bad job making my point across. I think I now understand what you were asking me to. I moved the sharing settings into the modal as you asked me to, and removed the configuration options from the settings menu altogether, but we still store the sharing settings in the store.

I really wouldn't want to waste anymore review time, but just in case it helps, here is a bit of an explanation regarding my initial confusion.

When I began working on this feature and added the sharing settings in the store, I was observing a pattern of "any option in the store should have a dedicated menu under the settings menu". So, when you asked me to move the sharing settings from the settings menu, I thought you were asking me to remove the options from the store altogether, and I was thinking that `setSettings` was used _alongside_ the dedicated settings menu.

Fast forward to today, the example I needed was in front of me all along: [`lyricsDisplay`](https://github.com/jeffvli/feishin/blob/a26b8c4839b39bf2ffb126472776fde5cc63dbfa/src/renderer/features/lyrics/components/lyrics-settings-form.tsx#L47-L57), for instance, is configured only through its very own settings form, and is not part of the general settings menu.

I am sorry for wasting your time with my previous question; I could've spent some more time trying to figure out the solution myself.

yunusey: Oh, and the preset buttons are also added (592fd56)!

yunusey:
Quick update on the changes since the last review!

I’ve added 5 preset buttons: `Server Default`, `1 Day`, `1 Week`, `1 Month`, and `1 Year`.

The only missing preset from substreamer is "90 Days". I skipped it for now because supporting it requires handling plural units (like `Days`) under `datetime`. Due to this same plural constraint, the dropdown menu currently relies on singular units as well.

I'd love your input on the i18n approach here: Is it better to add explicit keys for each preset string (e.g., dedicated keys for `1 Day` and `90 Days`, which is [how substreamer handles it](https://github.com/ghenry22/substreamer/blob/866fc16a26b06f9d827f447536e196b987786e23/src/i18n/locales/ms.json#L686-L691)), or should we define singular/plural time units once and build the labels dynamically?

I wanted to check in before making a call on adding new keys or refactoring that logic.

Here's a quick demo of the current state:


https://github.com/user-attachments/assets/84eea15c-d9ab-4027-a72f-14c043427d03



Let me know what you think!

PR #2289, opened 2026-07-26
feat: adds option to view top rated songs in the Favorite Songs section

dsimonds:
Adds an option to view a user's Top Rated songs on the Artist's Detail page

This partially satisfies the spirit #2205 (allowing ratings to be utilized more) by providing the option to view the user's top rated songs of the artist in the "Favorite songs" section, since some users rely mainly on star ratings instead of favorites. Now the user can select the new "Favorite | Ratings" segmented control option to switch between songs that are  favorited, or songs that have a rating of 3 or more stars.

**Favorite:**
ordered by: userFavorite > userRating > playCount > albumId > trackNumber
image

**Rating:**
ordered by: userRating > userFavorite > playCount > albumId > trackNumber
image

Notes:
- Jellyfin only has favorites, so the Jellyfin UI has remained unchanged, and will only show favorites. I just updated code to support the new query methods.
- Added support for 07ef323, so if a user chooses to only show Favorites or Ratings, then the "Favorite songs" section will only show favorited or top rated songs based on their selection.
- I have tested this with all three server types: (Navidrome, OpenSubsonic, and Jellyfin)

### DIFF
--- a/src/renderer/api/navidrome/navidrome-controller.ts
+++ b/src/renderer/api/navidrome/navidrome-controller.ts
@@ -557,6 +557,56 @@ export const NavidromeController: InternalControllerEndpoint = {
         );
     },
     getDownloadUrl: SubsonicController.getDownloadUrl,
+    getFavoriteSongs: async (args) => {
+        const { apiClientProps, query } = args;
+
+        const type = query.type === 'favorite' ? 'favorite' : 'rating';
+
+        if (type === 'rating') {
+            const res = await NavidromeController.getSongList({
+                apiClientProps,
+                query: {
+                    artistIds: [query.artistId],
+                    sortBy: SongListSort.RATING,
+                    sortOrder: SortOrder.DESC,
+                    startIndex: 0,
+                },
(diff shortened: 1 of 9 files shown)


src/renderer/api/navidrome/navidrome-controller.ts
```diff
@@ -557,6 +557,56 @@ export const NavidromeController: InternalControllerEndpoint = {
         );
     },
     getDownloadUrl: SubsonicController.getDownloadUrl,
+    getFavoriteSongs: async (args) => {
+        const { apiClientProps, query } = args;
+
+        const type = query.type === 'favorite' ? 'favorite' : 'rating';
```
jeffvli: Small nitpick, but the `type` const is redundant as you can just perform the comparison with `query.type`.

dsimonds: Thanks! Updated to just use `query.type` here, and in the same spot in subsonic-controller.ts

PR #2296, opened 2026-07-29
fix: table row playback controls glitch

maximelafarie:
Small fix on a glitch on the table row playback controls position.

Glitch on the first rows (interactions on the playback controls are impossible) :

Capture d’écran 2026-07-29 à 10 08 37

When the popover is in `up` position, the bug disappears (interactions with playback controls are possible):

Capture d’écran 2026-07-29 à 10 08 45

The fix aims to render the popover within the Portal, as the default value is `true` on the `MantineHoverCard` but it may be tweaked in any future development, so I thought it was better to keep it exposed.

### DIFF
--- a/src/renderer/components/item-list/item-table-list/columns/row-play-control-cell.tsx
+++ b/src/renderer/components/item-list/item-table-list/columns/row-play-control-cell.tsx
@@ -110,7 +110,7 @@ export const RowPlayControlCell = (
 
     return (
         <TableColumnTextContainer {...props} className={styles.fullSizeContent}>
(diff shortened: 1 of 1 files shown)


jeffvli:
Unfortunately using a portal on that component was causing an issue with React maximum update depth exceeded when scrolling quickly through the table that I was unable to debug.

There will need to be a different solution to this.

maximelafarie:
Hi @jeffvli, that's very clear. I didn't know the reason behind that. I'll look for another fix.

jeffvli:
Fixed in https://github.com/jeffvli/feishin/commit/4318c7af632fb48bbc181b19c82ef4f93e665e03

PR #2297, opened 2026-07-30
fix: preserve pitch toggle now works for MPV backend

sekiryl:
- Toggling off preserve pitch had no effect when MPV backend was being used.
- Preserve pitch toggle now correctly controls MPV's `audio-pitch-correction` property.


kgarner7:
Thanks!

PR #2300, opened 2026-07-30
feat: confirm before discarding the current queue

york9675:
It’s easy to accidentally press Play and immediately replace the current queue, which is sad.

This adds a confirmation before the queue is replaced or cleared.

The confirmation is configurable and enabled by default.

This also replaces the Clear Queue “X” icon with a trash icon, since the old icon could be mistaken for closing the queue view.


src/renderer/features/player/context/player-context.tsx
```diff
@@ -171,6 +171,36 @@ export const PlayerProvider = ({ children }: { children: React.ReactNode }) => {
...
+                        onConfirm={() => {
+                            closeAllModals();
+                            onConfirm();
+                        }}
+                    >
+                        <Text>{t('form.queueChangeConfirmation.description')}</Text>
+                    </ConfirmModal>
+                ),
+                title: t('form.queueChangeConfirmation.title'),
+            });
```
jeffvli, line 217: Can you add a "Do not show again" checkbox here so that users don't have to dig through the settings to disable it?

Just follow the same pattern as used in the `confirmLargeFetch` function in the same file, except you can use the settings state instead of local storage.

PR #2309, opened 2026-07-31
fix: use the ID3 getArtistInfo2 endpoint for artist info

### LINKED ISSUE #2308: [Bug]: Artist biography and similar artists come from an unrelated artist on Subsonic servers

- [x] Yes

### App Version

1.15.1

### Music Server and Version

Airsonic-Advanced (REST API 1.15.0)

Desktop Linux

### What happened?

On the album artist detail page, the "About" section and the "Related Artists" list belong to a completely different artist than the one the page is opened for.

Everything else on the page is correct — albums, top songs, external links. Only the biography and the similar artists are wrong. The mismatch is stable for a given artist page: the same unrelated artist is shown on every reload, and different artist pages get different unrelated artists.

Expected: the biography and similar artists of the artist whose page is open.

## Root cause

Feishin loads album artists through the ID3 endpoints (`getArtists.view` / `getArtist.view`), but loads artist info through the non-ID3 `getArtistInfo.view`. `getArtistInfo2` is not declared anywhere in the codebase.

- `src/renderer/api/subsonic/subsonic-api.ts` — the `getArtistInfo` contract
- `SubsonicController.getAlbumArtistInfo` and `NavidromeController.getAlbumArtistInfo` — both call it with the ID3 artist id

In the Subsonic API these two endpoints resolve the `id` parameter in two different id namespaces: `getArtistInfo` in the folder/media-file browsing namespace, `getArtistInfo2` in the ID3 tag namespace. The namespaces overlap numerically, so the id is valid in both and no error is raised.

Airsonic-Advanced resolves the id against `media_file` in `getArtistInfo`:

https://github.com/airsonic-advanced/airsonic-advanced/blob/68d11bfbfe051b0acaca2770c3c1f47f8d59201c/airsonic-main/src/main/java/org/airsonic/player/controller/SubsonicRESTController.java#L434-L440

and against the `artist` table in `getArtistInfo2`:

https://github.com/airsonic-advanced/airsonic-advanced/blob/68d11bfbfe051b0acaca2770c3c1f47f8d59201c/airsonic-main/src/main/java/org/airsonic/player/controller/SubsonicRESTController.java#L470-L476

Passing an ID3 artist id to `getArtistInfo` therefore makes the server resolve an unrelated row in `media_file` that happens to have the same numeric id, and return that artist's Last.fm data. No error is raised, so the wrong biography is indistinguishable from a correct one.

This affects any server where the two namespaces differ. I only have an Airsonic-Advanced instance, so everything below is read from the sources of the other servers and is not verified against running instances.

On Navidrome the bug should not be observable at all: [`GetArtistInfo`](https://github.com/navidrome/navidrome/blob/d23b68a4385d42b647cb2c349ba5e1ac36fc4c1e/server/subsonic/browsing.go#L307-L308) and [`GetArtistInfo2`](https://github.com/navidrome/navidrome/blob/d23b68a4385d42b647cb2c349ba5e1ac36fc4c1e/server/subsonic/browsing.go#L327-L328) are both backed by the same [`getArtistInfo(r)`](https://github.com/navidrome/navidrome/blob/d23b68a4385d42b647cb2c349ba5e1ac36fc4c1e/server/subsonic/browsing.go#L281-L291) and differ only in the shape of the response — which would explain why this went unnoticed.

On gonic the same cause looks like it produces a different symptom: [`ServeGetArtistInfo`](https://github.com/sentriz/gonic/blob/fc28428204601286e8da626eea6a395ee3a2ccce/server/ctrlsubsonic/handlers_by_folder.go#L303-L305) returns an empty response and only [`ServeGetArtistInfoTwo`](https://github.com/sentriz/gonic/blob/fc28428204601286e8da626eea6a395ee3a2ccce/server/ctrlsubsonic/handlers_by_tags.go#L320) looks up the artist, which would mean the biography is missing entirely there rather than wrong. Confirmation from someone running gonic or another Subsonic server would be welcome.

### Steps to reproduce

1. Connect Feishin to an Airsonic / Airsonic-Advanced server
2. Open an album artist page for an artist that has a Last.fm biography
3. The "About" section shows the biography of a different artist, and "Related Artists" lists that other artist's similar artists

### Relevant log output

```shell
# $ID is an ID3 artist id, taken from getArtists.view — i.e. exactly the id Feishin
# passes to this endpoint. Both calls hit the same Airsonic-Advanced server.

$ curl -s "$SRV/rest/getArtistInfo.view?$AUTH&id=$ID" | jq -r '.["subsonic-response"].artistInfo.biography'
<biography of an unrelated artist, matching what the UI renders>

$ curl -s "$SRV/rest/getArtistInfo2.view?$AUTH&id=$ID" | jq -r '.["subsonic-response"].artistInfo2.biography'
<biography of the requested artist, i.e. the correct one>

# same id, same server, two different artists
```

noctuum:
Closes #2308

## Problem

On the album artist detail page the "About" section and the "Related Artists" list belong to a completely different artist than the one the page is opened for. Albums, top songs and external links on the same page are correct.

## Cause

Album artists are loaded through the ID3 endpoints (`getArtists.view` / `getArtist.view`), but artist info was requested through the non-ID3 `getArtistInfo.view`. The two endpoints resolve the `id` parameter in two different id namespaces, and the namespaces overlap numerically, so the id is valid in both and no error is raised.

Airsonic-Advanced resolves the id against `media_file` in `getArtistInfo`:

https://github.com/airsonic-advanced/airsonic-advanced/blob/68d11bfbfe051b0acaca2770c3c1f47f8d59201c/airsonic-main/src/main/java/org/airsonic/player/controller/SubsonicRESTController.java#L434-L440

and against the `artist` table in `getArtistInfo2`:

https://github.com/airsonic-advanced/airsonic-advanced/blob/68d11bfbfe051b0acaca2770c3c1f47f8d59201c/airsonic-main/src/main/java/org/airsonic/player/controller/SubsonicRESTController.java#L470-L476

## Fix

`getAlbumArtistInfo` now calls `getArtistInfo2.view` in both the Subsonic and the Navidrome controller, and the response is typed by a new `artistInfo2` schema.

| Server | Before | After | Basis |
| --- | --- | --- | --- |
| Airsonic / Airsonic-Advanced | biography and similar artists of an unrelated artist | correct artist | tested |
| gonic | no biography at all — `getArtistInfo` returns an empty response there | correct artist | expected, from reading the sources |
| Navidrome | already correct — both endpoints are backed by the same handler | unchanged | expected, from reading the sources |
| Jellyfin | not affected — separate code path, biography comes from `Overview` | unchanged | code inspection, not exercised |

Only the Airsonic-Advanced row was verified on a running server. The other rows are inferred from the server sources, and I would appreciate a check by someone who runs those servers.

<details>
<summary>Screenshots (Airsonic-Advanced)</summary>

**Before**

before1
before2

**After**

after1
after2

</details>

### Notes for review

- No fallback to `getArtistInfo` is kept on purpose. That endpoint does not fail on an ID3 id, it returns a plausible-looking biography of the wrong artist, so a fallback would silently reintroduce the bug instead of guarding against it.
- `getArtistInfo2` requires Subsonic API 1.11.0. The client already advertises `v=1.13.0`, so no version bump is needed.
- The new schema is written against actual server responses rather than copied from the old one: `albumCount` arrives as a number from Airsonic although the previous schema declared it a string, `coverArt` is missing on some entries, and gonic omits `similarArtist` entirely when the list is empty. These schemas are compile-time types only — the ts-rest client returns the raw body without parsing — so the mismatch surfaced as inaccurate types rather than a runtime error. `String(artist.id)` is needed because the shared `id` schema is `number | string` while `RelatedArtist.id` is `string`.

### Testing

Tested on Airsonic-Advanced (REST API 1.15.0) only — that is the one server I have:

- the same ID3 artist id returns an unrelated artist from `getArtistInfo` and the correct artist from `getArtistInfo2`;
- running the app from source, the biography and the "Related Artists" list now both match the artist on the page;
- `pnpm typecheck` and `eslint src/` are clean.

Not tested on Navidrome, gonic, LMS, Ampache or Jellyfin.

### DIFF
--- a/src/renderer/api/navidrome/navidrome-controller.ts
+++ b/src/renderer/api/navidrome/navidrome-controller.ts
@@ -289,7 +289,7 @@ export const NavidromeController: InternalControllerEndpoint = {
     getAlbumArtistInfo: async (args) => {
         const { apiClientProps, query } = args;
 
-        const artistInfoRes = await ssApiClient(apiClientProps).getArtistInfo({
+        const artistInfoRes = await ssApiClient(apiClientProps).getArtistInfo2({
             query: {
                 id: query.id,
                 ...(query.limit != null && { count: query.limit }),
@@ -300,7 +300,7 @@ export const NavidromeController: InternalControllerEndpoint = {
             return null;
         }
 
-        const artistInfo = artistInfoRes.body.artistInfo;
+        const artistInfo = artistInfoRes.body.artistInfo2;
         const imageUrl =
             artistInfo?.largeImageUrl ||
             artistInfo?.mediumImageUrl ||
@@ -312,8 +312,8 @@ export const NavidromeController: InternalControllerEndpoint = {
             imageUrl,
             similarArtists:
                 artistInfo?.similarArtist?.map((artist) => ({
-                    id: artist.id,
-                    imageId: artist.id,
+                    id: String(artist.id),
+                    imageId: String(artist.id),
                     imageUrl: null,
                     name: artist.name,
                     userFavorite: Boolean(artist.starred) || false,
--- a/src/renderer/api/subsonic/subsonic-api.ts
+++ b/src/renderer/api/subsonic/subsonic-api.ts
@@ -98,12 +98,12 @@ export const contract = c.router({
             200: ssType._response.getArtist,
         },
     },
-    getArtistInfo: {
+    getArtistInfo2: {
         method: 'GET',
-        path: 'getArtistInfo.view',
+        path: 'getArtistInfo2.view',
         query: ssType._parameters.artistInfo,
         responses: {
-            200: ssType._response.artistInfo,
+            200: ssType._response.artistInfo2,
         },
     },
     getArtists: {
--- a/src/renderer/api/subsonic/subsonic-controller.ts
+++ b/src/renderer/api/subsonic/subsonic-controller.ts
@@ -496,7 +496,7 @@ export const SubsonicController: InternalControllerEndpoint = {
     getAlbumArtistInfo: async (args) => {
         const { apiClientProps, query } = args;
 
-        const artistInfoRes = await ssApiClient(apiClientProps).getArtistInfo({
+        const artistInfoRes = await ssApiClient(apiClientProps).getArtistInfo2({
             query: {
                 id: query.id,
                 ...(query.limit != null && { count: query.limit }),
@@ -507,14 +507,14 @@ export const SubsonicController: InternalControllerEndpoint = {
             return null;
         }
 
-        const artistInfo = artistInfoRes.body.artistInfo;
+        const artistInfo = artistInfoRes.body.artistInfo2;
 
         return {
             biography: artistInfo?.biography || null,
             similarArtists:
                 artistInfo?.similarArtist?.map((artist) => ({
-                    id: artist.id,
-                    imageId: artist.coverArt ?? artist.id,
+                    id: String(artist.id),
+                    imageId: artist.coverArt ?? String(artist.id),
                     imageUrl: null,
                     name: artist.name,
                     userFavorite: Boolean(artist.starred) || false,
--- a/src/shared/api/navidrome/navidrome-normalize.ts
+++ b/src/shared/api/navidrome/navidrome-normalize.ts
@@ -392,7 +392,9 @@ const normalizeAlbum = (
 
 const normalizeAlbumArtist = (
     item: z.infer<typeof ndType._response.albumArtist> & {
-        similarArtists?: z.infer<typeof ssType._response.artistInfo>['artistInfo']['similarArtist'];
+        similarArtists?: NonNullable<
+            z.infer<typeof ssType._response.artistInfo2>['artistInfo2']
+        >['similarArtist'];
     },
     server?: null | ServerListItem,
 ): AlbumArtist => {
@@ -448,8 +450,8 @@ const normalizeAlbumArtist = (
         playCount: item.playCount || 0,
         similarArtists:
             item.similarArtists?.map((artist) => ({
-                id: artist.id,
-                imageId: artist.id,
+                id: String(artist.id),
+                imageId: String(artist.id),
                 imageUrl: null,
                 name: artist.name,
                 userFavorite: Boolean(artist.starred) || false,
--- a/src/shared/api/subsonic/subsonic-normalize.ts
+++ b/src/shared/api/subsonic/subsonic-normalize.ts
@@ -245,14 +245,14 @@ const normalizeSong = (
 const normalizeAlbumArtist = (
     item:
         | (z.infer<typeof ssType._response.albumArtist> & {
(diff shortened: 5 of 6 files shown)


kgarner7 (changes requested):
No. This is an incorrect change.
`getArtistInfo` is **specifically** used because the spec notes that it fetches data from last.fm (or other providers). Just replacing this with `getArtistInfo2` is not the right call. You could provide a toggle (although for Navidrome, there's no difference), but this alone is incorrect.

noctuum:
@kgarner7 Could you point me at the part of the spec you have in mind? Reading the OpenSubsonic docs, both endpoints are last.fm-backed and were introduced in the same version:

- [`getArtistInfo`](https://opensubsonic.netlify.app/docs/endpoints/getartistinfo/) — "Returns artist info with biography, image URLs and similar artists, using data from last.fm." Since 1.11.0.
- [`getArtistInfo2`](https://opensubsonic.netlify.app/docs/endpoints/getartistinfo2/) — "Similar to `getArtistInfo`, but organizes music according to ID3 tags." Since 1.11.0.

The [`artistInfo2`](https://opensubsonic.netlify.app/docs/responses/artistinfo2/) response carries the same fields — `biography`, `musicBrainzId` and the three image URLs. As far as I can tell the only differences are the id namespace and the type of the `similarArtist` entries (`Artist` vs `ArtistID3`), not the data source or the provider.

What makes this a bug rather than a preference is the id. `getAlbumArtistInfo` is always called with an id that came from `getArtists.view` / `getArtist.view`, i.e. an ID3 id, and `getArtistInfo` resolves its id in the folder / media-file namespace instead. On Airsonic-Advanced that is `mediaFileService.getMediaFile(id)` versus `artistDao.getArtist(id)` in `getArtistInfo2` (permalinks in the description). The two namespaces overlap numerically, so the lookup succeeds instead of failing and the server returns the last.fm data of an unrelated item — that is #2308.

So, concretely: which scenario do you have in mind where `getArtistInfo` returns correct data for an ID3 artist id? If Feishin had a folder-browsing path that produced folder ids, a toggle would make sense to me, but I could not find one — every call site of this endpoint gets its id from the ID3 endpoints.

If the concern is instead servers that implement `getArtistInfo` but not `getArtistInfo2`, I can add a fallback that only triggers on an error response. I left it out on purpose because `getArtistInfo` does not fail on an ID3 id, it answers with a plausible-looking biography of the wrong artist, so an unconditional fallback would quietly restore the bug.

noctuum:
@kgarner7 Both endpoints return last.fm data - `getArtistInfo2` is just the ID3 variant, that's the only difference in the spec.

I went through every route into the artist page, including "go to artist" on a song listed through the folder browser. On Airsonic all of those ids come from the artist table, so `getArtistInfo` is always called with an id from the wrong namespace - it resolves to a different row and returns someone else's biography. Two curl calls with the same id are in the issue description.

Is there a concrete case where `getArtistInfo` returns correct data for an ID3 id?

PR #2312, opened 2026-08-02
Support Navidrome External Authentication (SSO) with OIDC

Tascate:
I use SSO with passkey-only login for pretty much everything. It's secure and easy so syncing Navidrome Web + Feishin login methods is even more convenient. This PR adds support for Navidrome's [External Authentication](https://www.navidrome.org/docs/getting-started/extauth-quickstart/) allowing API access **without Subsonic credentials.** Authenticate externally **without Feishin seeing usernames or passwords**.

Similar to #1678 but forgoes grabbing cookies and instead, obtains access tokens from an IDP. This needs some additional setup with a middleware, i.e. reverse proxy authentication or api gateway, to verify access tokens and pass user info to Navidrome.

## Server Requirements
1. Existing Navidrome Configuration with [External Authentication](https://www.navidrome.org/docs/getting-started/extauth-quickstart/)
2. Reverse Proxy Authentication **validates acces token** and **forwards user headers**
    - Requires token auth middleware (ex: OAuth2-proxy, Traefik OIDC plugin, NGINX with lua-resty-openidc, Pomerium, Envoy Proxy, etc.)
        - OAuth2-proxy requires `skip-jwt-tokens` and either `set-xauthrequest` or `pass-user-headers`
    - Validates tokens with OIDC provider and passes on user claims to Navidrome
3. OIDC Provider supports PKCE + Public Clients (no client secret)
    - Ex: Authentik, Authelia, Keycloak, etc.
    - Defaults `client_id` to `feishinapp`

## Areas the code touches
SSO is a big change to authentication but to break it down:
1. Add a SSO Navidrome Server-> Finds/Ensures proper OIDC issuer from provided link -> Creates SSO login link for User to sign into and registers callbacks -> Successfully signs in user and fetches+saves access+refresh token for web requests
2. Upon swapping Servers or selecting Servers with SSO -> Refreshes access/refresh token and logs in
3. On API request fails -> Lazily refresh access/refresh token
    - If Refresh token is invalidated during a refresh -> Opens SSO modal so user can complete login flow again
5. Server Logout -> Clears access Token
6. Delete Server -> Clears + Revokes Refresh Token

## So what code does this add?
- External SSO login flow option for Navidrome
    - `openid-client` package handles OIDC logic and fetching tokens
- Validation for SSO servers in Add/Edit server form
- API calls / asset fetches using an access token
- Lazy Refreshes access/refresh token on failed API requests
- Store encrypted refresh tokens (desktop only)
- SSO Reauthentication login flow on expired refresh tokens
    - Opens SSO Modal and handles Reauthentication
- SSO login callback handling for Electron build and Web build

## Caveats
- `/rest` endpoint requires reverse proxy auth, preventing access for other Subsonic clients
    - Configure requests with no Authorization header are let through
        - Other Subsonic clients without SSO will bypass [External Authentication](https://www.navidrome.org/docs/getting-started/extauth-quickstart/)
- Issue with Web Feishin
    - Audio playback and Images currently fails since the browser fetches without authorization headers
        - OAuth2-proxy with same-site cookies should allow playback

## UI

open-add-server
edit-serverpng
sso-modal

Tested working with Generic OIDC  + oAuth2-proxy


jeffvli:
@kgarner7 When you have time, could you review this?

kgarner7:
1. This doesn't really solve the general problem of reverse proxy auth; there can be more than just oauth/oidc provider. It's also completely useless for me specifically
2. At least in the development, this branch introduces an unacceptable amount of latency for every request.

Perhaps other users may find it useful, but I do not.

kgarner7:
Comparing with dev, it appears that the latency appears to be unrelated to the PR; it just sucks in latest `development`. This point can be disregarded (it was throttling)

That being said, I did also observe I had difficulty logging into existing servers when switching to this branch; for whatever reason, the only way I could log into my non-sso app after manually resetting the username/password.

Tascate:
Closing this since this setup is very specific and SSO integration with Navidrome/Subsonic is still very underdeveloped.
Better to support something like [API Key Auth Extension](https://opensubsonic.netlify.app/docs/extensions/apikeyauth/) with compatible Subsonic servers and extend Navidrome to support it (i.e. with a plugin)

PR #2318, opened 2026-08-04
fix(player): stop radio before playing from the queue panel

mihaitom:
## Problem

While a radio stream is playing, selecting a track in the queue panel
and pressing Play does nothing — the track is silently
ignored. Switching back to track playback via a playlist or the track
list in the center panel works fine.

## Root cause

#2039 (f3c0b68a) stopped the active radio stream before starting track
playback, but only in `addToQueueByType`'s `Play.NOW`/`Play.SHUFFLE`
branches — the paths used when playing from a playlist/track list.

The queue panel's row-play action goes through `mediaPlay`/
`mediaPlayByIndex` instead, which update `player.index`/`player.status`
but never touch the radio store. While `isRadioActive` stays true,
`audio-players.tsx` keeps `RadioWebPlayer` mounted instead of the
normal player, and `RadioWebPlayer` only reacts to `currentStreamUrl` —
so the index/status change from the queue is silently ignored.

## Fix

Apply the same radio-stop call from #2039 to `mediaPlay` and
`mediaPlayByIndex` in `player.store.ts`.

## Test plan

- [ ] Play an internet radio station
- [ ] Select a track in the queue panel (right sidebar) and press Play
- [ ] Confirm the track starts playing and the radio stream stops
- [ ] Repeat via double-click on a queue row

### DIFF
--- a/src/renderer/store/player.store.ts
+++ b/src/renderer/store/player.store.ts
@@ -1134,6 +1134,10 @@ export const usePlayerStoreBase = createWithEqualityFn<PlayerState>()(
                 mediaPlay: (id?: string) => {
                     let playIndex: number | undefined;
 
+                    if (useRadioPlayerStore.getState().currentStreamUrl) {
+                        useRadioPlayerStore.getState().actions.stop();
+                    }
+
                     set((state) => {
                         if (id) {
                             const queue = state.getQueue();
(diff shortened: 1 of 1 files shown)


jeffvli:
Thanks for the PR.

I fixed it separately in https://github.com/jeffvli/feishin/commit/12aa000b1bb5b873ab6560dd351f330d30ff55f9.

There was already a [previous PR](https://github.com/jeffvli/feishin/pull/2039) which attempted to solve it the same way but I'd rather not add more radio logic into the player store as it gets messy with all the different playback methods.

Instead I just added a listener to the play event inside the radio player directly which should cleanly handle it.

PR #2337, opened 2026-08-09
Remote control: library browsing, queue management, and action sheets

mihaitom:
Extends the existing Remote control (Settings → Remote) from a single now-playing screen into a full phone-remote app, organized into five tabs (bottom tab bar): Home, Tracks, Playlists, Radio, Queue.

- **Home** — now-playing screen with tap-to-fullscreen artwork (replacing the old show/hide-image toggle button), favorite/rating, and bigger-touch-target volume/position sliders with a mute button that remembers the pre-mute level.
- **Tracks** — searchable, paginated list. Tap a row to play it immediately; long-press opens an action sheet with:
  - **Play** — Now / Next / Last, plus shuffled variants of each
  - **Track Radio** — Now / Next / Last
  - **Add to Playlist** — searchable playlist picker
- **Playlists** — searchable, paginated list. Tap to play; long-press opens the same Play submenu (Now/Next/Last, shuffled variants).
- **Radio** — paginated station list, tap to play.
- **Queue** — live queue view mirroring the desktop queue in real time: drag-to-reorder, swipe-to-remove, tap a row to jump to it, and the same long-press action sheet (Play/Track Radio/Add to Playlist) as the Tracks tab for any queued item.

All of this is served over the existing WS connection via a new request/response protocol (correlated by requestId, with a timeout guard so a client is never left hanging if the desktop never answers). Broadcast messages are now serialized once per message instead of once per connected client, and the initial state/queue/radio snapshot is only sent to a client **after** it authenticates — previously it was sent unconditionally right on connect, which leaked playback state to an unauthenticated client whenever a Remote password was configured.

Before:
IMG_0498

Home View:
IMG_0499

Fullscreen Artwork:
IMG_0500

Tracks:
IMG_0501

Tracks (Search + Action Sheet):
IMG_0502

Playlists:
IMG_0503

Radio:
IMG_0504

Queue:
IMG_0505


src/remote/hooks/use-long-press.ts
jeffvli, line 1: Mantine exposes a [useLongPress](https://mantine.dev/hooks/use-long-press/) hook. Is there anything specific in your implementation that requires it to be custom?

src/remote/pages/queue-page.tsx
```diff
@@ -0,0 +1,102 @@
...
+                                    send({ event: 'queue-jump', uniqueId: item.uniqueId })
+                                }
+                                onLongPress={() => setActiveTrack({ id: item.id, name: item.name })}
+                                onRemove={() => handleRemove(item.uniqueId)}
+                                onReorderDragEnd={() => handleReorderDragEnd(item.uniqueId)}
+                                onReorderDragStart={() => {
+                                    isDraggingRef.current = true;
+                                }}
+                            />
+                        ))}
```
jeffvli: It's worth virtualizing the queue rows. Shouldn't be too hard, just use the existing react-window-v2 package that's already used in the app.

The only thing I'm not sure of is if it works properly with Motion's Reorder component.

src/remote/components/tab-bar.tsx
```diff
@@ -0,0 +1,56 @@
...
+
+interface Tab {
+    icon: React.ReactNode;
+    label: string;
+    to: string;
+}
+
+const TABS: Tab[] = [
+    { icon: <RiHome5Line size={22} />, label: 'Home', to: '/' },
+    { icon: <RiMusic2Line size={22} />, label: 'Tracks', to: '/tracks' },
```
jeffvli: Potentially worth renaming this to "Library" and adding a toggle on the page to switch between a track and album view? You would also need to implement the Album fetcher.

src/remote/hooks/use-long-press.ts
mihaitom: First iteration used mantines useLongPress. I think it had something to do with highlighting text, while longpressing. Its been a few days since I wrote this... 


>    // `user-select: none` (remote-reset.css) should already stop this, but
>    // some mobile browsers/WebViews run their own native long-press-to-select
>    // gesture on an independent timer, not tied to CSS at all — clearing the
>    // selection only at pointerdown and once our own timer fires leaves a gap
>    // where the native gesture can still win (its threshold doesn't have to
>    // match ours) and leave text highlighted after the sheet opens. Watching
>    // `selectionchange` for the whole gesture closes that gap: any selection
>    // the browser creates while a press is in flight gets wiped immediately,
>    // regardless of which timer produced it.

src/remote/components/tab-bar.tsx
mihaitom: When in album view, is there another view for the tracks in that album, or does the whole selected album get in the queue?

jeffvli: Keep it simple with only the whole albums.

src/remote/components/wrapped-slider.tsx
```diff
@@ -1,46 +1,25 @@
...
 import { Text } from '/@/shared/components/text/text';
 
-const PlayerbarSlider = ({ ...props }: SliderProps) => {
+const PlayerbarSlider = ({
+    thumbSize = '1.3rem',
+    ...props
+}: SliderProps & { thumbSize?: string }) => {
     return (
         <Slider
-            styles={{
```
kgarner7, line 10: I'll be honest, I'm not too fond of the changes to the slider here. I can understand the thumb, but I prefer the rest to be the primary color

src/remote/components/remote-container.tsx
```diff
@@ -32,86 +50,167 @@ export const RemoteContainer = () => {
...
+                                    onDoubleClick={() => debouncedSetRating(0)}
+                                    style={{ margin: 'auto' }}
+                                    value={song.userRating ?? 0}
+                                />
+                            </Tooltip>
+                        </div>
+                    )}
+                </Group>
             )}
-            <Group gap={0} grow>
```
kgarner7, line 82: I can understand the appeal for this more compact version of home page, but I personally _greatly_ prefer having the cover art in full size (and, the information in larger format). Additionally, the ability to toggle whether to show an image is preferred.
If you wish to keep this format, I could accept it as an alternative view, although I would still suggest making the image and text larger.

src/remote/components/track-row.tsx
```diff
@@ -0,0 +1,54 @@
...
+    track: RemoteTrackItem;
+}
+
+export const TrackRow = ({ onLongPress, onPlay, track }: TrackRowProps) => {
+    const longPress = useLongPress({ onClick: onPlay, onLongPress });
+
+    return (
+        <ListRow {...longPress}>
+            <Thumbnail
+                fallbackIcon={<Icon icon="emptySongImage" size={18} />}
```
kgarner7: If the thumbnail fails to fetch, then it will be a fallback. Unlike the main page, this does not result in proxying the URL. I would suggest either supporting proxying, and/or suggesting if the user has unresolved images, the user sets the public URL.

src/remote/components/player-image.tsx
```diff
@@ -1,16 +1,19 @@
...
+export const PlayerImage = ({ className, src }: PlayerImageProps) => {
     const send = useSend();
 
     return (
         <img
-            className={styles.container}
+            className={clsx(styles.container, className)}
             onError={() => send({ event: 'proxy' })}
             src={src?.replaceAll(/&(size|width|height)=\d+/g, '')}
         />
```
kgarner7, line 31: Given that the other tags (tracks/playlists/radio) can get the image url using the public URL instead, it might be better to ensure that this flow does the same as well (probably remote server instead). Then, the proxy _could_ be removed (or restricted).

src/remote/pages/tracks-page.tsx
```diff
@@ -0,0 +1,61 @@
...
+        searchTerm: debouncedSearchTerm || undefined,
+    });
+
+    return (
+        <Stack gap="md" p="md">
+            <TextInput
+                onChange={(e) => setSearchTerm(e.currentTarget.value)}
+                placeholder="Search tracks…"
+                value={searchTerm}
+            />
```
kgarner7: I'm not sure if it's mantine or the font, but this is very hard to read in light mode. In dark mode, it's much clearer.

Image

kgarner7 (changes requested):
A few comments. As jeffvli noted, please make sure to virtualize the rows. I stressed the remote using a 38k play queue, and it caused the application to freeze completely.

mihaitom:
Thanks for the review, here's what changed:

@jeffvli
- Kept the custom useLongPress — pointer events, an iOS Safari text-selection guard, click-vs-long-press handling, and haptics, none of which Mantine's hook covers. Known issue: text selection can still stick after a slow release, not fully solved.
- Queue is now virtualized (react-window + a custom Motion-based drag, since Reorder.Group doesn't support virtualization and pragmatic-drag-and-drop has no touch support). Also fixed an O(n²) bug. One visible change: dragging now shows a static drop-indicator line instead of live-cascading the other rows.
- Added a Library/Albums tab (/tracks → /library). Tapping an album plays it immediately.

Screen Shot 2026-08-20 at 16 23 30

@kgarner7
- Sliders now match the desktop playerbar slider (neutral by default, primary on hover/focus/active).
- Thumbnails fall back to a placeholder icon on broken images.
- Image proxying: the image URL passed through the proxy is the same one useRemoteUrl already resolves for direct use — i.e. whatever address the phone can already reach the server on. So proxying it isn't exposing anything new, it's just a fallback for when a direct image request fails (some servers/CORS setups don't like being hit directly from the phone's browser). Wanted to spell that out since it's not obvious from the diff alone.
- Placeholder contrast: looks like a side effect of a color-scheme bug fixed elsewhere in this PR (Mantine's internal scheme wasn't staying in sync with the theme toggle) — should be resolved now.

Also in this round: theme parity fixes (accent color pushed from desktop, the scheme-sync bug above), "confirm queue changes" now actually reaching the phone, auto-reconnect, a persistent mini-player bar, a home view redesign, and a reload bug fix.

Screen Shot 2026-08-20 at 16 22 44

src/remote/pages/playlists-page.tsx
```diff
@@ -0,0 +1,66 @@
...
+                    {items.map((playlist) => (
+                        <PlaylistRow
+                            key={playlist.id}
+                            onLongPress={() => setActivePlaylist(playlist)}
+                            onPlay={() =>
+                                confirmedSend({ event: 'play-playlist', id: playlist.id })
+                            }
+                            playlist={playlist}
+                        />
+                    ))}
```
jeffvli: You should follow the same pattern as the Queue page and virtualize the playlist rows as well.

mihaitom: done

src/remote/pages/tracks-page.tsx
mihaitom: I fixed something in  [mantine](https://github.com/jeffvli/feishin/pull/2337/changes/50c75283f96770849030e4855b7b4bc4efc5dbe4#diff-126a19adb48b0d7187b176af82bc7246301ed11200b3dcfbcab26085bfb897c2R50), but its still a bit hard to read.
It's the default light theme from feishin itself.

jeffvli:
Changes look fine to me.

@kgarner7 I'll leave the final check to you.

kgarner7:
1. For the queue, I see that to remove items you would swipe to remove. On desktop, this behavior isn't immediately obvious (to me, at least), maybe add a remove icon (or a context menu?)
2. In the queue, it looks like if you scroll away from images, they are still fetched when not in viewport. For large queues (or scrolling fast), this can hit ND rate limits pretty quickly. At _least_, having the request be cancelled if done?

kgarner7:
3. For remote operations (adding to the queue), you might want to have some request/response to know when the operation completed (lest you override the queue multiple times).

mihaitom:
1. in the actionsheet is now a remove from queue action. also added a button to clear the queue.
2. Thumbnails now load through the same viewport-gated useNativeImage loader the desktop app already uses for its own image grids (fetch + AbortController instead of a plain ).
3. Added a request/response ack for the queue-mutating remote actions (play/next/last/shuffle variants, track radio, add-to-playlist). The actionsheet shows a spinner while waiting for the response.

kgarner7:
1. For the hold actions, maybe it would make sense to reduce the delay it takes to open the context menu? Or, support opening it via right click (on desktop) as well
3. I can't quite see the request/ack flow happening. Somewhat related, but when opening a context menu to queue up an item, there's a _second_ confirmation for clear queue that's hidden behind the first, that is hard to view. Maybe make that one a _popup_ modal (or rework some other flow so that you don't have both on top of each other?)

paulo-roger:
could it work for the docker app too?

mihaitom:
@paulo-roger 
The remote server needs Electron's main process (it opens a real TCP/WS listen socket, serves files off disk, and talks to the renderer over Electron IPC), none of which exists in the docker/web build. Making it work there would mean running the server as its own separate service, which is a bigger feature than this PR.

Unrelated, but similar idea: I tried a [mobile-optimized view](https://github.com/mihaitom/feishin-connect/tree/feature/mobile-default-view) built into the main app, that looks like this remote control, a while back.

@kgarner7 
1. rightclick noch works on desktop. reduced the longpress delay to 350ms
2. Don`t know how I missed this. Guess I never tried to press play from inside the actionsheet. Should be fixed now. the sheet closes itself when the "discard queue?" confirm comes up and confirm shows a spinner until the ack lands.

kgarner7:
Thanks! Looks good from my side. I think just fix the merge conflicts, and then maybe one final review from @jeffvli?

PR #2356, opened 2026-08-17
feat: add a ListenBrainz Discover page

GraysonCAdams:
A Discover page: ListenBrainz recommendations, weekly playlists, new releases, similar-listener picks and music news, filtered against what you already own and have heard. Owning a record and having heard one turn out to be different questions, and there is no endpoint that answers the second one, so the app walks a listener's full listen history once and keeps it in IndexedDB.

That walk is the slow part. ListenBrainz's documented allowance is 30 requests per 10 seconds, and simply respecting it is not enough: staying inside it the whole way still got the connection dropped after a few dozen pages, with no signal in the headers beforehand. Every request in the app, including the walk, now goes through one shared gate that reads the rate-limit headers and holds requests back before the service does it for us, with the walk yielding to whatever the page itself needs.

Since this was opened: suggestions persist across restarts instead of refetching everything on every cold launch, and the page no longer visibly re-renders through several different subsets of the same row before settling. Each card's second line now varies in wording rather than repeating the same sentence down a whole row, and every card gets a small dismiss control so a bad suggestion never comes back.

image
image
image
image
image


kgarner7:
I appreciate the work, but this is more than I'd be comfortable having right now. If/when plugin support comes in, this would be appreciated.

GraysonCAdams:
Understood, thanks for taking a look. I'll keep maintaining this as my own fork at https://github.com/312-dev/feishin-plus and revisit porting it to a plugin whenever that support lands.

PR #2366, opened 2026-08-19
feat(sharing): add direct service links to Share menu

GraysonCAdams:
The Share menu's "Create share link" action now expands into a submenu:
the same link-creation flow first, then direct links to the same song,
album, or artist on Spotify, Deezer, Tidal, Qobuz, and YouTube - sourced
from MusicBrainz's url-rels rather than guessed, so a service only shows
up when MusicBrainz's own editors actually linked it.

Most libraries have no embedded MusicBrainz id (Picard tagging isn't
universal), so the fallback resolves the artist by name and browses
their whole catalog for a title match instead of text-searching the
title directly - MusicBrainz's search endpoint can't include url-rels,
and a title that repeats across an artist's catalog (radio edits,
remasters, deluxe reissues) returns a different arbitrary sample of the
ties on every search call, so text-searching one candidate at a time
misses the duplicate that actually carries the links about as often as
it finds it.


jeffvli:
Thanks for your work, but unfortunately not a feature that I believe is suitable for Feishin at the moment.
If/when a plugin system is added this would be a good add.

PR #2373, opened 2026-08-22
fix: PowerSaveBlocker not being initialized

mihawk90:
As discussed on Discord:

Fixes an incorrect condition preventing the power save blocker from being initialized.

* adds some user-facing logging in the dev tools for start and stop
* ~~also removes redundant `useEffect()` block.~~ re-added as per Discord
* changes the settings descriptions to clarify what they actually do

Tested locally and the block now appears in the KDE Power Management widget:
image


src/renderer/features/player/hooks/use-power-save-blocker.ts
```diff
@@ -31,11 +31,13 @@ export const usePowerSaveBlocker = () => {
...
 
     useEffect(() => {
-        if (!preventSleepOnPlayback || !preventSuspendOnPlayback) return;
+        if (!preventSleepOnPlayback && !preventSuspendOnPlayback) return;
 
         if (status === PlayerStatus.PLAYING) {
+            console.info('Playback started - starting power save blocker');
             startPowerSaveBlocker();
         } else {
+            console.info('Playback stopped - stopping power save blocker');
```
jeffvli: Use the logger function instead of directly using console.

mihawk90: Fixed. Also changed the previous 2 usages in that file while I was at it.

Confirmed logging works on the artifact too.

PR #2395, opened 2026-08-27
fix(player): correctly restore resume timestamp on app launch

TheSpyder:
Feishin stores the player timestamp in a persistent way, and is capable of rehydrating it, but there are bugs that prevent this. With MPV it resets to zero on quit; the web player does seem to persist correctly but on mount the rendering displays zero. This rendering error does correct itself when playback begins but it's a bit disconcerting. The fixes:

Web player: storing the timestamp when not playing can overwrite the rehydrated timestamp with zero (particularly on mount)
MPV player: When the app quits there is a [media stop with reset false](https://github.com/jeffvli/feishin/blob/aa761b17ef4b122f2dfa49712c3b6945b2f4360c/src/renderer/features/player/audio-player/hooks/use-main-player-listener.tsx#L115). The stop handler was checking for this to avoid a _seek_ to zero, but still setting the store to zero. This should also mean a scrobble on quit sends the correct timestamp.

I use macOS desktop and Jellyfin 10.11.11, I've verified playback resume on both web and MPV. Claude verified the web player on linux using my Jellyfin server. I haven't tested this anywhere else.

### DIFF
--- a/src/renderer/features/player/audio-player/web-player.tsx
+++ b/src/renderer/features/player/audio-player/web-player.tsx
@@ -127,16 +127,16 @@ export function WebPlayer() {
                 return;
             }
 
+            if (usePlayerStoreBase.getState().player.status !== PlayerStatus.PLAYING) {
+                return;
+            }
+
             if (num === 1) {
                 setTimestamp(e.playedSeconds);
             }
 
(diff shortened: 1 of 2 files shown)


kgarner7:
Thanks!

PR #2397, opened 2026-08-28
fix(player): reset seek timestamp on song play

ermmmwafle:
Due to autonext being handled differently than manual next, for whatever reason if you go back to a previous song and autonext progresses back to the song, it will resume to the old timestamp due to state.player.seekToTimestamp not being updated when changing songs most of the time.

an example scenario of this bug would be
1. You are 90 seconds into song 30. 
2. You click song 29. 
3. Let it play until autonext. 
4. Instead starting at the beginning of song 30, it starts 90 seconds in.

This is somewhat of a band-aid fix since the underlying issue is with autonext since it does not happen with manual next.

### DIFF
--- a/src/renderer/features/player/hooks/use-update-current-song.ts
+++ b/src/renderer/features/player/hooks/use-update-current-song.ts
@@ -5,7 +5,11 @@ import { useCallback } from 'react';
 import { api } from '/@/renderer/api';
 import { queryKeys } from '/@/renderer/api/query-keys';
 import { usePlayerEvents } from '/@/renderer/features/player/audio-player/hooks/use-player-events';
-import { updateQueueSong } from '/@/renderer/store/player.store';
+import {
+    uniqueSeekToTimestamp,
+    updateQueueSong,
+    usePlayerStoreBase,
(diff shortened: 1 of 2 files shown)


jeffvli:
I'm unable to reproduce this issue given your steps.

Can you be more specific for each step, what player you're using, possible video reproduction etc?

ermmmwafle:
I am on the web audio player.
Here's a video showcase of me reproducing the issue.
https://github.com/user-attachments/assets/de563d71-8542-416d-adff-ebd1444b683e
I am ~26 seconds in the middle of a song before I click on the previous song. Once the previous song ends and proceeds to the next song (the song i was originally on), it resumes at ~26 seconds in instead of starting from 0.

jeffvli:
Thanks, I was able to reproduce it with your additional details.

Though instead of resetting the seekToTimestamp player state in multiple different handlers which is easy to lose track of, it might make sense to instead add a subscriber which resets the state when the current song changes using the `onCurrentsongChange` callback.

You could probably stick it in the event in the `use-update-current-song.ts` hook with a comment noting the behavior.

PR #2398, opened 2026-08-28
feat: Add scrubbable waveform analysis to the playing track information

booinspace:
Adds a scrubbable waveform to the playing track at the bottom which can be toggled (DEFAULT OFF) in the settings.  This enhances the listening experience by visualising parts of the track like rekordbox does.

* Toggle for the waveform enabled or not during playback
* Fade in/out after analysis is performed
* Circle loader while analysis is going on
* Customisation of the waveform size

Tested and built locally, here are some screenshots of the changes:
Screenshot 2026-08-28 at 18 04 41


jeffvli:
There's already an existing waveform playerbar. Is there a reason you added this (besides the rgb)?

PR #2399, opened 2026-08-29
fix: prevent stale MPV commands during runtime player switches

### LINKED ISSUE #2362: [Bug]: when playing through MPV in feishn app, audio mismatch

- [x] Yes

### App Version

1.15.1

### Music Server and Version

Navidrome 0.63.2

Desktop Linux

### What happened?

When playing music through mpv as Audio player, the audio miss matches, in opus file format perticularly, audio starts with a delay.

In the provided video, if u look closely, when switched to mpv, whenever i switch the song, it starts with previous song position, 
and trust me it plays the prev song `for a reason` even if i switch to `sharat`

https://github.com/user-attachments/assets/51cd3e21-ef31-4144-a39f-ae81ebd86e0f

### Steps to reproduce

1. switch to mpv audio player
2. play an opus file, it will start with a delay
3. after playing, switch to diffrent song, it will  still play the prev song 

### Relevant log output

```shell

```

mvanhorn:
Represent MPV readiness as component state in `MpvPlayerEngine` and keep it false while the old process is being stopped and the new process and queue are initialized. Switching from the web player to MPV while playback is active can briefly control the previous MPV process before the replacement process and queue are ready.

Start a song with the web player, switch to MPV while it is playing, and verify only the selected song is audible, its timestamp advances from the correct position, seek controls work, and the MPV visualizer receives audio; Pause a song in the web player before switching to MPV and verify the fresh MPV queue remains paused until playback is explicitly resumed.

Closes #2362

### DIFF
--- a/src/renderer/features/player/audio-player/engine/mpv-player-engine.tsx
+++ b/src/renderer/features/player/audio-player/engine/mpv-player-engine.tsx
@@ -10,6 +10,7 @@ import { AudioPlayer, PlayerOnProgressProps } from '/@/renderer/features/player/
 import { useRadioStore } from '/@/renderer/features/radio/hooks/use-radio-player';
 import { getMpvProperties } from '/@/renderer/features/settings/components/playback/mpv-properties';
 import {
+    setMpvInitialized,
     usePlaybackSettings,
     usePlayerActions,
     usePlayerSong,
@@ -67,6 +68,7 @@ export const MpvPlayerEngine = (props: MpvPlayerEngineProps) => {
     useEffect(() => {
         const handleMpvReload = () => {
             setIsInitialized(false);
+            setMpvInitialized(false);
             setReloadTrigger((prev) => prev + 1);
         };
 
@@ -89,6 +91,7 @@ export const MpvPlayerEngine = (props: MpvPlayerEngineProps) => {
     useEffect(() => {
         isMountedRef.current = true;
         setIsInitialized(false);
+        setMpvInitialized(false);
         let isCancelled = false;
 
         const initializeMpv = async () => {
@@ -162,6 +165,7 @@ export const MpvPlayerEngine = (props: MpvPlayerEngineProps) => {
 
             if (!isCancelled) {
                 setIsInitialized(true);
+                setMpvInitialized(true);
             }
         };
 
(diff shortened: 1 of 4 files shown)


kgarner7 (changes requested):
Thanks! This _helps_ the issue, but doesn't quite solve it as an unexpected IPC `player-get-audio-devices`  (`mpvPlayer.getAudioDevices()`) can create a temporary MPV during the same time as initialization. Do you think you would be able to apply the same initialize check to the `getAudioDevices()` call as well? (this could be done by say, promoting the initialize to a context).

mvanhorn:
Good catch, and it turned out to be worth fixing on both sides.

The root cause is in main: `player-get-audio-devices` spawns a throwaway mpv whenever the main instance isn't running, and that includes the entire window where `player-initialize` is awaiting `createMpv` — `mpvInstance` is null across that await. So the handler couldn't tell "no mpv" from "mpv is starting right now".

Main now tracks the in-flight create, and the device handler waits for it and reuses the real instance instead of starting a second one beside it. If startup genuinely failed it still falls through to the temporary instance, so the standalone case keeps working.

On the renderer side I took your suggestion, with one deviation worth flagging: `MpvPlayerEngine` sits below the PlayerProvider, so lifting its state into that context meant moving it above the provider. `player.store.ts` already had the pattern — `hydrated` is a top-level runtime flag that `partialize` excludes from persistence — so I added `mpvInitialized` the same way, published it from the engine's four transitions, and gated the device fetch on it. Happy to move it into the context instead if you'd rather keep it out of the store.

One behaviour change: with the gate, the dropdown stays empty until mpv is up rather than listing devices from a throwaway instance. The refresh button recovers it, since `MPV_RELOAD` flips the flag and the effect re-runs.

Locally: typecheck and lint clean.

src/renderer/features/player/audio-player/engine/mpv-player-engine.tsx
```diff
@@ -52,10 +53,10 @@ export const MpvPlayerEngine = (props: MpvPlayerEngineProps) => {
     } = props;
 
     const [internalVolume, setInternalVolume] = useState(volume / 100 || 0);
+    const [isInitialized, setIsInitialized] = useState(false);
```
kgarner7: Given that `mpvInitialized` exists, would it make more sense to just use only that (`useMpvInitialized`) and just externally call `setMpvInitialized` rather than pass it down?

kgarner7 (changes requested):
Small nit. I can confirm that this behavior prevents the phantom MPV, and it's stable when switching between players. Thanks!

mvanhorn:
Good call, done in 1206f22.

The local `isInitialized` useState is gone; the component now reads `useMpvInitialized()` and the existing `setMpvInitialized` calls stay as the single write path, so there's no longer a parallel copy of the same state to keep in sync.

Thanks for confirming the phantom MPV fix holds up across player switches.

kgarner7:
Thanks!

PR #2401, opened 2026-08-29
feat: add basic useNativeImage cache

matildepark:
Right now ~~we don't cache images~~ (we might on network? see edit), ~~so if you return to a view, we refetch everything all over again. We even refetch the same album image over and over within an album view, and given we have a maximum of 6 simultaneous connections (HTTP/1.1) this can prevent users from listening to music until we fetch the same image dozens of times.~~

We don't get Chrome's image caching because `useNativeImage` fetches to a blob and hands `` a fresh `blob:` URL every time; _that_ exists because Jellyfin needs an Authorization header, which `` can't send. The decoded-bitmap cache never hits, and so every load re-decodes the blob over again.

This PR just lets us store an arbitrary number in the module itself (defaulting to 500 images, small) and fetch from that in-memory cache, deleting the least-recently used entries first. The entries are shared across components, and currently processing requests share signatures; this dedupes the 11 fetches to 1 in the case of an 11-song album, and zero if we've loaded it before. We also preserve the current abort-on-unmount logic by counting subscribers on pending requests, aborting the request when there are no more pending subscribers to the request.

Finally we add a guard to avoid malformed images being written to cache by checking for a `Content-Length` match.

edit: I do notice other people mention the 'stale cache' (eg. #2363); and it's possible that we still have an _HTTP_ cache so images can be stale from that. But that doesn't make sense to me given how Feishin works in my use: images flash in and out, nothing is retained, we constantly wait for six images at a time to process ... in whatever case, we still avoid re-decodes and still deduplicate our requests for images across components with this change and the behaviour is preferable. I leave it to reviewers if they have more background on the network / cache stack here.

### DIFF
--- a/src/shared/components/image/use-native-image.ts
+++ b/src/shared/components/image/use-native-image.ts
@@ -9,6 +9,12 @@ interface NativeImageState {
     status: 'error' | 'idle' | 'loaded' | 'loading';
 }
 
+interface PendingRequest {
+    controller: AbortController;
+    promise: Promise<string>;
+    subscribers: number;
+}
+
 interface UseNativeImageArgs {
     enabled: boolean;
     fetchPriority?: FetchPriority;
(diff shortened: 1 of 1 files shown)


matildepark:
Hmm, found an easier tact. I'll stage that instead.

PR #2402, opened 2026-08-29
fix: dedupe fetching art by album id for Navidrome users

matildepark:
We fetch cover art for songs by song ID, meaning we fire off [album length] number of requests when loading an album, or if we have the same album in a playlist, we wait to load art over and over again for every song on the album. But since we have the album ID in question available to request by, we can deduplicate the requests to just the album; if no album, we fall back to song ID.

### DIFF
--- a/src/shared/api/navidrome/navidrome-normalize.ts
+++ b/src/shared/api/navidrome/navidrome-normalize.ts
@@ -263,7 +263,7 @@ const normalizeSong = (
             songCount: null,
         })),
         id,
-        imageId: id,
+        imageId: item.albumId || id,
         imageUrl: null,
         lastPlayedAt: normalizePlayDate(item),
         lyrics: item.lyrics ? item.lyrics : null,


kgarner7 (changes requested):
You can make it an _option_, but for some, the track art is **different** from the album art, so this is a significant loss.

matildepark:
Suit yourself.

kgarner7:
You didn't have to close it. If you made this configurable (e.g., an option "only show album art") and generic, it'd be fine. As it is, you would break any user who had separate track id.

matildepark:
I don't think it makes any sense as an option (and I don't think we need _more_ options in this application). I find the current behaviour a significant hamper to my own experience and spent the morning trying to find a mitigation that made the most sense. I have a library of 10,000 songs (not a lot, but not nothing) and have never encountered the idea of albums making per-track art. I do not deny that it might exist. If what you're saying is true, making it a global option would be nonsensical, and making it per-album even less.

PR #2404, opened 2026-08-30
feat: exclude other artist from artist play

### LINKED ISSUE #2023: [Feature]: exclude other artists from play/shuffle artist

- [x] Yes

Not server-specific

Right know if I press or hold Play on the artist page, to play/shuffle all artist tracks, it will include all albums from the "Appears on" category in their entirety.

Image

Image

However, I think it'll be much better and expected to include only specific tracks which relate to the artist in question and filter out the rest.

Here's an example:
Image
1. Valid candidate: album is by another artist, but the track relates to "Hocico"
2. Invalid candidate: album is by another artist and track has no relation to "Hocico"

This behavior also would be extra useful for large artist discographies with various releases types, where you can't easily play everything except for "appears on".

GerardGascon:
## Summary

When clicking play on an artist page, it shuffles all albums available. The problem is that the artist could have a single song present in a large album from another artist, filling the queue with songs that aren't from the artist you want to play.

This PR adds a filter in the album play button that excludes any song that doesn't match these conditions:
1. The album artist is present in the song's album artists list.
2. The album artist is present in the song's artist list.

Any other case is treated as unrelated and therefore excluded from the queue.

Closes #2023

### DIFF
--- a/src/renderer/features/artists/components/album-artist-detail-header.tsx
+++ b/src/renderer/features/artists/components/album-artist-detail-header.tsx
@@ -171,21 +171,20 @@ export const AlbumArtistDetailHeader = forwardRef<HTMLDivElement, AlbumArtistDet
                 const albumIds = flatSortedAlbums.map((album) => album.id);
                 if (albumIds.length === 0) return;
 
-                const isSongByArtist =
-                    (artistId?: string) =>
-                    (song: Song): boolean => {
-                        return (
-                            song.albumArtists.some((artist) => artist.id === artistId) ||
-                            song.artists.some((artist) => artist.id === artistId)
-                        );
-                    };
+                const filter = (song: Song) => {
+                    if (song.albumArtists.some((artist) => artist.id === albumArtistId)) {
+                        return true;
+                    }
+
+                    return song.artists.some((artist) => artist.id === albumArtistId);
+                };
 
                 addToQueueByFetch(
                     server.id,
                     albumIds,
                     LibraryItem.ALBUM,
                     type || playButtonBehavior,
-                    isSongByArtist(albumArtistId),
+                    { filter },
                 );
             },
             [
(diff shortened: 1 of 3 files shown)


src/renderer/features/artists/components/album-artist-detail-header.tsx
```diff
@@ -169,15 +174,29 @@ export const AlbumArtistDetailHeader = forwardRef<HTMLDivElement, AlbumArtistDet
...
+
+                const filteredArtistSongs = songs.filter((song) => {
+                    if (song.albumArtists.some((artist) => artist.id === albumArtistId))
+                        return true;
+
+                    return song.artists.some((artist) => artist.id === albumArtistId);
+                });
+
+                addToQueueByData(filteredArtistSongs, type || playButtonBehavior);
             },
```
jeffvli: I'd prefer not to include a direct async fetch here as using the addToQueueByFetch allows for cancellation on long queries (which may occur in this case as an album artist can have hundreds of albums). 

I'm open to adding an additional `options` arg to addToQueueByFetch which accepts a `filter` callback which you can provide to filter out songs based on condition.

```typescript
const filter = (song: Song) => {
    if (song.albumArtists.some((artist) => artist.id === albumArtistId)) {
        return true;
    }

    return song.artists.some((artist) => artist.id === albumArtistId);
}

addToQueueByFetch(serverId, id, itemType, type, { filter });

```

GerardGascon: Refactored this into passing the additional filter inside an options arg.

src/renderer/features/player/context/player-context.tsx
```diff
@@ -54,6 +55,8 @@ export interface PlayerContext {
         id: string[],
         itemType: LibraryItem,
         type: AddToQueueType,
+        options?: AddToQueueOptions,
+        additionalFilter?: (song: Song) => boolean,
```
jeffvli: I think you forgot to remove the extra `additionalFilter` param here.

GerardGascon: Whops, totally forgot to remove that yes. Fixed it.

PR #2407, opened 2026-08-31
fix: radio cover and metadata was incorrect/missing in some cases

### LINKED ISSUE #2378: [Bug]: Radio cover image display problems

- [x] Yes

### App Version

1.15.1

### Music Server and Version

Navidrome v0.63.2

Desktop Windows

### What happened?

I've started adding internet radios to my Navidrome setup but I've noticed some bugs with the Feishin implementation:
~~1. When editing a radio station if you click save without modifying the cover, the cover gets removed from the radio.~~ EDIT: This is an issue with Navidrome, not Feishin (https://github.com/navidrome/navidrome/issues/6057#issue-5299242877)
2. Feishin displays the radio cover in the corner, but it doesn't display the cover in the full size screen.
3. In Discord's RPC, the cover isn't displayed even though my setup is publicly accessible and I've checked that the image can be queried from outside internet. Also in the RPC I noticed that no matter the option selected it only displays the artist name while listening to a radio.

### Steps to reproduce

1. Create a radio on Navidrome and assign a cover
2. Try the things explained in the 'what happened' section and see the results.

### Relevant log output

```shell

```

GerardGascon:
## Summary

This PR fixes a couple of issues with internet radio metadata display:

1. Now the Discord RPC shows the currently playing radio following the display type config. Previosuly it was being ignored and only showed the current song artist.
2. Also, the Discord RPC now shows the radio cover image pulled from the server (in Navidrome setups).
    image

4. Finally, full screen player now shows the radio cover image.
    image

Closes #2378


kgarner7 (changes requested):
Calling `useItemImageUrl` (ultimately, `/rest/getCoverArt`) for the radio imgae in discord RPC is not safe; this results in user credentials being embedded in the discord rpc URL that anyone can see.

GerardGascon:
My bad, didn't notice that. I've reverted that change, but at the same time I've noticed that the existing implementation for song images also embeds the credentials in the URL.

kgarner7:
The large image key is only expected to be set if jellyfin (no credentials in the api) or if it's ND (or Subsonic, I will remove this case) where getAlbumInfo2 is expected to (guaranteed to in ND case) return a sharable URL (https://github.com/jeffvli/feishin/blob/694e36e731cbd9fa0bcd28cec1bdc6694d5b01f2/src/renderer/features/discord-rpc/use-discord-rpc.ts#L268).
If there are other cases I missed, please feel free to fix/point out those as well, thanks!

GerardGascon:
Oh okay, I understand, thanks for the clarification.

kgarner7 (changes requested):
Rather than make all of the other changes, to get the radio image to show properly, it would be better to just call

```typescript
    const radioImage = useItemImageUrl({
        id: currentRadioStationArt?.imageId || undefined,
        itemType: LibraryItem.RADIO_STATION,
        serverId: currentRadioStationArt?.serverId,
        type: 'fullScreenPlayer',
    });
```
and then pass that in to the `isPlayingRadio` `src`. This also results in less changes for other sources.

GerardGascon:
Alright, refactored this. Also, while doing this I've noticed that currently the `isPlayingRadio` is defined with this:

```ts
const isPlayingRadio = isRadioActive && isRadioPlaying;
```

which makes the image not show while the player is paused. I don't know if there's a special reason behind this. If not, I'd rather just use the `isRadioActive` boolean so it gets displayed while paused.

kgarner7:
> Alright, refactored this. Also, while doing this I've noticed that currently the `isPlayingRadio` is defined with this:
> 
> ```ts
> const isPlayingRadio = isRadioActive && isRadioPlaying;
> ```
> 
> which makes the image not show while the player is paused. I don't know if there's a special reason behind this. If not, I'd rather just use the `isRadioActive` boolean so it gets displayed while paused.

Sorry for the delayed response. I'd be to remove this check (only show the radio image and not alternate between radio and queued track), unless @jeffvli has other thoughts. It feels weird for me as well.

jeffvli:
You can change and see if there are any adverse effects. I believe I initially implemented it that way because the radio player was treated as a special app state as opposed to only a player so I needed both conditions.

GerardGascon:
Alright, changed all those fields and apparently there are no adverse effects.

In the meantime I have also fixed some mobile player issues related to this, where things like the seek bar was being shown during radio playback.

src/renderer/features/player/components/mobile-playerbar.tsx
```diff
@@ -70,6 +78,7 @@ export const MobilePlayerbar = () => {
                     <AnimatePresence initial={false} mode="popLayout">
                         {currentSong?.id && (
                             <div className={styles.imageWrapper}>
+                                {/*TODO: Radio image here*/}
```
kgarner7: Why is there a `todo` still here?

kgarner7 (changes requested):

src/renderer/features/player/components/mobile-playerbar.tsx
GerardGascon: Whoops, completely forgot to remove that, thanks.

kgarner7:
Thanks for your work!

PR #2414, opened 2026-09-02
feat: add scrobble minimum mode setting

ermmmwafle:
Setting allows you to only scrobble after a % of a song has passed, a duration in seconds has passed, or either one has passed (original functionality)

To make this guy happy
image


src/renderer/features/settings/components/general/scrobble-settings.tsx
```diff
@@ -88,6 +93,46 @@ export const ScrobbleSettings = memo(() => {
...
+                            },
+                        })
+                    }
+                />
+            ),
+            description: t('setting.scrobbleMinimumMode', {
+                context: 'description',
+            }),
+            title: t('setting.scrobbleMinimumMode'),
+        },
```
kgarner7: 1. This probably makes sense _above_ the other minimal scrobble configurations
2. If the configuration is not `BOTH`, it would make sense to hide the not applicable setting

kgarner7 (changes requested):

src/renderer/features/settings/components/general/scrobble-settings.tsx
ermmmwafle: done

kgarner7:
Thanks!

PR #2417, opened 2026-09-03
fix(subsonic): load artist favorite songs via getStarred

### LINKED ISSUE #2415: [Bug]: Subsonic adapter: artist page "Favorite Songs" omits starred tracks that are present in the artist's track list (Navidrome-native adapter unaffected)

- [x] Yes

### App Version

1.15.1

### Music Server and Version

Navidrome 0.63.2 (be10f89c)

Desktop Linux

### What happened?

Using the Subsonic server type, an artist's "Favorite Songs" section (via "View all") omits a track that is genuinely starred and attributed to that artist. Adding another favorite from the same album shifts which track is shown rather than increasing the count — suggesting a bounded/limited query or a client-side filter that misses tracks.

Key discriminator: The bug occurs only when connected via the Subsonic interface. Connecting the same server via the Navidrome-native interface shows the favorite correctly. This points to the Subsonic API adapter's logic for building an artist's favorited songs, not to server data.

Innocent causes ruled out:

Starred on the Navidrome server directly; heart persists across restart.
Appears in a library-wide favorites smart playlist (Is Favorite = true) — 7 there vs 6 in the artist section.
Appears in the Tracks view with the favorite filter.
Appears under the same artist page's "View all tracks," shown favorited — so it's attributed to this artist.
Missing only from the Subsonic-adapter artist page's "Favorite Songs" section.

Image
Image

### Steps to reproduce

1. Song A (album X, this artist) is starred but absent from the artist Favorite Songs section.
2. Star song B, also from album X.
3. A now appears; B is now missing. Count does not increase.
4. Persists after a full restart (not cache-on-action).

### Relevant log output

```shell

```

BetterAndBetterII:
## Summary

On the Subsonic adapter, artist page **Favorite Songs** called `getSongList` with `artistIds` only. That path album-scrapes via `getArtist` + `getAlbum`, then client-filters `userFavorite`. Starred tracks that are not present on the albums returned by `getArtist` are omitted (Navidrome-native is unaffected because it uses a different list API).

`getSongList` already supports `favorite: true` → `getStarred` plus an `albumArtists`/`artistIds` filter. The favorites branch now passes `favorite: true` (keeping `artistIds`) so starred songs for that artist come from `getStarred`.

## Test

```bash
node src/renderer/api/subsonic/subsonic-controller.get-favorite-songs.test.mjs
```

Fixes #2415

### DIFF
--- a/src/renderer/api/subsonic/subsonic-controller.get-favorite-songs.test.mjs
+++ b/src/renderer/api/subsonic/subsonic-controller.get-favorite-songs.test.mjs
@@ -1,0 +1,80 @@
+/**
+ * Regression for jeffvli/feishin#2415 — Subsonic artist Favorite Songs.
+ *
+ * Without favorite:true, getFavoriteSongs album-scrapes via getArtist+getAlbum and
+ * can omit starred tracks that getStarred still returns. Run:
+ *   node src/renderer/api/subsonic/subsonic-controller.get-favorite-songs.test.mjs
+ */
+/* eslint-disable @typescript-eslint/explicit-function-return-type */
+import assert from 'node:assert/strict';
+import { readFileSync } from 'node:fs';
+import { dirname, join } from 'node:path';
+import { fileURLToPath } from 'node:url';
+
+const __dirname = dirname(fileURLToPath(import.meta.url));
+const controllerSrc = readFileSync(join(__dirname, 'subsonic-controller.ts'), 'utf8');
+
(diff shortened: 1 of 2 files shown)


src/renderer/api/subsonic/subsonic-controller.get-favorite-songs.test.mjs
jeffvli, line 1: Please remove this test file

jeffvli (changes requested):

PR #2420, opened 2026-09-04
Fix macOS menu bar disappearing on screen wake

york9675:
On macOS, when the display wakes from sleep while the window is unfocused (e.g. on a secondary display), the menu bar can visually disappear.

Resolved by subscribing to `NSWorkspaceScreensDidWakeNotification` via `systemPreferences` to re-apply the application menu whenever the screen wakes, with proper cleanup when the window is closed.

### DIFF
--- a/src/main/menu.ts
+++ b/src/main/menu.ts
@@ -1,4 +1,10 @@
-import { BrowserWindow, Menu, MenuItemConstructorOptions, shell } from 'electron';
+import {
+    BrowserWindow,
+    Menu,
+    MenuItemConstructorOptions,
+    shell,
+    systemPreferences,
+} from 'electron';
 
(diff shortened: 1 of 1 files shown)


york9675:
Closing this PR as the issue still persists in production and I'm currently unable to fix it at this stage.

Worth noting: the current latest stable release (v1.15.1) does not have this issue, but it started appearing on the `dev` branch right after upgrading to Electron 43, so I suspect this is an upstream Electron issue. Additionally, this isn't limited to waking the display from sleep—it also happens simply when unlocking from the lock screen.

york9675:
Closing as unrelated. The macOS menu-bar regression was traced to #2269 and is being addressed separately.

PR #2421, opened 2026-09-05
feat(player): add a synced music video panel

GraysonCAdams:
Feishin can already show cover art, lyrics or a visualizer while a track plays. This adds a fourth option: the track's music video, played muted and lined up with whatever engine is producing sound, so playback itself is untouched and the video is only a picture.

A panel opens from a new playerbar button. yt-dlp searches YouTube, and each candidate is scored by landmark fingerprinting against the local file rather than trusted on a title match. That scoring is also where the playback offset comes from, so a video with a twelve second intro starts twelve seconds in rather than at zero. A candidate that turns out to be one album-art still held for the whole song is rejected on a detached element before anything is shown and the next one is tried; a track with no confident match falls back to the configured content.

Sync runs against `timestamp.store`, so it never needs to know which engine is playing. Small errors are closed by running the muted video a few percent fast or slow, which is invisible where a seek is not, and only errors past 700ms are worth the jump. The panel is its own surface rather than a tab in the full-screen player because Chromium closes a picture-in-picture window when the element it is bound to unmounts, and a tab strip unmounts on every switch.

Cached files are served through a `feishin-video` protocol handler with byte-range support, registered next to the existing `feishin` font scheme, so nothing about how the renderer itself is served changes. The fingerprint runs in a worker, and the main process extracts audio with ffmpeg already at the rate the matcher works in, so nothing needs resampling on the main thread and a few thousand FFTs do not compete with a video that is playing.

yt-dlp and ffmpeg are fetched into userData when the feature is first enabled rather than bundled, since it defaults to off and most installs will never need them. An explicit path override is there for anyone who has their own.

## Test plan

- `pnpm run typecheck`, eslint and stylelint pass.
- Enable it in Settings, play a track with a well known video, and confirm it loads muted and lined up while the app's own audio is unaffected.
- Seek and pause from the playerbar, and confirm the video follows.
- Pop the video out, confirm it keeps playing, and confirm closing that window hands playback back to the panel rather than stopping it.
- Play a track whose only YouTube uploads are album-art stills and confirm the fallback rather than a still picture.

Exercised on Windows with mpv against Navidrome. Not yet run on macOS or Linux, or against the web player.


kgarner7:
I am _very uncomfortable_ integrating yt-dlp in any form. Perhaps in the future something that uses local video, but not in this form.

GraysonCAdams:
Totally understand, and thanks for taking a look. That's a fair line to draw. I'll keep it going in my fork for anyone who wants it: https://github.com/312-dev/feishin-plus

GraysonCAdams:
Feel free to use this wrapper code for a local system should you find a less grey-area solution for music videos.

Issue #2111, opened 2026-06-03
[Bug]: multi-disc albums not being picked up (only the first disc is picked up)


elendil7:
### I have already checked through the existing (both open AND closed) bug reports and found no duplicates

- [x] Yes

### App Version

1.13.0

### Music Server and Version

Jellyfin version 10.11.10

### What local environments are you seeing the problem on?

Desktop Windows

### What happened?

As per the title, multi-disc albums not being picked up (only the first disc is picked up). This applies to all albums i've tested; only the first disc's tracks are displayed in feishin.

See image; only CD1 folder is detected despite album also having CD2 folder:
<img width="1390" height="1197" alt="Image" src="https://github.com/user-attachments/assets/8d25753a-a5e9-43bf-a1b4-98916874f2f4" />

See image for folder structure (synced to server where jellyfin resides via sftp):

<img width="959" height="275" alt="Image" src="https://github.com/user-attachments/assets/c369d6fd-e466-4cbc-ac2f-e05ca01d1389" />

<img width="1026" height="759" alt="Image" src="https://github.com/user-attachments/assets/e2bc3cd7-5e99-4239-8bf3-5547c747589b" />

<img width="1020" height="756" alt="Image" src="https://github.com/user-attachments/assets/cd055c15-4264-4132-9412-30f2066f537d" />

### Steps to reproduce

N/A

### Relevant log output

```shell

```

jeffvli:
Can you check the Jellyfin web ui to see if the songs are missing from there as well?

elendil7:
Yes, the jellyfin web ui contains all tracks from both CDs

<img width="2509" height="1036" alt="Image" src="https://github.com/user-attachments/assets/e61f8cff-819f-4420-b1e8-13c1e95e48cf" />

<img width="2290" height="1245" alt="Image" src="https://github.com/user-attachments/assets/f7c01813-7354-4302-b4f9-04b3392c36a6" />

elendil7:
Bump

daSmirnov:
Just ran into same issue.  In my case files are all in same folder.  And separated with the 'part of set' metadata, 1 or 2 depending on which disc.  Jellyfin interface separates into disc 1 and 2 as above.  But Feishin only showing disc 1.

Issue #2041, opened 2026-05-18
[Bug]: Albums from different artists with the same titles are grouped together


lummolulu:
### I have already checked through the existing bug reports and found no duplicates

- [x] Yes

### App Version

1.11.1-alpha-20260516

### Music Server and Version

Jellyfin 10.11.8

### What local environments are you seeing the problem on?

Desktop Windows

### What happened?

If I have two or more albums from different artist but sharing the same title, they are grouped together as if they're from the same album. 

Here's a screenshot of this happening.

<img width="865" height="628" alt="Image" src="https://github.com/user-attachments/assets/2be96bb9-9fa2-4268-a520-d6eb2b800d69" />

This doesn't seem to be a metadata issue, since on Jellyfin they are treated like different releases.

<img width="563" height="406" alt="Image" src="https://github.com/user-attachments/assets/2f63f539-c40a-4b34-8045-4b8c197a34e7" />

<img width="976" height="752" alt="Image" src="https://github.com/user-attachments/assets/4fd69968-78b0-4d2f-909d-5c4cd0b3a652" />

<img width="842" height="580" alt="Image" src="https://github.com/user-attachments/assets/afd2618e-efaf-42c9-81f5-5ee84996ae58" />

Weirdly enough, if you search for "SPIRAL" on Feishin, it will show the two albums, not one. But both of them share the same songs.

### Steps to reproduce

1. Have two releases from different artists sharing the same title
2. Search for one of the releases
3. The release will have songs from the other release with the same name.

### Relevant log output

```shell

```

jeffvli:
Related: https://github.com/jeffvli/feishin/issues/287

PR #2422, opened 2026-09-05
fix(api): preserve Jellyfin single album songs

### LINKED ISSUE #2076: [Bug]: Singles can't be played nor does not appear in UI

- [x] Yes

### App Version

1.12.1

### Music Server and Version

Jellyfin 1.10.7

Desktop Linux

### What happened?

Singles don't have the have the songs inside the "album", used to work before it auto updated, pressing play doesn't play anything

Image

Image

Image

Tested in jellyfin web it does show and does play the songs correctly

### Steps to reproduce

1. Go to any album with only one song or is a single
2.  It appears as if there aren't any songs in it

### Relevant log output

```shell

```

atirna:
## Summary

Preserves the songs returned for a Jellyfin album detail request.

## Problem

The stale `AlbumId` filter removes tracks from some singles even though the request is already scoped by `ParentId`.

## Fix

Pass the parent-scoped response to album normalization unchanged.

Fixes #2076

## Testing

- `pnpm lint:staged`
- `pnpm typecheck`

### DIFF
--- a/src/renderer/api/jellyfin/jellyfin-controller.ts
+++ b/src/renderer/api/jellyfin/jellyfin-controller.ts
@@ -546,11 +546,10 @@ export const JellyfinController: InternalControllerEndpoint = {
             throw new Error('Failed to get album detail');
         }
 
-        // Workaround for Jellyfin bug that returns items that share the same album name
-        const albumIdSet = new Set([query.id]);
-        const songs = songsRes.body.Items.filter((item) => albumIdSet.has(item.AlbumId!));
-
-        return jfNormalize.album({ ...res.body, Songs: songs }, apiClientProps.server);
+        return jfNormalize.album(
+            { ...res.body, Songs: songsRes.body.Items },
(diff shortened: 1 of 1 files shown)


jeffvli:
Can you confirm that the bug noted in the comment for the lines you changed is fixed on Jellyfin?

Because if you remove those lines, that will reintroduce a different bug.

atirna:
the comment's bug was about the old query, not this one.

the filter (d9da588c, #2041) covered `AlbumIds: query.id`, which Jellyfin matches by id or album name (the getSongList bodge documents the same), so same-named releases leaked. bf501da6 then switched this request to `ParentId: query.id` for #2111: descendants of the album item only, no name matching. since then the filter could only drop songs ParentId legitimately returns: the #2076 singles have the album as parent, but their AlbumId field points at their own single entry.

nothing needed fixing jellyfin-side, the workaround was for a parameter this request stopped using in august.

PR #2440, opened 2026-09-10
feat: avoid triggering auto dj when repeat all is enabled

### LINKED ISSUE #2437: [Bug]: "Repeat All" doesn't loop when playing album

- [x] Yes

### App Version

1.15.1

### Music Server and Version

Navidrome 0.62.0 (1b46b977)

Desktop Linux

### What happened?

When I play an album while "Repeat All" is selected, the playback will not loop around to the first song of the album after the last one was played.
Instead, random songs from the library will be played, the same behaviour as when "Repeat disabled" is selected.
Furthermore, when repeat is disabled, I would expect playback to stop after the final track, not play random songs from the library.

[This issue comment](https://github.com/jeffvli/feishin/issues/2130#issuecomment-4933287513) further indicates that the behaviour observed here is indeed incorrect.

Finally, I just want to say that Feishin is awesome and beats all the other players I have tried!

### Steps to reproduce

1. View an Album (click on it)
2. Click "Play" (next to "Next" and "Last")
3. Enable "Repeat All" in the player interface
4. Skip to the last song in the queue and then skip to shortly before the end (or let it play naturally until this point)
5. Observe that once the song finishes, random songs from the library are added to the queue and are played

(Steps 2 and 3 can be swapped)

### Relevant log output

```shell

```

GerardGascon:
## Summary

When the user has selected to repeat the current queue (repeat all), Auto DJ shouldn't be triggered, as this would make the queue to continue growing and never repeat.

Fixes #2437

### DIFF
--- a/src/renderer/features/player/hooks/use-auto-dj.ts
+++ b/src/renderer/features/player/hooks/use-auto-dj.ts
@@ -5,6 +5,7 @@ import { eventEmitter } from '/@/renderer/events/event-emitter';
 import { runAutoDjAlbumIds } from '/@/renderer/features/player/auto-dj/auto-dj-albums';
 import { runAutoDjSongs } from '/@/renderer/features/player/auto-dj/auto-dj-songs';
 import { useIsPlayerFetching, usePlayer } from '/@/renderer/features/player/context/player-context';
+import { usePlayerRepeat } from '/@/renderer/store';
 import {
     AUTO_DJ_STRATEGY,
     isShuffleEnabled,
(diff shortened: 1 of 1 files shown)


src/renderer/features/player/hooks/use-auto-dj.ts
```diff
@@ -64,6 +67,10 @@ export const useAutoDJ = () => {
                     return;
                 }
 
+                if (playerRepeat == PlayerRepeat.ALL) {
+                    return;
+                }
+
```
jeffvli, line 73: Might as well include PlayerRepeat.ONE here as well.

jeffvli (changes requested):

src/renderer/features/player/hooks/use-auto-dj.ts
GerardGascon: I'm unsure if that should be done. As repeat one only repeats the currently playing song meaning that the issue described doesn't apply here.

Also, while repeat one automatically loops the current track, you can still manually skip to next track. So I find ok to keep the Auto DJ running for the case where you manually skip the song. I don't know what you think about that.

jeffvli: The more I think about it the more I'd rather not apply this change. I consider the AutoDJ to be a player behavior override, and preventing it from triggering based on the player loop setting is counter to that.

GerardGascon: Okay, understandable.

Issue #1990, opened 2026-04-29
[Bug]: Playerbar slider gets stuck often


Aabyonis:
### I have already checked through the existing bug reports and found no duplicates

- [x] Yes

### App Version

1.11.0

### Music Server and Version

Jellyfin 10.10.7

### What local environments are you seeing the problem on?

Desktop Windows

### What happened?

https://github.com/user-attachments/assets/34190be8-1156-497c-8cc9-cd0ece39017a

### Steps to reproduce

playing a song sometimes is enough really

### Relevant log output

```shell

```

jeffvli:
Is that with mpv or the web player?
I can't say that I have seen this issue happen regularly.

Aabyonis:
> Is that with mpv or the web player? I can't say that I have seen this issue happen regularly.

Oh i was using the web player, i tried using MPV but for some reason nothing plays anymore?

edit: im stupid i didnt have MPV downloaded, ill try again

Allinen:
+1 still an issue (web player). Happens to me very often

PR #2441, opened 2026-09-10
fix: seeker stuck

Allinen:
Related issue: https://github.com/jeffvli/feishin/issues/1990
Fixes the seeker / playbar getting stuck. Reproducable by starting playback and immediately seeking to the beginning of the track. Seemingly mainly affecting firefox / gecko based browsers (workes fine on chromium). Tested working on waterfox and chromium.

Worth testing further, maybe there is some specific edge-case I didn't think of where it still fails (or maybe a regression and old working stuff broke)


jeffvli:
Thanks!

Issue #2446, opened 2026-09-12
[Feature]: Refresh lyrics on web client


Allinen:
### I have already checked through the existing feature requests and found no duplicates

- [x] Yes

### Is this a server-specific feature?

Not server-specific

### What do you want to be added?

Currently the desktop client has a "Clear"/"Refresh" button which can re-fetch the lyrics from the server in case they get updated, but there is no equivelant on the web client to re-fetch and update the cached lyrics.

PR #2447, opened 2026-09-12
feat: refresh lyrics on web client

Allinen:
Related issue: https://github.com/jeffvli/feishin/issues/2446

Adds the feature to refresh / re-fetch lyrics on the web client. In web client context,  the "Clear" text doesn't make sense, so it is always "Refresh".  Tested working on desktop and waterfox. 
Note: because there is no "search" button on the web version, it is not centered as cleanly 
image
Compare to desktop client 
image
If this is an issue, it can be addressed.

### DIFF
--- a/src/renderer/features/lyrics/lyrics-actions.tsx
+++ b/src/renderer/features/lyrics/lyrics-actions.tsx
@@ -31,6 +31,7 @@ interface LyricsActionsProps {
     languages: { label: string; value: string }[];
     offsetMs: number;
     onExportLyrics: () => void;
+    onRefreshLyric: () => void;
     onRemoveLyric: () => void;
     onSearchOverride: (params: LyricsOverride) => void;
     onToggleOverlayLayer?: (key: string) => void;
@@ -72,6 +73,7 @@ export const LyricsActions = ({
     languages,
     offsetMs,
     onExportLyrics,
+    onRefreshLyric,
     onRemoveLyric,
     onSearchOverride,
     onToggleOverlayLayer,
@@ -297,14 +299,14 @@ export const LyricsActions = ({
                     }}
                     variant="subtle"
                 />
-                {isDesktop && sources.length ? (
+                {sources.length ? (
                     <Button
                         disabled={isActionsDisabled}
-                        onClick={onRemoveLyric}
+                        onClick={isDesktop ? onRemoveLyric : onRefreshLyric}
                         uppercase
                         variant="subtle"
                     >
-                        {hasLyrics ? t('common.clear') : t('common.refresh')}
+                        {hasLyrics && isDesktop ? t('common.clear') : t('common.refresh')}
                     </Button>
                 ) : null}
                 {isDesktop && sources.length && onTranslateLyric && !hasServerTranslationLayer ? (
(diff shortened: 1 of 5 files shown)


src/renderer/features/lyrics/lyrics.tsx
```diff
@@ -390,6 +391,23 @@ export const Lyrics = ({ fadeOutNoLyricsMessage = true, settingsKey = 'default'
         await queryClient.invalidateQueries({ queryKey: lyricsKey });
     }, [currentSong, lyricsKey]);
 
+    const handleOnRefreshLyric = useCallback(async () => {
+        if (!currentSong || !lyricsKey) return;
+
+        clearRemoteLyricsCache(currentSong.id);
```
jeffvli: Can you explain your intention behind this?

The web client does not have access to the electron main / preload so this function would do nothing.

Allinen: oh you're right, at first I had a separate button for the refresh that was also on the desktop client but that was stupid so I removed it and just combined them. That was just leftover code the desktop version was calling

src/renderer/features/lyrics/lyrics.tsx
```diff
@@ -390,6 +390,21 @@ export const Lyrics = ({ fadeOutNoLyricsMessage = true, settingsKey = 'default'
...
+            prev
+                ? {
+                      ...prev,
+                      remoteAuto: null,
+                      suppressRemoteAuto: false,
+                  }
+                : prev,
+        );
+        await queryClient.invalidateQueries({ queryKey: lyricsKey });
+    }, [currentSong, lyricsKey]);
```
jeffvli: After double checking, there's nothing functionally different from `handleOnRemoveLyric`. The `suppressRemoteAuto: true` only applies towards external lyrics fetching (which is Desktop-only). You can remove this.

src/renderer/features/lyrics/lyrics-actions.tsx
```diff
@@ -297,14 +299,14 @@ export const LyricsActions = ({
...
+                {sources.length ? (
                     <Button
                         disabled={isActionsDisabled}
-                        onClick={onRemoveLyric}
+                        onClick={isDesktop ? onRemoveLyric : onRefreshLyric}
                         uppercase
                         variant="subtle"
                     >
-                        {hasLyrics ? t('common.clear') : t('common.refresh')}
+                        {hasLyrics && isDesktop ? t('common.clear') : t('common.refresh')}
```
jeffvli, line 307: Revert the onClick handler to use the onRemoveLyric.

jeffvli (changes requested):

src/renderer/features/lyrics/lyrics-actions.tsx
Allinen: you're right, basically just boils down enabling the button on the web and forcing the refresh text. Since the funcionality is the same it makes me wonder why web never had refresh or was it just something that was not noticed?

jeffvli: It was my understanding that it *did* have the refresh button. The lyrics component was rewritten a short while back for the karaoke lyrics implementation so it must have been an oversight there.

PR #2453, opened 2026-09-13
feat(player): add setting to restart queue on end

### LINKED ISSUE #2452: [Feature]: Option to restart queue to the first track when it finishes playing

- [x] Yes

Not server-specific

Currently, when the queue finishes playing (repeat off, no Auto-DJ), the player stops and the queue position stays parked on the last track. Pressing play again just replays the last track of the queue.
  
  ### Desired behavior
  
A new opt-in setting that, when the queue ends under those same conditions, resets the queue position back to the first track instead of staying on the last one (playback still stops either way - this only affects where the queue "head" ends up). This matches how Spotify, Apple Music and other Navidrome clients' queues behave.
The current behavior would remain the default setting.

msanlop:
Adds an opt-in "Loop queue back to start on end" setting (Playback settings, off by default). When the queue finishes playing with repeat off, this resets the queue position to the first track instead of leaving it parked on the last one - closer to Spotify/Apple Music. Playback still stops either way, this only changes where the queue position ends up.

Repeat one/all, shuffle, and Auto-DJ are unaffected - the change is scoped to the single `shouldStop` branch in `calculateNextIndex` that only fires on repeat-off + last track.

  Closes https://github.com/jeffvli/feishin/issues/2452


jeffvli:
I'd rather not add a new setting just for this behavior.

It's a fairly minor behavior change so I don't mind using that as the default and it does make sense to me to do so.

jeffvli (changes requested):
Remove the new setting and just use as default behavior.

msanlop:
Alright, I figured maybe there was a reason for the current behavior, or that users would miss it. But restarting the queue seems more common for music players and makes more sense to me. I think youtube music and video players like vlc don't restart the queue

jeffvli (approved):

Issue #2403, opened 2026-08-29
[Feature]: Proposal: show Jellyfin audiobook libraries


TheSpyder:
### I have already checked through the existing feature requests and found no duplicates

- [x] Yes

### Is this a server-specific feature?

Jellyfin

### What do you want to be added?

Audio tracks in Jellyfin music libraries do not support storing a resume position, and Jellyfin marks them as played (server-side) the moment a track starts. This makes listening to longer audio tracks (e.g. concert recordings, 60+ minutes long) quite awkward. I have a smart playlist that I tried to set up to only show unplayed recordings and when using Music tracks it removes the currently playing track immediately.

Thank you for merging my PR #2395, that helps a lot, but if I switch to listening to other music the position in the long recording is still lost. After searching around for ideas I came across a workaround - move these recordings into a _books_ library, where they show up as audiobooks. Jellyfin does persist the playback position for these.

Feishin has quickly become my favourite Jellyfin music player so I had Claude take a crack at adding audiobook support. I'm not going to PR this unless you agree to the change, it still needs some cleanup, but I've been running it for a week and flushed out the bugs.
https://github.com/TheSpyder/feishin/compare/development...feat/audiobooks

There are three commits:
- Read the `PlaybackPositionTicks` property in Jellyfin's item data and seek to it when the song starts
- Allow skip-stop-scrobble events for audiobooks, since the server does persist the position for them
- Extend Feishin's library queries to show books

The third commit expanded after I decided to make Feishin's song query dynamic based on the library type rather than updating the hard-coded `Audio` to `Audio | AudioBook`, other than that the changes turned out to be quite small.

I'm hoping you'll at least consider accepting the playback position fixes, if the whole idea is too much. My audiobook playlist shows up in Feishin even without these changes so those two minor fixes are enough to solve my needs.

jeffvli:
I'm tentatively ok with supporting audiobooks, but I would like it to be scoped as an optional feature toggled via the server settings.

<img width="381" height="618" alt="Image" src="https://github.com/user-attachments/assets/6dfdeed0-d750-47cf-915b-8bb42f00c707" />

I took a look at your commits and I do already have a few suggestions but feel free to open your PR and I can review when I have time.

TheSpyder:
Thank you for considering my request! I will work on cleaning up the claude-isms and add a server settings toggle for showing audiobook libraries.

TheSpyder:
[...]

PR #2462, opened 2026-09-14
Audiobook support

TheSpyder:
This implements my audiobook proposal, resolving #2403. I found some significant places to clean up the AI mess; hopefully they were the ones you hinted at.

I've added a checkbox to the Jellyfin server config (both regular and quick-connect screens) to enable audiobooks. When this is enabled:
- Song requests sent to the server include audiobooks as a filter. Jellyfin is fine with requesting ItemTypes `Audio,AudioBook` even though libraries can never contain both.
- The library list and scan include libraries of the "Books" type.
- The folder view shows audiobook items instead of empty folders in Books libraries.
- Playlists already displayed audiobook items before this change (that's what gave me the idea to do this).

I made a few small quality-of-life changes across all playback for Jellyfin content, which I did not restrict to when the setting is enabled:
- Playing a resumable item will read the timestamp from the server and start from there (after a small delay, I checked and this is still necessary as it is in `use-queue-restore.ts`).
- Playing a resumable item will send progress updates.
- Stop scrobble events are sent for resumable items.
- I swapped the order of song change scrobbling to send the stop for the old track before the start for the new one. They're both `async` so there's no guarantee, but it seems more correct to me.


src/renderer/features/player/hooks/use-scrobble.ts
```diff
@@ -251,6 +251,8 @@ export const useScrobble = () => {
             const currentSong = usePlayerStore.getState().getCurrentSong();
             const mediaType = currentSong?._itemType.includes('song') ? 'song' : 'podcast';
             const useTicks = currentSong?._serverType === ServerType.JELLYFIN;
+            const hasPlaybackReport =
+                currentSong?._serverType === ServerType.JELLYFIN && currentSong.isResumable;
```
jeffvli, line 255: Use a different variable name here as this becomes incredibly confusing because that same variable name is also scoped in the other scrobble functions with the purpose of denoting the `ServerFeature.REPORT_PLAYBACK` status.

jeffvli (changes requested):
Looks good overall! Only a small minor change needed.

PR #2468, opened 2026-09-14
fix(library): keep grid item height when all rows are disabled

### LINKED ISSUE #2467: [Bug]: grid item height gets to max height when all grid rows get disabled

- [x] Yes

### App Version

1.16.0

### Music Server and Version

Navidrome v0.64.0

Desktop Windows

### What happened?

On any grid, when disabling all the row items, the row height gets extended to max height

With title, artist and year enabled:

Image

With all fields disabled:

Image

### Steps to reproduce

1. Go to albums page
2. Ensure you are using a grid layout
3. Configure it to disable all fields

### Relevant log output

```shell

```

GerardGascon:
Fixes #2467 

I essentially replaced a || with a ??. Which was causing the 0 item count to be interpreted as a null value and then defaulting to max value.


GerardGascon:
idk why the checks are failing

jeffvli:
It was a problem with the vite config, if you rebase development onto your branch it should resolve the build errors.

GerardGascon:
Alright, now it seems to work

PR #2469, opened 2026-09-14
FIX: tray icon missing on Linux and app not exiting with "Exit to tray"

ilgabbo:
Fix missing tray icon on Linux and stuck process on quit

Since 1.16.0 the tray icon doesn't show up on Linux (KDE/Wayland here, GNOME reported too). Turns out Electron 43.4.1 has a regression in the StatusNotifierItem registration (electron/electron#53213), fixed in 43.5.0. Bumped to 43.7.0.

While looking into it I also noticed that forceQuit is only set on macOS. On Linux with "Exit to tray" enabled, a SIGTERM (logout, kill, menu Quit) gets swallowed by the close handler: the window is hidden but the process never exits. Removed the platform check, before-quit means a real quit everywhere.

Tested on KDE Plasma 6 / Wayland: tray icon is back, kill <pid> now exits cleanly. I don't have macOS or Windows at hand, would appreciate a quick check there.


src/main/index.ts
```diff
@@ -372,9 +372,7 @@ let currentSidebarCollapsed = false;
 let currentShuffleEnabled = false;
 
 app.on('before-quit', () => {
-    if (isMacOS()) {
-        forceQuit = true;
-    }
```
jeffvli, line 377: The forceQuit is specifically for native macOS behavior because it needs to call app.exit(). The tray exit button calls app.quit() which should handle the quit behavior for Windows and Linux.

Did you try by only bumping the electron version and excluding this change?

ilgabbo: Yes, I did try with just the bump. Tray icon is back and Quit from the tray works fine, that path goes through exitFromTray so it was never the problem.

What's broken is a quit that doesn't come from the tray, e.g. kill <pid> or a session logout. Electron turns SIGTERM into app.quit(), but exitFromTray is false, so with "Exit to tray" on the close handler does preventDefault() and hides the window. That cancels the quit and the app just keeps running invisible until you kill -9 it.

Bump only, on KDE Wayland:
`App quitting { reason: 'before-quit' }
Main window hidden to tray`

and pgrep still shows the processes. With the change:
`App quitting { reason: 'forceQuit' }
Main window closed`

Nothing changes on macOS since the flag was already set there.

jeffvli: I overlooked your initial description about SIGTERM. Your case is valid.

Issue #1372, opened 2025-12-11
[Feature]: Allow custom ordering of artist top songs


trecoool:
### I have already checked through the existing bug reports and found no duplicates

- [x] Yes

### App Version

1.0 beta

### Music Server and Version

0.59

### What local environments are you seeing the problem on?

Desktop Windows

### What happened?

When looking at an artist's page, the songs order is unrelated to their rating.
I would expect to see favorited songs, then songs ordered by their rating.
Maybe they are ordered according to an online ranking? If that's the case, this rating should be used after the personnal rating, for songs not rated.
Alternatively, unrated songs could also be ordered according to their playcount.
A setting to fine tune all this would be great! Like having access to filters for this section as well.

### Steps to reproduce

- Like and favorite some songs
- Go to artist page of these songs, and see that in the top section their rating does not affect their ordering.

### Relevant log output

```shell

```

jeffvli:
Yeah, not a bug. These are ordered from external sources.

jeffvli:
These are only ordered by playCount to match the behavior of the original Community based order. I might implement a custom ranking configuration in the future to take into account favorite / rating.

mattbloody:
> These are only ordered by playCount to match the behavior of the original Community based order. I might implement a custom ranking configuration in the future to take into account favorite / rating.

do it!

PR #2488, opened 2026-09-16
fix(library): keep top songs mounted when empty

strecke:
Fixes a UX dead-end on the artist detail page where switching the "Top Songs" section from **Community** to **Personal** on an artist with 0 personal plays caused the section to unmount (`return null;`).
Because the active view is stored in `localStorage`, this removed the `SegmentedControl` toggle from the DOM, locking the user out across other artists.

Also adjusts the visibility check so the section remains hidden if an artist has no top tracks in either source (e.g. non-track contributors or artists with 0 community and 0 personal plays), while staying mounted if at least one source has data.

### Root Cause
In commit `bc6cd5b01`, `if (!isLoading && songs.length === 0) return null;` was added for layout spacing. When `f56a836f` (#1372) added the personal/community toggle with `localStorage` support, returning `null` on an empty active source removed the toggle from the DOM, locking the user into "Personal" mode.

### Changes
- Queries both `community` and `personal` top songs in parallel in `AlbumArtistMetadataTopSongsContent`.
- Only unmounts the section (`return null;`) if *neither* source has tracks (`hasAnyTopSongs === false`).
- Renders empty state `common.noResultsFromQuery` when the active source has no tracks.
- Also renders empty state when search filtering yields no matching tracks.
- Renders the "View all" button and icons only when the current source has songs (`songs.length > 0`).

### Verification
- Switched from Community (with plays) to Personal on an artist without personal plays. The section stays mounted, displays empty state, and allows toggling back.
- Verified that artists with 0 tracks in both sources unmount without blank space.
- Verified client-side search displays the empty state when no tracks match.
- Tested both web and desktop versions.

#### Community
image

#### Personal
image

#### Search
image


jeffvli:
Valid, I noticed this behavior previously but never got around to fixing it myself.
Thanks!

PR #2493, opened 2026-09-17
fix(library): keep favorite songs mounted when empty

strecke:
I came across the exact same UX behavior (#2488) in another place: the "Favorite Songs" section on the Album Artist detail page.

It fixes a UX dead-end where switching the "Favorite songs" section from Favorites to Ratings on an artist with 0 rated songs caused the section to unmount. Because the active view is stored in localStorage, unmounting the section permanently removed the SegmentedControl toggle from the DOM, locking the user into the empty "Ratings" state across other artists.

### Changes
- Query both favorites and ratings (respecting Jellyfin which do not support ratings).
- The section now only unmounts when neither source has tracks (hasAnyFavoriteSongs === false). Artists with no favorites or ratings remain uncluttered.
- When the active source has no tracks but the section is mounted, display the centered common.noResultsFromQuery text.
- Display the same empty state if client-side search filtering yields no results.
- Hide "View all" header button and play controls when there are no songs for the currently active source.

### Verification
- Switch from Favorites to Ratings on an artist with 0 rated songs. The section stays mounted, displays the empty state, and allows toggling back.
- Verified that artists with 0 tracks in both sources (0 favorites and 0 ratings) unmount without blank space.
- Verified client-side search displays the empty state when no tracks match.
- Tested on both web and desktop versions.
- **Note:** Not tested against Jellyfin (ratings query is conditionally disabled via existing server type check `canShowRatings`).


src/renderer/features/artists/components/album-artist-detail-content.tsx
```diff
@@ -712,10 +728,16 @@ const AlbumArtistMetadataFavoriteSongs = ({
         onLongPress: () => handlePlay(LONG_PRESS_PLAY_BEHAVIOR[Play.LAST]),
     });
 
-    const isLoading = favoriteSongsQuery.isLoading || !favoriteSongsQuery.data;
+    const isChecking =
+        (showFavorites && favoritesQuery.isLoading) || (canShowRatings && ratingsQuery.isLoading);
+    const isLoading =
+        favoriteSongsQuery.isLoading || (!favoriteSongsQuery.data && !favoriteSongsQuery.isError);
```
jeffvli: This should include the loading state for the ratingsQuery as well?

jeffvli (changes requested):

src/renderer/features/artists/components/album-artist-detail-content.tsx
strecke: favoriteSongsQuery is the active query (aliased to either ratingsQuery or favoritesQuery based on favoriteSongsQueryTypeFilter, matching the pattern from #2488). So when the ratings view is active, it already tracks ratingsQuery.isLoading.

Because favoritesQuery and favoriteSongsQuery sound almost identical, it's easy to misread. I think renaming favoriteSongsQuery to activeQuery would make this clearer!

jeffvli: You're right, I misread it as I was quickly scanning through.

PR #2494, opened 2026-09-17
fix(ui): keep table column resize within bounds

TidBits16:
## Overview
Columns in the queue could get squashed and become no longer accessible. This would require reinstalling in most cases to resolve. Limited columns from overlapping each other and added drag logic to make dragging more seamless.

## Before
very sad. many favorite buttons were lost this way 😔.

before

## After
(forgot to record my cursor) no more buttons getting lost into oblivion!

after

I tweaked the CSS to make the drag handle stay visible while mouse-down so it doesn't become invisible while adjusting.

Let me know  if there's any tweaks or adjustments that need to be made!


src/renderer/components/item-list/item-table-list/columns/row-index-column.module.css
TidBits16, line 1: I left justified the icons so they match the labels and to make finding the " | " easier but I totally can revert this change 👍

TidBits16:
found a bug - currently addressing it.

TidBits16:
Ok fixed :+1:

The columns also had a max-width so it would prevent them from being expanded to their _full potential!_ (could be a problem on widescreen or high dpi monitors :P). This is fixed as well.

jeffvli:
> Columns in the queue could get squashed and become no longer accessible. This would require reinstalling in most cases to resolve.

This comment is incorrect. The table configuration allows you to manually set the size if the size ever became zero, would not require a reinstall.

src/renderer/components/item-list/item-table-list/columns/title-artist-column.tsx
```diff
@@ -22,7 +22,7 @@ export const DefaultTitleArtistColumn = (props: ItemTableListInnerColumn) => {
     const rowItem = props.getRowItem?.(props.rowIndex) ?? (props.data as any[])[props.rowIndex];
     const row: object | undefined = (rowItem as any)?.id;
     const item = rowItem as any;
-    const align = props.columns[props.columnIndex]?.align || 'start';
```
jeffvli, line 25: Changing the default column orientation is one thing, but this change now breaks the column alignment configuration.

You have multiple changes in place which removes the column prop config and hardcodes it as 'start'. These need to be removed.

src/renderer/components/item-list/item-table-list/default-columns.ts
```diff
@@ -23,7 +23,7 @@ export const SONG_TABLE_COLUMNS: DefaultTableColumn[] = [
         width: 240,
     },
     {
-        align: 'center',
```
jeffvli, line 26: Please undo all changes ot the default columns

jeffvli (changes requested):

PR #2495, opened 2026-09-18
fix(visualizer): capture stereo in system loopback

xela-zone:
Chromium getDisplayMedia defaults to mono with voice processing on. Explicitly request 2 channels and disable AGC, echo, and noise filters.

# problem
when playing audio using MPV or another playback engine that uses desktop capture for visualization, the visualizer displays mono sound, instead of stereo.

# root cause
the default settings for `getDisplayMedia({audio: true})` appear to match a 'conference call' or similar use case, rather than a closer to raw capture. 

instead of using the defaults, i've changed it to explicitly request: 2 channels no AGC, echo cancellation, or noise suppression. 

# verification
verified by listening to Jerobeam Fenderson's *Oscillopscope Music*. it has heavy stereo panning.


jeffvli:
Thanks, did not realize that!

